PR #7, opened 2024-10-11
.github/worksflows: install composefs before testing

### LINKED ISSUE #5: CI is broken due to missing libcomposefs on Ubuntu
Since we took a depend on the composefs crate, we need to have libcomposefs available in the system, but it looks like we don't have that on Ubuntu.

We should probably run CI inside of some kind of a Fedora-based container.

allisonkarlitskaya:
Closes #5

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,9 +12,21 @@ env:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Ugh: Error: Operation not supported (os error 95)

That's the "your fs doesn't do fs-verity" thing.  Testing this stuff is gonna be rough.

PR #12, opened 2024-10-19
Use tempfile by default

cgwalters:
Per discussion this crate is widely used and there's no reason to carry even a small reimplementation.


allisonkarlitskaya (approved):
LGTM.  Thanks!

PR #23, opened 2024-11-04
repo: Use `?` when traversing directories

cgwalters:
This is more idiomatic and avoids a lot of indentation.


allisonkarlitskaya (approved):
Thanks.  I think I agree with this approach now (and indeed, I even wrote some code yesterday that looks like this).

It's a shame we don't have a failable iterator syntax in Rust yet.  Maybe someday...

PR #24, opened 2024-11-06
 Use composefs crate for fsverity digest reading

cgwalters:
tests: Add support for CFS_TEST_TMPDIR

Because we require fsverity in the tests, make this easy
to configure externally for different test harnesses to use.

Signed-off-by: Colin Walters <walters@verbum.org>

---

ci: Switch to using Fedora in a container

Prep for linking to the composefs crate and C shared library
to ensure we share code and maintain an incentive to contribute
to those.

Signed-off-by: Colin Walters <walters@verbum.org>

---

Use composefs crate for fsverity digest reading

To increase the motivation to improve that crate and the C library;
especially the C library, because we *must* have it anyways because
we depend on the binaries and we are just not going to rewrite
everything in Rust in the near future.

(And if we *did* start a rewrite of the composefs core I think
 that rewrite should live in that repo, not this one)

Signed-off-by: Colin Walters <walters@verbum.org>

---

### DIFF
--- a/src/fsverity/ioctl.rs
+++ b/src/fsverity/ioctl.rs
@@ -3,7 +3,7 @@ use std::os::fd::AsFd;
 use anyhow::Result;
 use rustix::ioctl;
 
-use super::FsVerityHashValue;
+use super::{FsVerityHashValue, Sha256HashValue};
 
 // See /usr/include/linux/fsverity.h
 #[repr(C)]
@@ -43,36 +43,10 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<()> 
     Ok(())
 }
 
-#[repr(C)]
-pub struct FsVerityDigest<F> {
-    digest_algorithm: u16,
-    digest_size: u16,
-    digest: F,
-}
-
-// #define FS_IOC_MEASURE_VERITY   _IORW('f', 134, struct fsverity_digest)
-type FsIocMeasureVerity = ioctl::ReadWriteOpcode<b'f', 134, FsVerityDigest<()>>;
-
-pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H> {
-    let digest_size = std::mem::size_of::<H>() as u16;
-    let digest_algorithm = H::ALGORITHM as u16;
-
-    let mut digest = FsVerityDigest::<H> {
-        digest_algorithm,
-        digest_size,
-        digest: H::EMPTY,
-    };
-
-    unsafe {
-        ioctl::ioctl(
-            fd,
-            ioctl::Updater::<FsIocMeasureVerity, FsVerityDigest<H>>::new(&mut digest),
-        )?;
-    }
-
-    if digest.digest_algorithm != digest_algorithm || digest.digest_size != digest_size {
-        Err(std::io::Error::from(std::io::ErrorKind::InvalidData))?
-    } else {
-        Ok(digest.digest)
-    }
+pub fn fs_ioc_measure_verity<F: AsFd>(fd: F) -> Result<Sha256HashValue> {
+    let mut digest = composefs::fsverity::Digest::new();
+    composefs::fsverity::fsverity_digest_from_fd(fd.as_fd(), &mut digest)?;
+    let mut r = Sha256HashValue::EMPTY;
+    r.copy_from_slice(digest.get());
+    Ok(r)
 }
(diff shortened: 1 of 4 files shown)


src/fsverity/ioctl.rs
```diff
@@ -43,36 +43,10 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<()>
...
-pub struct FsVerityDigest<F> {
-    digest_algorithm: u16,
-    digest_size: u16,
-    digest: F,
-}
-
-// #define FS_IOC_MEASURE_VERITY   _IORW('f', 134, struct fsverity_digest)
-type FsIocMeasureVerity = ioctl::ReadWriteOpcode<b'f', 134, FsVerityDigest<()>>;
-
-pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H> {
```
cgwalters, line 56: That said this code is quite elegant and nice; the generics usage for sha256/sha512 is cool. But OTOH there are no plans to use anything but sha256 for composefs.

What we could do though is move some of this "reimplement composefs core in Rust" into the composefs git repo under the composefs crate under an optional feature flag or something - might be especially useful to "cross test" i.e. verify the C and Rust codepaths do the same thing.

src/fsverity/ioctl.rs
```diff
@@ -43,36 +43,10 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<()>
...
-        )?;
-    }
-
-    if digest.digest_algorithm != digest_algorithm || digest.digest_size != digest_size {
-        Err(std::io::Error::from(std::io::ErrorKind::InvalidData))?
-    } else {
-        Ok(digest.digest)
-    }
+pub fn fs_ioc_measure_verity<F: AsFd>(fd: F) -> Result<Sha256HashValue> {
+    let mut digest = composefs::fsverity::Digest::new();
```
cgwalters, line 47: Also looking at this, perhaps we should drop the `Digest` wrapper and do what you have here and just have it write into an array. Definitely the compile-time known size is better than what we have in `.get()` at the moment.

OTOH, and this relates to a larger topic...I feel like in practice we end up converting to hex in many places anyways, so we may as well canonicalize to e.g. [this Sha256Digest](https://docs.rs/oci-spec/latest/oci_spec/image/struct.Sha256Digest.html) instead.

allisonkarlitskaya:
I'm afraid I don't like this change.  I really like the way the ioctl interface in rustix looks... and in general, if we can avoid the trouble of taking the dependency I'd really prefer to do so.  Indeed: we use `mkcomposefs` at runtime, but not from the tests that we run in GitHub actions.

I kinda get some of the higher-level ideas about unifying this repo with the composefs repo a bit more...

cgwalters:
> f we can avoid the trouble of taking the dependency I'd really prefer to do so. 

Is there a better reason that the below :arrow_down: (That I already addressed via using Fedora)?

> Indeed: we use mkcomposefs at runtime, but not from the tests that we run in GitHub actions.

Sure but in reality I think what we'll want is *integration* tests and it's those that can require fsverity. Requiring fsverity for unit tests isn't *wrong*, but it isn't right either.

cgwalters:
I do think we want to help get the composefs C core packaged for Ubuntu and Arch etc. in the near term which would help address the dependency issue.

cgwalters:
> I really like the way the ioctl interface in rustix looks... a

Me too, but to be clear this is just starting the ball rolling on deduping things. I don't think it makes sense for this crate to have its own dumpfile code either for example.

cgwalters:
We had a quick realtime debate on this and the conclusion was:

- We will depend on the `mkcomposefs` and `composefs-info dump` for the near term future
- But the libcomposefs C library is not doing enough today that we need to take a dependency on it

Instead we will:

- Make this repository be containers/composefs-storage or so (or maybe containers/composefs-rs)
- Merge the good code from the existing Rust crates in containers/composefs into that combined repository (most likely at least the dumpfile parser/generator)
- Publish the combined merged thing as `composefs` to crates.io

Incidentally in this I think we should actually switch over this crate to being a workspace...if you look at the `composefs` crate today it's just a thin binding for working with `mkcomposefs` and `composefs-info dump`, plus fsverity helpers. The fsverity helpers can be pure Rust for now, but those would I think continue to live in a `composefs` core crate.

I think then what is in `repository.rs` here should probably become a `composefs-storage` crate that builds on that with many more opinions about things such as all of the repo management, etc.

PR #27, opened 2024-11-07
repo: Use ? in another place

cgwalters:
Just reading the code came across another place we just want to be using `?`.


allisonkarlitskaya (approved):
Thanks!

PR #30, opened 2024-11-07
selinux relabel support

### LINKED ISSUE #29: SELinux support for `oci create-image`
The `oci create-image` command should label all the files using the default labels set in the policy from the container.

If a user wants to change a label for a file, they can modify the SELinux policy file mapping using semanage in a container layer and that will be taken into account by this.

To do that, we need to lookup all the files labels when we create the composefs blob.

Looking up a file label is done like that in C (from Allison's code):

```
#include <stdlib.h>
#include <unistd.h>
#include <selinux/label.h>
#include <selinux/selinux.h>
#include <sys/stat.h>
#include <err.h>
#include <stdio.h>
 
int
main (int argc, char **argv)
{
    struct selinux_opt opts[] = {
    };
 
 
  struct selabel_handle *hnd = selabel_open (SELABEL_CTX_FILE, opts, 0);
 
  for (int i = 1; i < argc; i++)
    {
      struct stat buf;
 
      if (lstat (argv[i], &buf) != 0)
        err(EXIT_FAILURE, "stat %s: ", argv[i]);
 
      char *ctx = NULL;
      if (selabel_lookup (hnd, &ctx, argv[i], buf.st_mode & S_IFMT) != 0)
        err(EXIT_FAILURE, "selabel_lookup('%s', %d): ", argv[i], buf.st_mode & S_IFMT);
 
      printf ("%s → %s\n", argv[i], ctx);
      freecon (ctx);
    }
 
  selabel_close (hnd);
}
```

As we need to use the SELinux policy from the container, we need to:
- figure out which policy is going to be used (or maybe we just extract all the policy available?)
- extract the SELinux file contexts mapping files from the container to a temporary path:
  - `/etc/selinux/{SELINUXTYPE}/contexts/files/file_contexts`
  - ...
  - See https://man7.org/linux/man-pages/man5/selabel_file.5.html
-  point libselinux to the extracted `files_contexts` by setting the folllowing `selinux_opt`:
  - `SELABEL_OPT_PATH` -> `/path/to/etc/selinux/{SELINUXTYPE}/contexts/files/file_contexts`

allisonkarlitskaya:
There are some outstanding items here:
 - [x] figure out which policy to check.  It's hardcoded to `targeted` right now
 - [x] provide a nice fallback if the policy isn't present.  It's currently fatal.
 - [x] ~figure out a nicer way to handle linking to `libselinux`, possibly making it a conditional feature?~ don't link libselinux: do it ourselves
 - [x] figure out a better way to handle mutability of hardlinks in the filesystem tree
 - [x] properly label our `/run/systemd/volatile-root` hack symlink to avoid this: `[    3.055118] audit: type=1400 audit(1731016644.835:3): avc:  denied  { read } for  pid=447 comm="systemd-gpt-aut" name="volatile-root" dev="tmpfs" ino=679 scontext=system_u:system_r:systemd_gpt_generator_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=lnk_file permissive=1
`
 - [x] drop `enforcing=0`

Closes #29

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,9 +18,7 @@ jobs:
     - name: Enable fs-verity on /
       run: sudo tune2fs -O verity $(findmnt -vno SOURCE /)
     - uses: actions/checkout@v4
-    - name: Build
-      run: cargo build --verbose
-    - name: Clippy
-      run: cargo clippy
-    - name: Run tests
-      run: cargo test --verbose
+    - run: cargo build --verbose
+    - run: cargo fmt --check
+    - run: cargo clippy -- -Dwarnings
+    - run: cargo test --verbose
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ hex = "0.4.3"
 indicatif = { version = "0.17.8", features = ["tokio"] }
 oci-spec = "0.7.0"
 rand = "0.8.5"
+regex-automata = "0.4.8"
 rustix = { version = "0.38.37", features = ["fs", "mount", "process"] }
 sha2 = "0.10.8"
 tar = "0.4.42"
--- a/examples/uki/Containerfile
+++ b/examples/uki/Containerfile
@@ -22,8 +22,17 @@ RUN <<EOF
     # we should install kernel-modules here, but can't
     # because it'll pull in the entire kernel with it
     # it seems to work fine for now....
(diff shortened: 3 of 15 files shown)


travier:
As far as I know, the SELinux policy type is taken from `/etc/selinux/config`:

```
$ grep "^SELINUXTYPE=" /etc/selinux/config`
SELINUXTYPE=targeted
```

PR #31, opened 2024-11-09
tests: Drop an unnecessary vec allocation

cgwalters:
We don't need to heap allocate a whole zeroed buffer here, we can create a virtual reader that returns exactly that number of zeroed bytes dynamically.


allisonkarlitskaya (approved):
Cool!

PR #33, opened 2024-11-12
dumpfile_parse: New module

cgwalters:
This was imported from the dumpfile.rs from
https://github.com/containers/composefs/commit/738793aac459fc735790870fcf686a09cd1a151e

Per discussion it's generally useful to be able to fully parse a composefs file back into a dumpfile.


src/dumpfile_parse.rs
```diff
@@ -0,0 +1,867 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+/// A serializable composefs entry.
+///
+/// The `Display` implementation for this type is defined to serialize
+/// into a format consumable by `mkcomposefs --from-file`.
+pub enum Item<'p> {
+    /// A regular file
+    Regular {
```
allisonkarlitskaya, line 83: This is an insufficient representation for regular files: they also need to be able to carry a payload in order to reference the backing file.  See https://man.archlinux.org/man/composefs-dump.5.en#PAYLOAD

src/tests/assets/special.dump
```diff
@@ -0,0 +1,10 @@
+/ 4096 40555 2 0 0 0 1633950376.0 - - - trusted.foo1=bar-1 user.foo2=bar-2
```
allisonkarlitskaya, line 1: The testcase would benefit from the addition of a non-inline regular file.

allisonkarlitskaya (changes requested):
I think it would be a great benefit if this code had a bidirectional interaction with the code in `image.rs`.  That's the main "filesystem structure" in this crate and it would be a bit weird not to interact with it.  I've already made quite some efforts to use the printer code from this module to print from that structure (we can discuss this today), and parsing to it would be helpful as well.  The `oci/tar` module does that, which should be a fairly similar vibe.

I don't think it makes sense to just drop this code in without attempting to integrate it.  As it stands, the only thing that you can do with this structure after you read it in is to print it back out again.  We should discuss this, though.

cgwalters:
OK addressed the bugs, this one should be in good shape now!

> I don't think it makes sense to just drop this code in without attempting to integrate it. As it stands, the only thing that you can do with this structure after you read it in is to print it back out again. We should discuss this, though.

Well I dunno, there's quite a collection of stuff in this repository already, I don't think we're in any state to be saying we're committing to long term APIs.

I agree we should try to use this, but given that I think others *will* want to be able to parse composefs blobs, I think the code is valuable and again I see this as part of agreeing to merge our two projects.

cgwalters:
OK right, this has the same issue that motivated the CI changes in https://github.com/containers/composefs-rs/pull/24
I pulled those two changes into this PR.

allisonkarlitskaya (approved):

PR #36, opened 2024-11-15
Add support for creating a composefs from a directory

### LINKED ISSUE #26: use canonical recommend inlining
See
https://github.com/containers/composefs/pull/395

We should expose this constant in our core crate, and also use it ourselves.

### LINKED ISSUE #34: Import and compute composefs' fs-verity in a multi-staged build
In order to be able to run the entire build process from a multi-staged build via a single Containerfile build, we need cfsctl to read the OCI content using a --mount/--from option like in https://gitlab.com/fedora/bootc/base-images/-/blob/main/Containerfile.

cfsctl would read the entire container archive, generate the composefs dumpfile, pass it to mkcomposefs and finally get the fs-verity of the resulting EROFS.

Then this layer could build the UKI using this fs-verity-hash.

As we need this hash as a label (i.e. outside of the container), we would then store that as a file or output that on stdout at the end and another step in the build process would add the hash as a label (as it's not possible to add labels with values generated from stages right AFAIK).

allisonkarlitskaya:
This is basically the default mode of operation for `mkcomposefs`, but we're interested in getting exactly the same image as if we had created it from the layer tarballs.

This requires a bit more work, including adding selinux labels and some very careful thinking about the root directory.

This introduces a new `examples/unified` which builds the container in a single `podman build`.  It gets the same result inside and outside (or at least it does for me) and boots up normally.

A preparatory commit defines the inlining constant to 64 and uses it, which means that repositories need to be recreated.

Closes #34
Closes #26

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -1,16 +1,31 @@
-use std::{ffi::OsStr, mem::MaybeUninit, path::Path};
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, HashMap},
+    ffi::OsString,
+    ffi::{CStr, OsStr},
+    mem::MaybeUninit,
+    os::unix::ffi::{OsStrExt, OsStringExt},
+    path::Path,
+    rc::Rc,
+};
 
-use anyhow::Result;
+use anyhow::{bail, ensure, Result};
 use rustix::{
-    fd::OwnedFd,
-    fs::{fdatasync, linkat, mkdirat, mknodat, openat, symlinkat, AtFlags, FileType, OFlags, CWD},
+    fd::{AsFd, OwnedFd},
+    fs::{
+        fdatasync, fstat, getxattr, linkat, listxattr, mkdirat, mknodat, openat, readlinkat,
+        symlinkat, AtFlags, Dir, FileType, Mode, OFlags, CWD,
+    },
     io::{read_uninit, write, Errno},
 };
 
 use crate::{
-    image::{DirEnt, Directory, Inode, Leaf, LeafContent, Stat},
+    fsverity::{digest::FsVerityHasher, Sha256HashValue},
+    image::{DirEnt, Directory, FileSystem, Inode, Leaf, LeafContent, Stat},
     repository::Repository,
+    selabel::selabel,
     util::proc_self_fd,
+    INLINE_CONTENT_MAX,
 };
 
 fn set_file_contents(dirfd: &OwnedFd, name: &OsStr, stat: &Stat, data: &[u8]) -> Result<()> {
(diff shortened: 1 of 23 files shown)


src/fs.rs
```diff
@@ -97,7 +112,217 @@ fn write_directory_contents(dir: &Directory, fd: &OwnedFd, repo: &Repository) ->
...
+        // flistxattr() and fgetxattr() don't with with O_PATH fds, so go via /proc/self/fd. Note:
+        // we want the symlink-following version of this call, which produces the correct behaviour
+        // even when trying to read xattrs from symlinks themselves.  See
+        // https://gist.github.com/allisonkarlitskaya/7a80f2ebb3314d80f45c653a1ba0e398
+        let filename = proc_self_fd(fd);
+
+        let mut xattrs = BTreeMap::new();
+
+        let names_size = listxattr(&filename, &mut [])?;
+        let mut names = vec![0; names_size];
```
cgwalters, line 139: Why isn't this `vec![0u8; names_size]` to start?

src/fs.rs
```diff
@@ -97,7 +112,217 @@ fn write_directory_contents(dir: &Directory, fd: &OwnedFd, repo: &Repository) ->
...
+        let actual_names_size = listxattr(&filename, &mut names)?;
+        ensure!(
+            actual_names_size == names.len(),
+            "xattrs changed during read"
+        );
+
+        let names: Vec<u8> = names.into_iter().map(|c| c as u8).collect(); // fml
```
cgwalters, line 146: Then we could drop this

src/fs.rs
```diff
@@ -97,7 +112,217 @@ fn write_directory_contents(dir: &Directory, fd: &OwnedFd, repo: &Repository) ->
...
+
+        let mut buffer = [0; 65536];
```
cgwalters, line 148: Should also be `0u8` here right?

src/fs.rs
```diff
@@ -97,7 +112,217 @@ fn write_directory_contents(dir: &Directory, fd: &OwnedFd, repo: &Repository) ->
...
+            Ok(Inode::Leaf(self.read_leaf(dirfd, name, ifmt)?))
+        }
+    }
+}
+
+pub fn read_from_path(path: &Path, repo: Option<&Repository>) -> Result<FileSystem> {
+    let mut reader = FilesystemReader {
+        repo,
+        inodes: HashMap::new(),
+        st_dev: u64::MAX,
```
cgwalters, line 299: Ditto re `Option`

src/fs.rs
```diff
@@ -97,7 +112,217 @@ fn write_directory_contents(dir: &Directory, fd: &OwnedFd, repo: &Repository) ->
...
+            between readdir() and fstat()"
+        );
+
+        let mtime = buf.st_mtime as i64;
+
+        if buf.st_dev != self.st_dev {
+            if self.st_dev == u64::MAX {
+                self.st_dev = buf.st_dev;
+            } else {
+                bail!("Attempting to cross devices while importing filesystem");
```
cgwalters, line 175: There's also https://docs.rs/openat2/latest/openat2/struct.ResolveFlags.html#associatedconstant.NO_XDEV

cgwalters (approved):
Really nice work overall, just some nits

src/image.rs
```diff
@@ -172,10 +186,10 @@ impl FileSystem {
         FileSystem {
             root: Directory {
                 stat: Stat {
-                    st_mode: 0o755,
-                    st_uid: 0,
-                    st_gid: 0,
-                    st_mtim_sec: 0,
+                    st_mode: u32::MAX, // assigned later
```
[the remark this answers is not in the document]
allisonkarlitskaya: That would require adding `Option` on the `Stat` structure, which doesn't make sense for any case other than this one.  This is ugly, but it works.

src/fs.rs
allisonkarlitskaya: This API is annoying and awful.  We don't know the names size in advance, so we have to query it in order to allocate the buffer.  Fair enough: you do that by submitting a request with an empty buffer and get the correct size returned.  Then you allocate and do it again.  There's a race, of course, if the xattrs changed on disk in the meantime.

I was using a static buffer here at first (which I also shared with the static buffer I use for reading the actual values) but that brings us to the next problem: this API is defined in terms of `c_char` which is `i8` — not `u8`.  A slice of `c_char` might sound like it kinda makes sense when you're dealing with a C API, but when you write it (equivalently) as `[i8]` you notice how bizarre it is.  I can't really see this as anything other than a rustix bug, to be honest...

src/fs.rs
allisonkarlitskaya: I'd love to drop it, but your above suggestion, for example, gets me:

```
mismatched types
   --> src/fs.rs:140:54
    |
140 |         let actual_names_size = listxattr(&filename, &mut names)?;
    |                                 ---------            ^^^^^^^^^^ expected `&mut [i8]`, found `&mut Vec<u8>`
    |                                 |
    |                                 arguments to this function are incorrect
    |
    = note: expected mutable reference `&mut [i8]`
               found mutable reference `&mut Vec<u8>`
note: function defined here
   --> /var/home/lis/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.40/src/fs/xattr.rs:140:8
    |
140 | pub fn listxattr<P: path::Arg>(path: P, list: &mut [c::c_char]) -> io::Result<usize> {
    |        ^^^^^^^^^


```

allisonkarlitskaya: We could also transmute, but I've been absolutely loathe to write the word `unsafe`... there are only a very small number of counterexamples in this entire repository.

src/fs.rs
allisonkarlitskaya: This one is inferred to u8, because *that* API (`getxattr`) wants `[u8]`, not `[i8]`.

src/fs.rs
allisonkarlitskaya: We could do `Option` here indeed, and it would be cleaner.  I did it the way that I did it because I wanted to compare directly because the common case is that they will indeed be equal.  Excessive micro-optimization, to be sure.

src/fs.rs
allisonkarlitskaya: This is really interesting, and I didn't know about it.  Thanks.

rustix binds that via https://doc.servo.org/rustix/backend/fs/types/struct.ResolveFlags.html

The trouble with this is that it would make reading the root directory be different from all the other inodes: entering the bindmount from `CWD` would be considered a violation of the `NO_XDEV` rule.  Still, though: we have a bunch of other ugly code here that's ugly specifically because the root node *isn't* handled separately, and it ought to be.

I'm going to try to see if I can use this flag and unroll the root directory case a bit to get something a bit nicer all around.  Thanks!

PR #37, opened 2024-11-15
.github/workflows: fix a couple of errors

allisonkarlitskaya:
This got introduced by cdcda8d8d80d and I missed it during review because it didn't result in any red on the pull request (but the failing workflow was visible in the Actions tab).


allisonkarlitskaya:
Green means go!

PR #44, opened 2024-12-04
fs: clean up filesystem import state

### LINKED ISSUE #41: unified image build, Error: xattrs changed during read
composefs-rs/examples/unified $ ./build

```
[2/3] STEP 1/3: FROM 4d08a5ad0f7b068e4fb32adfb25f3c542f05c4939372c3c521e7dfeab5b06a53 AS kernel
[2/3] STEP 2/3: RUN --mount=type=bind,from=base,target=/mnt/base <<EOF (set -eux...)
+ mkdir -p /tmp/sysroot/composefs/images
+ mkdir -p /tmp/sysroot/composefs/objects
++ cfsctl --repo /tmp/sysroot create-image /mnt/base
Error: xattrs changed during read
```

I can build bls or uki image, but unified failed in kernel stage.

allisonkarlitskaya:
Clean up two aspects of importing filesystems:

First: instead of trying to avoid crossing mountpoints (by checking `.st_dev`) in order to use `.st_ino` values as a "unique key" for tracking hardlinks, just take the `(.st_dev, .st_ino)` pair.  This resolves some issues with overlayfs where adjacent files can end up with different `.st_dev`.

Second: we were abusing the first setting of `st_dev` as a flag field to determine if we were `stat()`-ing the root inode or not, which was always a bit evil.  We did this to avoid collecting the mtime of the root directory for determining the newest file in the filesystem.  Let's change our approach here: we can collect the mtime when adding items to their parent directories (which is never done for the root directory). Use a impl on `Inode` to make it easier to collect that information.

These changes together turn .stat() into a pure operation (with respect to the state of the reader), so remove the `&mut self` reference.

Fixes #41

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name = "composefs"
 version = "0.2.0"
 edition = "2021"
+rust-version = "1.83.0"
 description = "Rust library for the composefs filesystem"
 keywords = ["composefs"]
 license = "MIT OR Apache-2.0"
--- a/src/dumpfile_parse.rs
+++ b/src/dumpfile_parse.rs
@@ -180,7 +180,7 @@ fn unescape_limited(s: &str, max: usize) -> Result<Cow<[u8]>> {
 
 /// Unescape a byte array according to the composefs dump file escaping format.
 fn unescape(s: &str) -> Result<Cow<[u8]>> {
-    return unescape_limited(s, usize::MAX);
+    unescape_limited(s, usize::MAX)
 }
 
 /// Unescape a string into a Rust `OsStr` which is really just an alias for a byte array,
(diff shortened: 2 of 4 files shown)


allisonkarlitskaya:
The CI failures look like the actions runners got the 2024-enabled 1.85 rust release and clippy isn't following our `edition = "2021"` line in `Cargo.toml`.  I tried eliding those lifetimes and it doesn't work in 2021.

allisonkarlitskaya:
Okay, so it's not an edition thing, it's just a version thing: this got added in 1.83.0.

I didn't know that changes like this could happen between releases...  The new version is in Fedora now, in any case, so let's just depend on it.

PR #48, opened 2024-12-04
fsverity: Use rustix Result, add fs_ioc_measure_verity_optional

cgwalters:
First switch to use `rustix::io::Result` since this is a raw low level API and we don't need generic errors.

Change the "wrong algorithm" error to `EILSEQ` which is what the C composefs library is doing today.

Next, add an API that maps the errno values to `Option` for callers that want to not have missing verity be a fatal error.

### DIFF
--- a/src/fsverity/ioctl.rs
+++ b/src/fsverity/ioctl.rs
@@ -1,9 +1,21 @@
 use std::os::fd::AsFd;
 
-use anyhow::Result;
+use rustix::io::Errno;
 use rustix::ioctl;
+use thiserror::Error;
 
 use super::FsVerityHashValue;
+
+/// Measuring fsverity failed.
+#[derive(Error, Debug)]
+pub enum MeasureVerityError {
+    #[error("i/o error")]
+    Io(#[from] std::io::Error),
+    #[error("Expected algorithm {expected}, found {found}")]
+    InvalidDigestAlgorithm { expected: u16, found: u16 },
+    #[error("Expected digest size {expected}")]
+    InvalidDigestSize { expected: u16 },
+}
 
 // See /usr/include/linux/fsverity.h
 #[repr(C)]
@@ -22,7 +34,10 @@ pub struct FsVerityEnableArg {
 // #define FS_IOC_ENABLE_VERITY    _IOW('f', 133, struct fsverity_enable_arg)
 type FsIocEnableVerity = ioctl::WriteOpcode<b'f', 133, FsVerityEnableArg>;
 
-pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<()> {
+/// Enable fsverity on the target file. This is a thin safe wrapper for the underlying base `ioctl`
+/// and hence all constraints apply such as requiring the file descriptor to already be `O_RDONLY`
+/// etc.
+pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> std::io::Result<()> {
     unsafe {
         ioctl::ioctl(
             fd,
@@ -43,6 +58,7 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<()> 
     Ok(())
 }
 
+/// Core definition of a fsverity digest.
 #[repr(C)]
 pub struct FsVerityDigest<F> {
     digest_algorithm: u16,
@@ -53,7 +69,10 @@ pub struct FsVerityDigest<F> {
 // #define FS_IOC_MEASURE_VERITY   _IORW('f', 134, struct fsverity_digest)
 type FsIocMeasureVerity = ioctl::ReadWriteOpcode<b'f', 134, FsVerityDigest<()>>;
 
-pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H> {
+/// Measure the fsverity digest of the provided file descriptor.
+pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(
+    fd: F,
+) -> Result<Option<H>, MeasureVerityError> {
     let digest_size = std::mem::size_of::<H>() as u16;
     let digest_algorithm = H::ALGORITHM as u16;
 
@@ -63,16 +82,48 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H> 
         digest: H::EMPTY,
     };
 
-    unsafe {
+    let r = unsafe {
         ioctl::ioctl(
             fd,
             ioctl::Updater::<FsIocMeasureVerity, FsVerityDigest<H>>::new(&mut digest),
-        )?;
+        )
+    };
+    match r {
+        Ok(()) => {
+            if digest.digest_algorithm != digest_algorithm {
+                return Err(MeasureVerityError::InvalidDigestAlgorithm {
+                    expected: digest.digest_algorithm,
+                    found: digest_algorithm,
+                });
+            }
+            if digest.digest_size != digest_size {
+                return Err(MeasureVerityError::InvalidDigestSize {
+                    expected: digest.digest_size,
+                });
+            }
+            Ok(Some(digest.digest))
+        }
+        // This function returns Ok(None) if there's no verity digest found.
+        Err(Errno::NODATA | Errno::NOTTY | Errno::OPNOTSUPP) => Ok(None),
+        Err(Errno::OVERFLOW) => Err(MeasureVerityError::InvalidDigestSize {
+            expected: digest.digest_size,
+        }),
+        Err(e) => Err(std::io::Error::from(e).into()),
     }
+}
 
-    if digest.digest_algorithm != digest_algorithm || digest.digest_size != digest_size {
-        Err(std::io::Error::from(std::io::ErrorKind::InvalidData))?
-    } else {
-        Ok(digest.digest)
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::fsverity::Sha256HashValue;
+
+    #[test]
+    fn test_measure_verity_opt() -> anyhow::Result<()> {
+        let tf = tempfile::tempfile()?;
+        assert_eq!(
+            fs_ioc_measure_verity::<_, Sha256HashValue>(&tf).unwrap(),
+            None
+        );
+        Ok(())
     }
 }
--- a/src/fsverity/mod.rs
+++ b/src/fsverity/mod.rs
@@ -1,7 +1,13 @@
+use std::os::fd::AsFd;
+
+use anyhow::Result;
+use ioctl::MeasureVerityError;
+use thiserror::Error;
+
 pub mod digest;
 pub mod ioctl;
 
-pub trait FsVerityHashValue {
+pub trait FsVerityHashValue: Eq + AsRef<[u8]> {
     const ALGORITHM: u8;
     const EMPTY: Self;
 }
@@ -18,4 +24,62 @@ pub type Sha512HashValue = [u8; 64];
 impl FsVerityHashValue for Sha512HashValue {
     const ALGORITHM: u8 = 2;
     const EMPTY: Self = [0; 64];
+}
+
+/// A verity comparison failed.
+#[derive(Error, Debug)]
+pub enum CompareVerityError {
+    #[error("failed to read verity")]
+    Measure(#[from] MeasureVerityError),
+    #[error("fsverity is not enabled on target file")]
+    VerityMissing,
+    #[error("Expected digest {expected} but found {found}")]
+    DigestMismatch { expected: String, found: String },
+}
+
+/// Require the fsverity digest to be present.
+pub fn measure_verity_digest<F: AsFd, H: FsVerityHashValue>(
+    fd: F,
+) -> Result<H, CompareVerityError> {
+    match ioctl::fs_ioc_measure_verity::<_, H>(fd)? {
+        Some(found) => Ok(found),
+        None => Err(CompareVerityError::VerityMissing),
+    }
+}
+
+/// Compare the fsverity digest of the file versus the expected digest.
+pub fn ensure_verity<F: AsFd, H: FsVerityHashValue>(
+    fd: F,
+    expected: &H,
+) -> Result<(), CompareVerityError> {
+    let found = measure_verity_digest::<_, H>(fd)?;
+    if expected == &found {
+        Ok(())
+    } else {
+        Err(CompareVerityError::DigestMismatch {
+            expected: hex::encode(expected),
+            found: hex::encode(found.as_ref()),
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::fsverity::Sha256HashValue;
+
+    #[test]
+    fn test_verity_missing() -> anyhow::Result<()> {
+        let tf = tempfile::tempfile()?;
+        match measure_verity_digest::<_, Sha256HashValue>(&tf) {
+            Err(CompareVerityError::VerityMissing) => {}
+            o => panic!("Unexpected {o:?}"),
+        }
+        let h = Sha256HashValue::default();
+        match ensure_verity(&tf, &h) {
+            Err(CompareVerityError::VerityMissing) => {}
+            o => panic!("Unexpected {o:?}"),
+        }
+        Ok(())
+    }
 }
--- a/src/mount.rs
+++ b/src/mount.rs
@@ -14,7 +14,7 @@ use rustix::{
     },
 };
 
-use crate::fsverity;
+use crate::{fsverity, util};
 
 struct FsHandle {
     pub fd: OwnedFd,
@@ -192,11 +192,8 @@ impl<'a> MountOptions<'a> {
         let image = std::fs::File::open(self.image)?;
 
         if let Some(expected) = self.digest {
-            let measured: fsverity::Sha256HashValue =
-                fsverity::ioctl::fs_ioc_measure_verity(&image)?;
-            if expected != hex::encode(measured) {
-                panic!("expected {:?} measured {:?}", expected, measured);
-            }
+            let expected = util::parse_sha256(expected)?;
+            fsverity::ensure_verity(&image, &expected)?;
         }
 
         mount_fd(image, self.basedir, mountpoint)
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -20,9 +20,8 @@ use sha2::{Digest, Sha256};
 
 use crate::{
     fsverity::{
-        digest::FsVerityHasher,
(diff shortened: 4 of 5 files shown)


src/fsverity/ioctl.rs
```diff
@@ -71,8 +75,33 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
     }
 
     if digest.digest_algorithm != digest_algorithm || digest.digest_size != digest_size {
-        Err(std::io::Error::from(std::io::ErrorKind::InvalidData))?
+        Err(Errno::ILSEQ)
```
allisonkarlitskaya, line 78: If we do that then we need to also decide what to do about `EOVERFLOW`.  According to the docs, https://docs.kernel.org/filesystems/fsverity.html#fs-ioc-measure-verity we get `EOVERFLOW` if the digest is longer than specified in the digest_size.  So if we expect sha256 (which we do) but the file has sha512, then we'll fail at the syscall level with `EOVERFLOW` and never make it to this point.

Potential ways forward:
 - always use a larger buffer
 - return `EOVERFLOW` also for digest mismatch detected here
 - detect the `EOVERFLOW` case and turn it into `EILSEQ`.

I personally find `EILSEQ` a fairly weird choice, but `EOVERFLOW` is also sort of strange.  Anything but `EINVAL` :sweat_smile: 

But what about #25?  Maybe this would be a good chance to look into that.  This is Rust, not C: we can do better...

src/fsverity/ioctl.rs
```diff
@@ -71,8 +75,33 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
         Ok(digest.digest)
     }
 }
+
+/// Like [`fs_ioc_measure_verity`] but does not fail if the file does not have fsverity enabled.
+pub fn fs_ioc_measure_verity_optional<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<Option<H>> {
+    match fs_ioc_measure_verity(fd) {
+        Ok(e) => Ok(Some(e)),
+        // See e.g. `lcfs_fd_measure_fsverity()`
+        Err(Errno::NODATA | Errno::NOSYS | Errno::NOTTY) => Ok(None),
```
allisonkarlitskaya, line 89: I really think `ENOSYS` is only gonna show up if `ioctl()` itself is not a supported system call (ie: never).  According to the docs:

> EOPNOTSUPP: the kernel was not configured with fs-verity support, or the filesystem superblock has not had the ‘verity’ feature enabled on it. (See [Filesystem support](https://docs.kernel.org/filesystems/fsverity.html#filesystem-support).)

src/fsverity/ioctl.rs
```diff
@@ -71,8 +75,33 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::fsverity::Sha256HashValue;
+
+    #[test]
+    fn test_measure_verity_opt() -> anyhow::Result<()> {
+        let tf = tempfile::tempfile()?;
+        assert!(fs_ioc_measure_verity_optional::<_, Sha256HashValue>(&tf)
+            .unwrap()
```
allisonkarlitskaya, line 103: `?` here, since you're already in a `Return`-returning function.

src/fsverity/ioctl.rs
```diff
@@ -71,8 +75,33 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
+            .is_none());
```
allisonkarlitskaya, line 104: `assert_eq!()` with `None`?

src/fsverity/ioctl.rs
```diff
@@ -71,8 +75,33 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
     if digest.digest_algorithm != digest_algorithm || digest.digest_size != digest_size {
-        Err(std::io::Error::from(std::io::ErrorKind::InvalidData))?
+        Err(Errno::ILSEQ)
     } else {
         Ok(digest.digest)
     }
 }
+
+/// Like [`fs_ioc_measure_verity`] but does not fail if the file does not have fsverity enabled.
+pub fn fs_ioc_measure_verity_optional<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<Option<H>> {
```
allisonkarlitskaya, line 85: ...or maybe just return `Option<H>>` from the other version of the call, and update the users of that API?  That would let us punt on the errno/anyhow/thiserror thing.

allisonkarlitskaya (changes requested):
Thanks for working on this.  I think we could definitely clean this up, and I'm generally a big fan of `Ok(None)` for the "does not exist" case.

I've often thought that `measure` is usually *not* what we want to do, what what we really want is to verify that the measurement matches a certain expected value.  I've been thinking about adding an API for that and using it from all of the existing callers that do their own comparisons.  Thoughts?

src/fsverity/ioctl.rs
cgwalters: There's a bit of an idiom here - in tests specifically I use `unwrap()` on the APIs we actually want to test because it gives us precise line numbers.

It isn't relevant here, but it is in other places where one might be invoking the API under test multiple times, and if you use `?` then on a test case failure you'd be wondering which specific one failed.

I didn't create this idiom, it came from some Rust code Luca Bruno maintained IIRC. Do other people do it? Probably...

src/fsverity/ioctl.rs
cgwalters: (If we change it here let's change the C composefs code too, I was just mirroring that)

I think you're probably right at a baseline; my suspicion is where some of this `ENOSYS` usage comes in is dealing with seccomp filters.

Dunno, how strongly do you feel? I am OK to drop it in both places, but it feels like a very-low-value, low-but-nonzero-risk case.

Or I guess we could just drop it here, and *not* change the composefs C code, and see if we ever hit it anywhere as this code gets used more widely...

src/fsverity/ioctl.rs
cgwalters: > But what about https://github.com/containers/composefs-rs/issues/25? Maybe this would be a good chance to look into that. This is Rust, not C: we can do better...

I started looking at it (I can push up my WIP branch), but it's quite an infectious thing, although I tried starting from a different place.

I'll take a crack at updating this PR instead and target it at just this code.

cgwalters:
OK, I reworked this.

src/fsverity/ioctl.rs
```diff
@@ -53,7 +69,10 @@ pub struct FsVerityDigest<F> {
 // #define FS_IOC_MEASURE_VERITY   _IORW('f', 134, struct fsverity_digest)
 type FsIocMeasureVerity = ioctl::ReadWriteOpcode<b'f', 134, FsVerityDigest<()>>;
 
-pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H> {
+/// Measure the fsverity digest of the provided file descriptor.
+pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(
```
cgwalters, line 73: Tangential but since this function is much less of a trivial wrapper for the ioctl, I am less certain about having the module named `ioctl` (or less invasively perhaps we could rename this function to just `measure_verity` or something? anyways...)

src/fsverity/ioctl.rs
allisonkarlitskaya: I like this logic.  Thanks for explaining.

allisonkarlitskaya: See also, though, (in case you don't know it) the `backtrace` feature of `anyhow`.

src/fsverity/ioctl.rs
```diff
@@ -1,10 +1,22 @@
...
 
+/// Measuring fsverity failed.
+#[derive(Error, Debug)]
+pub enum MeasureVerityError {
+    #[error("i/o error")]
+    Io(#[from] std::io::Error),
+    #[error("Expected algorithm {expected}, found {found}")]
+    InvalidDigestAlgorithm { expected: u16, found: u16 },
+    #[error("Expected digest size {expected}")]
+    InvalidDigestSize { expected: u16 },
```
allisonkarlitskaya, line 17: I feel like this could be mixed with `InvalidDigestAlgorithm`.  I'm not sure how useful it is to have those integers included there.

src/fsverity/ioctl.rs
allisonkarlitskaya: Not so tangential.  It's actually a good point.

I think it would make sense to add the compare/error stuff maybe at the toplevel `fsverity/mod.rs`?

src/fsverity/ioctl.rs
```diff
@@ -63,16 +82,48 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
+            }
+            if digest.digest_size != digest_size {
+                return Err(MeasureVerityError::InvalidDigestSize {
+                    expected: digest.digest_size,
+                });
+            }
+            Ok(Some(digest.digest))
+        }
+        // This function returns Ok(None) if there's no verity digest found.
+        Err(Errno::NODATA | Errno::NOTTY) => Ok(None),
```
allisonkarlitskaya, line 107: The fsverity docs say `EOPNOTSUPP` for no kernel support.  I guess I'd count that as "file doesn't have fs-verity".

src/fsverity/ioctl.rs
```diff
@@ -63,16 +82,48 @@ pub fn fs_ioc_measure_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<H>
...
+        Err(Errno::OVERFLOW) => Err(MeasureVerityError::InvalidDigestSize {
+            expected: digest.digest_size,
+        }),
+        Err(e) => Err(std::io::Error::from(e).into()),
```
allisonkarlitskaya, line 111: Yikes... is this due to some missing linkage between `rustix` `Errno` and `thiserror`?  Is there a reason you used `std::io::Error` in our error struct instead of `Errno`?  Just to hide rustix as an implementation detail?  (That's a good reason...)

src/fsverity/ioctl.rs
allisonkarlitskaya: Oh.  That's exactly what you did.  Thanks!

src/fsverity/mod.rs
```diff
@@ -19,3 +25,57 @@ impl FsVerityHashValue for Sha512HashValue {
     const ALGORITHM: u8 = 2;
     const EMPTY: Self = [0; 64];
 }
+
+pub fn sha256_from_str(s: &str) -> Result<Sha256HashValue> {
```
allisonkarlitskaya, line 29: I feel like we already have one of those in `util.rs`?

src/fsverity/mod.rs
```diff
@@ -19,3 +25,57 @@ impl FsVerityHashValue for Sha512HashValue {
...
+    VerityMissing,
+    #[error("Expected digest {expected} but found {found}")]
+    DigestMismatch { expected: String, found: String },
+}
+
+/// Compare the fsverity digest of the file versus the expected digest.
+pub fn compare_verity<F: AsFd, H: FsVerityHashValue>(
+    fd: F,
+    expected: &H,
+) -> Result<(), CompareVerityError> {
```
allisonkarlitskaya, line 50:
```suggestion
pub fn compare_verity(fd: impl AsFd, expected: impl FsVerityHashValue) -> Result<(), CompareVerityError> {
```

src/fsverity/mod.rs
```diff
@@ -19,3 +25,57 @@ impl FsVerityHashValue for Sha512HashValue {
...
+            if expected == found {
+                Ok(())
+            } else {
+                Err(CompareVerityError::DigestMismatch {
+                    expected: hex::encode(expected),
+                    found: hex::encode(found.as_ref()),
+                })
+            }
+        }
+        None => Err(CompareVerityError::VerityMissing),
```
allisonkarlitskaya, line 62: I like putting these "early error case" things up top (but I'm super inconsistent about it).  I'm still developing my taste and was wondering if you have any opinion about this stuff...

src/fsverity/mod.rs
```diff
@@ -19,3 +25,57 @@ impl FsVerityHashValue for Sha512HashValue {
...
+    fd: F,
+    expected: &H,
+) -> Result<(), CompareVerityError> {
+    match ioctl::fs_ioc_measure_verity::<_, H>(fd)? {
+        Some(ref found) => {
+            if expected == found {
+                Ok(())
+            } else {
+                Err(CompareVerityError::DigestMismatch {
+                    expected: hex::encode(expected),
```
allisonkarlitskaya, line 57: Cool.  Using strings here is a nice way to avoid the types being awful.

src/mount.rs
```diff
@@ -192,11 +192,8 @@ impl<'a> MountOptions<'a> {
         let image = std::fs::File::open(self.image)?;
 
         if let Some(expected) = self.digest {
-            let measured: fsverity::Sha256HashValue =
-                fsverity::ioctl::fs_ioc_measure_verity(&image)?;
-            if expected != hex::encode(measured) {
-                panic!("expected {:?} measured {:?}", expected, measured);
-            }
+            let expected = sha256_from_str(expected)?;
+            fsverity::compare_verity(&image, &expected)?;
```
allisonkarlitskaya, line 196: Looking at it at a call site I think the "compare" verb is wrong.  I'd expect that for a function with a boolean return value.  We probably want something along the lines of "ensure" "check" "assert" etc. that implies that it's an error if it doesn't match.

src/repository.rs
```diff
@@ -98,8 +99,7 @@ impl Repository {
         fs_ioc_enable_verity::<&OwnedFd, Sha256HashValue>(&ro_fd)?;
 
         // double-check
-        let measured_digest: Sha256HashValue = fs_ioc_measure_verity(&ro_fd)?;
-        assert!(measured_digest == digest);
+        fsverity::compare_verity(&ro_fd, &digest)?;
```
allisonkarlitskaya, line 102: Thanks for de-panicking this.  Could maybe use some `.context()` though (although that's also kinda pointless if we're moving away from `anyhow`...).

src/repository.rs
```diff
@@ -123,12 +123,8 @@ impl Repository {
         expected_verity: &Sha256HashValue,
     ) -> Result<OwnedFd> {
         let fd = self.openat(filename, OFlags::RDONLY)?;
-        let measured_verity: Sha256HashValue = fs_ioc_measure_verity(&fd)?;
-        if measured_verity != *expected_verity {
-            bail!("bad verity!")
-        } else {
-            Ok(fd)
-        }
+        fsverity::compare_verity(&fd, expected_verity)?;
```
allisonkarlitskaya, line 126: So... much... cleanup!

src/repository.rs
```diff
@@ -194,7 +190,8 @@ impl Repository {
     pub fn check_stream(&self, sha256: &Sha256HashValue) -> Result<Option<Sha256HashValue>> {
         match self.openat(&format!("streams/{}", hex::encode(sha256)), OFlags::RDONLY) {
             Ok(stream) => {
-                let measured_verity: Sha256HashValue = fs_ioc_measure_verity(&stream)?;
+                let measured_verity: Sha256HashValue = fs_ioc_measure_verity(&stream)?
```
allisonkarlitskaya, line 193: Low-key surprised to see that we legit only had a single case where "option" wasn't the right API.

src/fsverity/ioctl.rs
cgwalters: Yeah, but by default you don't get backtraces out of `cargo test` either way. Of course you can, this is just all about improving the default case for what happens when a test fails.

src/repository.rs
allisonkarlitskaya: On the other hand, this seems to be the only API user that evaded the 'compare'-ification.  I wonder if `Option` on the base API was wrong here, after all.  Hm.

allisonkarlitskaya (commented):
This looks great.  Thanks for working on this.

The 'compare' thing is such a nice cleanup, and I'm really happy to see the (slow) introduction of `thiserror`.

A lot of the comments below are meant as discussion points.

src/fsverity/ioctl.rs
allisonkarlitskaya: Ya.  I definitely like your approach better.

src/fsverity/ioctl.rs
cgwalters: Yes, if we exposed the rustix one then semver bumps in rustix would become semver bumps for us.

src/fsverity/ioctl.rs
cgwalters: Yeah the chance of anything external using them is probably nil. But if one is going to use thiserror, it's pretty cheap to add this type of stuff and in particular it becomes useful for tests to precisely validate things.

I don't feel strongly though, if you want to drop them we can

src/fsverity/mod.rs
cgwalters: I briefly tried that but I ran into type inference errors on the invocation of `fs_ioc_measure_verity`

src/fsverity/mod.rs
cgwalters: Hmm in `match` my sense of style is that expected comes first more usually.

src/fsverity/ioctl.rs
allisonkarlitskaya: That's good to know, thanks.  I'll make sure to think about that going forward...

src/repository.rs
cgwalters: I added an intermediate function which requires verity

cgwalters:
I pushed some updates, but feel free to take over and push any changes you want tomorrow!

src/fsverity/mod.rs
```diff
@@ -19,3 +25,61 @@ impl FsVerityHashValue for Sha512HashValue {
...
+    DigestMismatch { expected: String, found: String },
+}
+
+/// Require the fsverity digest to be present.
+pub fn measure_verity_digest<F: AsFd, H: FsVerityHashValue>(
+    fd: F,
+) -> Result<H, CompareVerityError> {
+    match ioctl::fs_ioc_measure_verity::<_, H>(fd)? {
+        Some(found) => Ok(found),
+        None => Err(CompareVerityError::VerityMissing),
```
allisonkarlitskaya, line 46: It feels kinda weird how we take a syscall errno, convert it to None, then convert it back to a synthetic error.  But I think that's my C brain talking.

In reality `VerityMissing` is a way more meaningful error than `ENODATA` or `ENOTTY` or whatever random crap might be thrown at us.

allisonkarlitskaya (approved):

allisonkarlitskaya:
Thanks!

PR #50, opened 2024-12-04
fs: Work around rustix listxattr c_char wonkiness

cgwalters:
See https://github.com/bytecodealliance/rustix/pull/1061#issuecomment-2518467534 Basically current rustix 0.38 sometimes may expose the element type as a signed or unsigned variant, depending on the version of linux-raw-sys and rustix feature flags.

Add a cast to work around this; if we cared to drop the unsafe we could use the zerocopy crate.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -136,6 +136,10 @@ impl FilesystemReader<'_> {
         let mut xattrs = BTreeMap::new();
 
         let names_size = listxattr(&filename, &mut [])?;
+        // See https://github.com/bytecodealliance/rustix/pull/1061#issuecomment-2518467534
+        // Basically current rustix 0.38 sometimes may expose the element type as a signed
+        // or unsigned variant, depending on the version of linux-raw-sys and rustix feature
+        // flags.
         let mut names = vec![0; names_size];
         let actual_names_size = listxattr(&filename, &mut names)?;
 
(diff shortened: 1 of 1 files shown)


src/fs.rs
```diff
@@ -136,6 +136,10 @@ impl FilesystemReader<'_> {
         let mut xattrs = BTreeMap::new();
 
         let names_size = listxattr(&filename, &mut [])?;
+        // See https://github.com/bytecodealliance/rustix/pull/1061#issuecomment-2518467534
+        // Basically current rustix 0.38 sometimes may expose the element type as a signed
+        // or unsigned variant, depending on the version of linux-raw-sys and rustix feature
+        // flags.
         let mut names = vec![0; names_size];
```
cgwalters, line 143: A confusing thing here is that type inference is kicking in for the `0` here and we had conflicting requirements before in the wrong case. It'd help to be able to explicitly specify this array as the target type but I don't think rustix exposes that type alias right now even though it probably should because it's part of the public API.

src/fs.rs
```diff
@@ -149,6 +153,8 @@ impl FilesystemReader<'_> {
 
         let mut buffer = [0; 65536];
         for name in names.split_inclusive(|c| *c == 0) {
+            // SAFETY: casting i8 to u8 is safe.
+            let name: &[u8] = unsafe { std::mem::transmute(name) };
```
allisonkarlitskaya, line 157: I really wanted to avoid `unsafe` but my old `//fml` solution was causing warnings in the case that `c_char` ended up already being `u8`.  I guess this one won't.

I also expect this to be relatively short-lived until rustix figures out the solution here...

allisonkarlitskaya (approved):

PR #52, opened 2024-12-07
build-sys: Lower rust-version to 1.82

cgwalters:
This is what's in C9S (a primary dev target for me) at the moment. The claimed rationale for the bump around lifetime elison rules doesn't seem correct (this builds just fine for me with 1.82).

I suspect it's more newer clippy has newer suggestions.


allisonkarlitskaya (approved):
Fine by me.  Thanks!

PR #53, opened 2024-12-07
tree-wide: Add Debug impls

cgwalters:
I find this really useful as a general baseline; I'm a "printf debugger" person by default. Add derives tree wide (plus two manual impls). Enable deny-by-default lint.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,5 @@
+#![deny(missing_debug_implementations)]
+
 pub mod dumpfile;
 pub mod dumpfile_parse;
 pub mod fs;
--- a/src/splitstream.rs
+++ b/src/splitstream.rs
@@ -15,11 +15,13 @@ use crate::{
     util::read_exactish,
 };
 
+#[derive(Debug)]
 pub struct DigestMapEntry {
     pub body: Sha256HashValue,
     pub verity: Sha256HashValue,
 }
 
+#[derive(Debug)]
 pub struct DigestMap {
     pub map: Vec<DigestMapEntry>,
 }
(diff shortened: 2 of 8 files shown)


shawn111:
Do you consider add tokio tracing?

src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(missing_debug_implementations)]
```
allisonkarlitskaya, line 1: Hardcore.

allisonkarlitskaya (commented):
I generally like the idea.  Is this to make it easier to add quick debugging statements here and there?

I've been looking at making more use of the `log` crate as well (like in my erofs branch).

Is there a way to get this only enabled on debug builds?  I've read that the code that this generates is a bit expensive, but it would be kinda nice to find out if any `{:?}` slips into production code.  I understand that's supposed to be an anti-pattern...

src/lib.rs
cgwalters: > Is there a way to get this only enabled on debug builds? 

Not that I know of.

> I've read that the code that this generates is a bit expensive,

Expensive...for compilation speed? Runtime? I think it used to in some cases generate a lot of code which would slow down compilation a bit, but I think it's really well optimized nowadays (it's builtin to the compiler).

https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/fmt-debug.html exists (linked from https://github.com/johnthagen/min-sized-rust ) but I think you pretty much have to be writing `#[no_std]` to care is my impression, and we are quite far away from that.

>  but it would be kinda nice to find out if any {:?} slips into production code.

I have never had an issue with that, you have to pretty explicitly type `{:?}`, it doesn't exactly flow off the fingertips :smile:

allisonkarlitskaya (approved):

src/splitstream.rs
```diff
@@ -63,6 +65,17 @@ pub struct SplitStreamWriter<'a> {
     pub sha256: Option<(Sha256, Sha256HashValue)>,
 }
 
+impl std::fmt::Debug for SplitStreamWriter<'_> {
```
allisonkarlitskaya, line 68: Thanks for the extra effort here.

PR #54, opened 2024-12-08
tests: Use similar-asserts

cgwalters:
I was using this in the previous composefs rust code and I find it useful. Prep for more testing work.


allisonkarlitskaya (approved):
This looks really cool, thanks!

PR #57, opened 2024-12-16
src: add internal erofs writer code

### LINKED ISSUE #56: Add an internal erofs writer implementation
We're starting to get increasing friction with using `mkcomposefs`, mostly related to Debian/Ubuntu:
 - we have to install it from git in GitHub Actions because Ubuntu doesn't have a package
 - for the same reason, it's blocking #45
 - it would be nice if we could take our main rust workflow back out of the container and use it directly on GitHub

At the same time I've been worried about the specification of the `containers.composefs.fsverity` attribute, since it's basically currently defined in terms of "run `mkcomposefs` and measure the output".  Particularly because erofs is rather sparsely documented...

As such, a dual goal:
 - add an in-tree erofs writer
 - produce a detailed document describing how to create a composefs erofs image such that someone doing an independent implementation would have a decent chance of getting the same output as us

I have a WIP branch here, and have made quite a bit of progress: https://github.com/allisonkarlitskaya/composefs-rs/tree/erofs/src/erofs

I think it would be extremely nice if this writer produced the same output as `libcomposefs`.  Testing the two implementations against each other would be amazing.

We should try to do one of these two things:
 - make our erofs writer exactly compatible with the erofs writer in composefs
 - use our independent re-implementation as input to the discussion in https://github.com/containers/composefs/issues/198 with the goal of modifying `libcomposefs` to output a new format

Or, maybe some mix of the two.

allisonkarlitskaya:
This introduces experimental code for writing erofs images ourselves, instead of using the external mkcomposefs CLI.

It's currently disabled by default.  You can test it by setting the `COMPOSEFS_FORMAT=new` environment variable.

This currently produces a different output than the output of mkcomposefs, which is why it's gated behind an environment variable. The plan is to add a compatibility mode to our internal writer code so that it produces as similar of an output as possible and then switch over to using it once we are convinced that it's equivalent.  Then the `COMPOSEFS_FORMAT=` variable will disable this compatibility mode.

There's also a `COMPOSEFS_DUMP_EROFS=1` environment variable (which works with both `mkcomposefs` and our internal code) which will dump the erofs layout for diffing.  There's also a standalone `erofs-debug` binary that will do the same.

Additionally, this introduces two new files in docs:
- a detailed description of the parts of erofs that we use
- a document which attempts to describe the decisions made in creating an erofs composefs image (in terms of which order the files are in, etc).

The main idea here is to start a serious effort towards standardizing the composefs label we want to start adding to container images: it should be possible to define what will be in that label by way of documentation instead of saying "run this software and use the output". These two new documents, taken together with the existing "oci.md" form a rough (and still incomplete) outline for that.

Many thanks to Gao Xiang <xiang@kernel.org> for helping clarify many points about the erofs file format for the documentation.

Closes #56


allisonkarlitskaya:
cc @hsiangkao

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+    -   `SB_CHKSUM` (`0x0001`): set if the checksum field in the superblock is
+        populated.  Otherwise, the checksum is ignored.
+    -   `MTIME` (`0x0002`): at first, erofs named the timestamp fields `ctime`
+        instead of `mtime`.  That got changed a long time ago, and this flag
+        got added to indicate filesystems that were created with the new
+        semantics. This flag has absolutely zero impact at run time: the kernel
+        ignores it.
+    -   `XATTR_FILTER` (`0x0004`): set if the xattr bloom filter should be
+        used.  Read about this in the inode section.
+*   `blkszbits`: log2 of the block size.  Better set this to 12 (4096).
```
hsiangkao, line 50: yes, also it may be helpful to indicate that it's useful to make the image mountable on any page sizes.

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+*   `root_nid`: the reference to the root inode.  See the inodes section for
+    what that means.  Normally inodes are stored in u64, but this is somewhat
+    randomly a u16, which means that you're gonna need to put the root
```
hsiangkao, line 53: For our initial use cases, `__le16` is enough, however for the recent incremental builds, if the root inode is updated and outplaced update is applied, it may not be enough.
Currently, it's been worked around to inplace update the root inode, but I will find a chance to extend it as 64 bits later.

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+
+A rudimentary form of compression is supported on xattr names.  There are a
+number of hardcoded "common prefixes" defined with the `EROFS_XATTR_INDEX_`
+constants in `erofs_fs.h`.  Confusingly, although `LUSTRE` is present, it's not
+wired up in the kernel.  Don't use that one.
+
+The basic idea is that you find the prefix for your xattr from the list (like
+`user.` or `security.`) and then you store only the "suffix" part, along with
+the prefix index.  If you can't find a prefix, you use 0 (which is conceptually
+a prefix of "").  If the prefix matches the entire name then the suffix is `""`.
```
hsiangkao, line 91: We could define `0` as a valid prefix number, although 0 will be ignored and cannot be listed at all in kernel.

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+
+Note: you really need to do this "compression" step, because it's assumed
+during the lookup phase.  ie: if we're looking for an xattr `"user.xyz"` then
+we'll only consider the entries that have the prefix index for `user.` set on
+them.  If you didn't properly "compress" your xattr names, they won't be found.
```
hsiangkao, line 96: Yes, that is due to `0` will be always ignored by the kernel.

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+
+There's support in the erofs format for custom prefixes.  That's when the high
+bit of the prefix index is set.  These got added circa kernel version 6.4 with
+a patch series ending with `6a318ccd7e08` ("erofs: enable long extended
+attribute name prefixes") but aren't documented here because we don't use them.
```
hsiangkao, line 101: Anyway, I guess you could consider to use them since this feature is landed before overlayfs `data-only` layer feature, so if the new on-disk format uses `data-only` layer feature, I guess you could safely use it too.

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+
+The hash algorithm is xxh32.  The thing that gets hashed is not the name, but
+the "suffix" that's left after removing the prefix.  The seed is
+`EROFS_XATTR_FILTER_SEED` plus the prefix index.  The lower 5 bits of the hash
+value (0..31) are used to determine which bit is used.
+
+For some reason a bit value of 1 here indicates the absence of a particular
+xattr, which is opposite to the usual arrangement.  You'd think it was for
+compatibility, but the filter is only engaged if the feature bit is present in
+the superblock.
```
hsiangkao, line 233: that is due to incremental builds too, because some on-disk inodes could be previously generated and we don't need to touch them to do an incremental build

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+separate block (ie: the block directly following the inode data).  It is also
+permitted for the inode data itself to cross block boundaries.  There are a
+couple of caveats to be aware of, however:
+*   the alignment of inodes is 32 bytes, but the size of an extended inode is 64
+    bytes.  `mkfs.erofs` tries to ensure that extended inodes headers land
+    entirely within on disk block (for efficiency), but this isn't required by
+    the kernel.
+*   `mkfs.erofs` also tries to ensure that the inline data ends in the same
+    disk block as the last byte of the inode metadata (ie: inode header plus
+    xattrs).  This is theoretically not required by the kernel.
```
hsiangkao, line 290: yes, but the whole concept of inlined data is to save extra i/o as well as storage. Although the kernel supports it but it's never a design goal of `tail inline` inodes...

doc/erofs.md
```diff
@@ -0,0 +1,431 @@
...
+*   A bug present in the kernel before 6.12 meant that this was required for
+    inline symlink targets. This was fixed by `9ed50b8231e3` ("erofs: fix
+    incorrect symlink detection in fast symlink").
+*   In general, when faced with the task of writing out an inode with inline
+    data present, you may need to add padding bytes before the start of the
+    inode in order to ensure that the inline data falls within a single block.
```
hsiangkao, line 296: but the negative impacts of `tail inline` is that
previous in principle you could read `64` (4096/64) inodes at once for each metadata 4k i/o,
now you can only read 1 or 2 inodes if the tail inline is too huge.
It could cause some impacts to `ls` workload (because it just read metadata), so I tend to just inline _small_ data.

doc/erofs.md
allisonkarlitskaya: Right!  Makes sense, thanks!

doc/erofs.md
allisonkarlitskaya: I've been thinking about this IO amplification question a bit.  I have basically zero understanding of how filesystems actually work in the kernel, so this is possibly quite naive.

Maybe thinking about how much IO you need to do to load a single inode is a bit of a skewed metric.  I imagine that (during boot, for example) more or less the entire inode table is going to end up paged in and cached.  And in that case the best way to reduce IO would be to reduce the number of pages that need to be loaded, by using the least amount of padding possible, even if it hurts the alignment of a single inode. 

Do you have benchmarks or anything like that in this direction?

doc/erofs.md
allisonkarlitskaya: This is a pretty good point.

doc/erofs.md
hsiangkao: > I've been thinking about this IO amplification question a bit. I have basically zero understanding of how filesystems actually work in the kernel, so this is possibly quite naive.
> 
> Maybe thinking about how much IO you need to do to load a single inode is a bit of a skewed metric. I imagine that (during boot, for example) more or less the entire inode table is going to end up paged in and cached. And in that case the best way to reduce IO would be to reduce the number of pages that need to be loaded, by using the least amount of padding possible, even if it hurts the alignment of a single inode.
> 
> Do you have benchmarks or anything like that in this direction?

In principle, the kernel doesn't do any policy strategies, because it might be good for your workload, but not good for others (e.g. users just would like to read specific inodes).  So you might win for some cases, but lose for other useful cases.  It's not quite something to serve for generic use cases.

EROFS always reads needed block-sized metadata as other kernel fses.  if you need to warm up metadata, you could use fadvise in advance.

doc/erofs.md
allisonkarlitskaya: Keep in mind, though, that for composefs, this is only about directories.  We inline symlinks unconditionally but they're usually very small.  As for regular files, we only inline those when they're less than 32 bytes: otherwise they're handled as redirects to the data layer.

Being about directories it again turns into a question of the cost of a single operation vs the value of reducing the overall amount of IO (because adjacent operations effectively end up prepopulating the cache for future operations when things are more tightly packed).  Imagine the metric is `find /` instead of `ls` for example.

Unlike the question of padding between inodes, though, my gut feeling is that the question could go in the other direction here: wasting more space (in the blocks sections) helps keep the inode table smaller, which helps load it faster and keep it in the cache....

This is again where it would be nice to say something about benchmarks...

allisonkarlitskaya: What's your gut feeling about the appropriate number for inlining directories?  or: at what point do we do another block?

Oh: on that topic there is a slight ambiguity here that I never resolved.  Let's say I'm writing a small directory in 100 bytes and I put it into a single block.  What is the size field?  100 or 4096?  If it was inline it would obviously be 100.  This has a potential impact in knowing where the name of the last entry is terminated (end of file vs scanning for nul).  I'd say 100 but composefs currently does 4096.

Does it matter at all?

hsiangkao: > Keep in mind, though, that for composefs, this is only about directories. We inline symlinks unconditionally but they're usually very small. As for regular files, we only inline those when they're less than 32 bytes: otherwise they're handled as redirects to the data layer.
> 
> Being about directories it again turns into a question of the cost of a single operation vs the value of reducing the overall amount of IO (because adjacent operations effectively end up prepopulating the cache for future operations when things are more tightly packed). Imagine the metric is `find /` instead of `ls` for example.
> 
> Unlike the question of padding between inodes, though, my gut feeling is that the question could go in the other direction here: wasting more space (in the blocks sections) helps keep the inode table smaller, which helps load it faster and keep it in the cache....
> 
> This is again where it would be nice to say something about benchmarks...

So from the on-disk placement, there is no sliver bullet too. But I guess you could also consider fadvise warm up.

hsiangkao: > What's your gut feeling about the appropriate number for inlining directories? or: at what point do we do another block?
> 
> Oh: on that topic there is a slight ambiguity here that I never resolved. Let's say I'm writing a small directory in 100 bytes and I put it into a single block. What is the size field? 100 or 4096? If it was inline it would obviously be 100. This has a potential impact in knowing where the name of the last entry is terminated (end of file vs scanning for nul). I'd say 100 but composefs currently does 4096.
> 
> Does it matter at all?

It should be 100, 4096 might work if the trailing filename has `\0` but it's not recommended...

doc/erofs.md
allisonkarlitskaya: Can you please clarify what you meant by this?  Should the block size be larger or smaller than the page size in order to be able to mount it?

doc/erofs.md
allisonkarlitskaya: That's good to know.  I think I didn't fully understand how the "prefix" part of xattrs is so special when I was writing this...

doc/erofs.md
hsiangkao: on the modern platforms, the smallist page size is 4KiB. EROFS has supported < page size blocks for all platforms now. So leave it as 4KiB will keep images mountable all the time..

doc/image-format.md
```diff
@@ -0,0 +1,276 @@
...
+
+Ultimately, the erofs image that we produce needs to be used as a layer in an
+overlayfs stack.  There are a lot of cases where the thing that we write out
+only makes sense to overlayfs.  There are other cases where we need to avoiding
+writing out things that overlayfs would treat as "special".
+
+`libcomposefs` writes 256 files named from `00` to `ff` into the root directory
+as character devices with major/minor of (0, 0).  Those are overlayfs whiteouts
+and they are needed for older versions of overlayfs which don't support "data
+only" layers.  We don't target these versions, so *we don't add these files*.
```
cgwalters, line 97: Here and almost everywhere else...you use "we" to mean this codebase only but I think it'd be more helpful for us to consider "we" to also include the C composefs project.

So we'd rephrase this to something more like "1.1 format doesn't have these files" or so?

src/erofs/debug.rs
```diff
@@ -0,0 +1,377 @@
...
+
+use super::{
+    format::{self, CompactInodeHeader, ComposefsHeader, ExtendedInodeHeader, Superblock},
+    reader::{DirectoryBlock, Image, Inode, InodeHeader, InodeOps, InodeType, InodeXAttrs, XAttr},
+};
+
+macro_rules! print_fields {
+    ($ty: ty, $s: expr, $f: ident) => {{
+        let value = &$s.$f;
+        let default = if false { value } else { &Default::default() };
```
cgwalters, line 20: Confused by this

src/erofs/reader.rs
```diff
@@ -0,0 +1,460 @@
...
+use std::ops::Range;
+
+use zerocopy::{little_endian::U32, Immutable, KnownLayout, TryFromBytes};
+
+use super::format::{
+    CompactInodeHeader, ComposefsHeader, DataLayout, DirectoryEntryHeader, ExtendedInodeHeader,
+    FileType, InodeXAttrHeader, Superblock, XAttrHeader,
+};
+
+fn round_up(n: usize, to: usize) -> usize {
```
cgwalters, line 11: Second copy of this?

src/erofs/format.rs
```diff
@@ -0,0 +1,279 @@
...
+    b"trusted.",
+    b"lustre.",
+    b"security.",
+];
+
+/* Directories */
+
+#[derive(Clone, Copy, Debug, Default, Immutable, IntoBytes, TryFromBytes)]
+#[repr(u8)]
+pub enum FileType {
```
cgwalters, line 237: Why not use https://docs.rs/rustix/latest/rustix/fs/enum.FileType.html ?

Ohh, is it for the `IntoBytes`?

src/erofs/format.rs
```diff
@@ -0,0 +1,279 @@
...
+/* Superblock */
+
+pub const MAGIC_V1: U32 = U32::new(0xE0F5E1E2);
+pub const FEATURE_COMPAT_MTIME: U32 = U32::new(2);
+pub const FEATURE_COMPAT_XATTR_FILTER: U32 = U32::new(4);
+
+#[derive(Debug, Default, Immutable, IntoBytes, KnownLayout, TryFromBytes)]
+#[repr(C)]
+pub struct Superblock {
+    // vertical whitespace every 16 bytes (hexdump-friendly)
```
cgwalters, line 44: This comment seems misplaced?

cgwalters (commented):
A lot of code here obviously...I only skimmed it. I hope we can keep on iterating towards merging some of this work with the C composefs (especially docs) and aim towards interop.

src/erofs/format.rs
allisonkarlitskaya: Yes.  This is a mapping of integer values used by the erofs format.

src/erofs/debug.rs
allisonkarlitskaya: I ended up killing that code, but the idea here is that we need to coerce Default::default() to give us the correct type without using `<>` which we can't do in the macro because we have no static way to get the type.  We can't just do `!=` because this doesn't require that both sides have the same type.

src/erofs/format.rs
allisonkarlitskaya: It just means that there's an extra newline every time we hit a 16 byte boundary in the struct.  That makes it easier to see which part of the struct is which when looking at a hexdump.

It's all a bit redundant now, though, since we have a decent debugging tool.

src/erofs/debug.rs
```diff
@@ -0,0 +1,377 @@
...
+                hexdump(block);
+            }
+        }
+        println!();
+
+        offset = start + segment.size();
+    }
+    if offset < data.len() {
+        println!("{offset:08x} Padding");
+        let padding = &data[offset..data.len()];
```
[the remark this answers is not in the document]
allisonkarlitskaya: Thanks!

PR #58, opened 2025-01-10
tests: use `test_with` and skip tests that need `mkcomposefs`

mvo5:
This commit adds the `test_with` crate and skips the tests that require `mkcomposefs`.

It uses use the minimal configuration of `tests_with` and with that we now pull in 229 `cargo tree` lines vs the pervious 217 (fwiw, the default configuration of tests_with incrases the crates from 217 to 497). Even the extra 12 might be too much for this tiny feature, feel free to close again if you feel this is not worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ zstd = "0.13.2"
 
 [dev-dependencies]
 similar-asserts = "1.6.0"
+test-with = { version = "0.14", default-features = false, features = ["executable"] }
 
 [profile.dev.package.sha2]
 # this is *really* slow otherwise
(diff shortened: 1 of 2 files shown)


cgwalters:
Also missing [DCO](https://github.com/containers/composefs-rs/pull/58/checks?check_run_id=35439100335)

cgwalters:
> Even the extra 12 might be too much for this tiny feature, feel free to close again if you feel this is not worth it.

Eh, the core deps of that crate look sane to me

allisonkarlitskaya (commented):
This looks cool, thanks!

I don't worry too much about dev-dependencies...

I agree about locking down the import version, though.

allisonkarlitskaya (approved):
Thanks!

PR #67, opened 2025-01-31
use `composefs:[digest]` as the mount source, remove `mount` binary

allisonkarlitskaya:


allisonkarlitskaya:
CI is failing due to broken mirrors.  I've already retried a couple of times, so it looks like we're going to have to wait... :(

src/mount.rs
```diff
@@ -74,12 +74,13 @@ fn proc_self_fd<A: AsFd>(fd: A) -> String {
...
 }
 
-pub fn composefs_fsmount(image: impl AsFd, basedir: &Path) -> Result<OwnedFd> {
+pub fn composefs_fsmount(image: impl AsFd, name: &str, basedir: &Path) -> Result<OwnedFd> {
     let erofs = FsHandle::open("erofs")?;
     fsconfig_set_string(erofs.as_fd(), "source", proc_self_fd(&image))?;
     fsconfig_create(erofs.as_fd())?;
 
     let overlayfs = FsHandle::open("overlay")?;
+    fsconfig_set_string(overlayfs.as_fd(), "source", format!("composefs:{name}"))?;
```
cgwalters, line 83: This is different from what's in composefs-c since https://github.com/containers/composefs/pull/303

I don't have a really strong opinion on this, but it does seem like it'd be good to be consistent.

Offhand wouldn't including the digest bloat the output of `findmnt` by quite a bit? I use that often.

It feels to me that what we really want here more is a way to reliably retrieve structured metadata associated with an overlayfs, like xattrs on the mount point fd or something? This metadata issue also relates to https://github.com/containers/composefs/issues/280

allisonkarlitskaya: This is different from composefs-c indeed, and it does completely bloat findmnt output, causing it to split awkwardly over two lines (but not when using it with pipes, fortunately).

composefs-c writes "composefs" here.  I find that kinda insufficient though.  If we could replace the fstype field then I'd write "composefs" there.  The source of the filesystem should identify the underlying place the filesystem comes from... and it's definitely useful to be able to know (via somewhat standard tooling) which composefs image ID is under a given mount.  We could say "look in /proc/cmdline" but a) that might not be obvious, and b) what about composefs instances used for non-root filesystem?

If we could find another way to do this, I'd be open to it.  As a totally random aside: I've often wondered if we could rather use base64 or something to reduce the size of our object filenames.  That would apply to image files as well then, obviously...

We could maybe change the C version if you feel that consistency is important here.  I feel like this is definitely a useful feature.

allisonkarlitskaya: Actually, it's kinda not applicable to the C version: there isn't as strong of a policy there that the name of the file should be equal to the fs-verity digest... so showing the mount filename isn't as useful.

We could show the digest in the case that it was provided, I suppose... but that's sort of out of scope of this PR in composefs-rs.

cgwalters: > and it does completely bloat findmnt output, 

Now that I look at least on ostree systems we commonly have bind mounts set up, and `findmnt` prints the full path for those which would generally dwarf this checksum, so perhaps it doesn't matter.

> If we could find another way to do this, I'd be open to it. 

I had a vague memory that util-linux was doing something here, and yeah from `man 8 mount`:

>       x-*
>           The same as X-* options, but stored permanently in user space. This means the options are also available for umount(8) or other operations.

It's not documented AFAICS how it works but from a quick `strace -f mount -t tmpfs tmpfs -o nodev,nosuid,x-foo=bar /var/mnt/t` the `x-foo=bar` gets written to `/run/mount/utab`. And I do get the data back from `findmnt -m` but for some reason not `findmnt`?

So that is one option, and probably a reasonable one - use util-linux's code to do the mount and emit structured data like `x-digest=` etc.

---

However I do have another thought...I think the most canonical data we want here is not the fsverity digest but a way to find the erofs. We could use `name_to_handle_at` to get a file handle, and encode that? From a quick test here the file handle is 20 encoded bytes, smaller than a sha256 at least.

And from the file handle, one can use `open_by_handle_at` to get back a fd for it, and from that we can get the fsverity digest - but we could also just generally inspect the erofs which is probably useful.

allisonkarlitskaya: I'm not crazy about the handle thing.  it's a little bit too exotic.  Another option would just be the filename of the erofs but this will be even longer...

Issue #59, opened 2025-01-17
Add better error reporting for common cases


allisonkarlitskaya:
We have a few people trying out the examples lately and they tend to run into a couple of common issues for which we have terrible error messages:
 - can't enable fs-verity because the FS doesn't support it
 - skopeo isn't installed

The first one is 100% on us: we need a better error for that.  The second one is arguably something to be fixed in containers-image-proxy-rs.

See also #25

allisonkarlitskaya:
Silly github closed this automatically...

PR #68, opened 2025-02-14
fsverity: introduce EnableVerifyError::FilesystemNotSupported error

mvo5:
This commit improves the UX of the error reporting when enabling fs-verity fails because the underlying filesystem does not support this. Here we currently just error with:
```
Inappropriate ioctl for device"
```
with the new code we error with:
```
Filesystem not supported by fs-verity
```

Note that the test relies on the fact that /dev/shm is not supported by fs-verity.

Closes: https://github.com/containers/composefs-rs/issues/59

### DIFF
--- a/src/fsverity/ioctl.rs
+++ b/src/fsverity/ioctl.rs
@@ -17,6 +17,15 @@ pub enum MeasureVerityError {
     InvalidDigestSize { expected: u16 },
 }
 
+/// Enabling fsverity failed.
+#[derive(Error, Debug, PartialEq)]
+pub enum EnableVerifyError {
+    #[error("Filesystem does not support fs-verity")]
+    FilesystemNotSupported,
+    #[error("{0}")]
+    Errno(#[from] rustix::io::Errno),
+}
+
 // See /usr/include/linux/fsverity.h
 #[repr(C)]
 #[derive(Debug)]
@@ -38,9 +47,9 @@ type FsIocEnableVerity = ioctl::WriteOpcode<b'f', 133, FsVerityEnableArg>;
 /// Enable fsverity on the target file. This is a thin safe wrapper for the underlying base `ioctl`
 /// and hence all constraints apply such as requiring the file descriptor to already be `O_RDONLY`
 /// etc.
-pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> std::io::Result<()> {
+pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> Result<(), EnableVerifyError> {
     unsafe {
-        ioctl::ioctl(
+        match ioctl::ioctl(
             fd,
             ioctl::Setter::<FsIocEnableVerity, FsVerityEnableArg>::new(FsVerityEnableArg {
                 version: 1,
@@ -53,7 +62,13 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> std::io::Re
                 sig_ptr: 0,
                 __reserved2: [0; 11],
             }),
-        )?;
+        ) {
+            Err(rustix::io::Errno::NOTTY) | Err(rustix::io::Errno::OPNOTSUPP) => {
+                return Err(EnableVerifyError::FilesystemNotSupported)
+            }
+            Err(e) => return Err(e.into()),
+            Ok(_) => (),
+        }
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


src/fsverity/ioctl.rs
```diff
@@ -17,6 +17,15 @@ pub enum MeasureVerityError {
     InvalidDigestSize { expected: u16 },
 }
 
+/// Enabling fsverity failed.
+#[derive(Error, Debug, PartialEq)]
+pub enum EnableVerifyError {
+    #[error("Filesystem not supported by fs-verity")]
```
allisonkarlitskaya, line 23: I'd describe the situation with slightly different language: the filesystem doesn't support fs-verity.  The enablement code is in the filesystem, not in fs-verity.

src/fsverity/ioctl.rs
```diff
@@ -35,12 +44,16 @@ pub struct FsVerityEnableArg {
 // #define FS_IOC_ENABLE_VERITY    _IOW('f', 133, struct fsverity_enable_arg)
 type FsIocEnableVerity = ioctl::WriteOpcode<b'f', 133, FsVerityEnableArg>;
 
+// ENOTTY 25 Inappropriate ioctl for device
+// XXX: should we use an errno crate instead?
+static ENOTTY: i32 = 25;
```
allisonkarlitskaya, line 49: Are we missing that in `rustix::io::Errno`?

https://docs.rs/rustix/latest/rustix/io/struct.Errno.html#associatedconstant.NOTTY

src/fsverity/ioctl.rs
```diff
@@ -53,7 +66,13 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> std::io::Re
                 sig_ptr: 0,
                 __reserved2: [0; 11],
             }),
-        )?;
+        ) {
+            Err(e) if e.raw_os_error() == ENOTTY => {
```
allisonkarlitskaya, line 70: It should be possible to match directly on `Err(rustix::io::Errno::ENOTTY)` here, no?

src/fsverity/ioctl.rs
```diff
@@ -17,6 +17,15 @@ pub enum MeasureVerityError {
     InvalidDigestSize { expected: u16 },
 }
 
+/// Enabling fsverity failed.
+#[derive(Error, Debug, PartialEq)]
+pub enum EnableVerifyError {
```
allisonkarlitskaya, line 22: I wonder if this might be a bit short-sighted.  Currently this is no problem because it gets smoothed out by anyhow, but in the future if we want to start reporting errors from high-level operations on the repository, maybe we want some unified error type for that?

On the other hand, we already have MeasureVerityError and we can chain those things via `#from` in the unified error type, I suppose.

Also: we never said we won't break API, anyway!

allisonkarlitskaya (changes requested):
Thanks for this!

Can you also add some `.context()` at the spot where we enable fs-verity from `Repository`?  Maybe mentioning the repository path (or even complete path to the object file) would be helpful as well, at that point.

src/fsverity/ioctl.rs
mvo5: Yeah, makes sense, I updated this.

src/fsverity/ioctl.rs
mvo5: Indeed, silly me! Thanks and I simplified this now as suggested.

mvo5:
Sorry, I'm a Muppet - I fixed this right away based on the comments and then failed to foce push :(

src/fsverity/ioctl.rs
```diff
@@ -53,7 +62,11 @@ pub fn fs_ioc_enable_verity<F: AsFd, H: FsVerityHashValue>(fd: F) -> std::io::Re
                 sig_ptr: 0,
                 __reserved2: [0; 11],
             }),
-        )?;
+        ) {
+            Err(rustix::io::Errno::NOTTY) => return Err(EnableVerifyError::FilesystemNotSupported),
```
allisonkarlitskaya, line 66: Sorry for not mentioning that earlier, but there's also the case where the filesystem implementation supports the ioctl but the filesystem itself doesn't have it enabled.  This is the case for ext4, for example, where the `-O verity` flag was not specified to `mkfs.ext4` or `tune2fs`.  That's `EOPNOTSUPP`, I think:

https://www.kernel.org/doc/html/latest/filesystems/fsverity.html#fs-ioc-enable-verity

src/fsverity/ioctl.rs
```diff
@@ -17,6 +17,15 @@ pub enum MeasureVerityError {
     InvalidDigestSize { expected: u16 },
 }
 
+/// Enabling fsverity failed.
+#[derive(Error, Debug, PartialEq)]
+pub enum EnableVerifyError {
+    #[error("Filesystem does not supported fs-verity")]
```
allisonkarlitskaya, line 23: "supported" -> "support"

src/fsverity/ioctl.rs
mvo5: Nice catch, thank you! I added this now and started writing a test but the test got a bit involved because it needs to create an ext4 image without verity, mount it, write a file on it and enable verity on it (which requires root to mount the ext4 image). So I did not incude it but I can clean it up and push that too if desired (it will probably need something like the mount.rs:TmpMount to cleanly umount on drop).

allisonkarlitskaya (approved):
tvm!

PR #71, opened 2025-03-06
Issue clean up tar header handling

Johan-Liebert1:
- Use the tar crate for parsing archive entries
- Update tar crate to `0.4.44` from `0.4.42`

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -11,13 +11,15 @@ use anyhow::{bail, Context, Result};
 
 use crate::fsverity::Sha256HashValue;
 
+pub type StatXattrs = BTreeMap<Box<OsStr>, Box<[u8]>>;
+
 #[derive(Debug)]
 pub struct Stat {
     pub st_mode: u32,
     pub st_uid: u32,
     pub st_gid: u32,
     pub st_mtim_sec: i64,
-    pub xattrs: RefCell<BTreeMap<Box<OsStr>, Box<[u8]>>>,
+    pub xattrs: RefCell<StatXattrs>,
 }
 
 #[derive(Debug)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,5 @@
 #![deny(missing_debug_implementations)]
+#![allow(clippy::needless_return)]
 
 pub mod dumpfile;
 pub mod dumpfile_parse;
@@ -11,6 +12,7 @@ pub mod oci;
 pub mod repository;
 pub mod selabel;
 pub mod splitstream;
+pub mod trace;
 pub mod util;
 
 /// All files that contain 64 or fewer bytes (size <= INLINE_CONTENT_MAX) should be stored inline
--- a/src/oci/tar.rs
+++ b/src/oci/tar.rs
@@ -4,20 +4,19 @@ use std::{
     ffi::{OsStr, OsString},
     fmt,
     io::Read,
-    os::unix::prelude::{OsStrExt, OsStringExt},
     path::PathBuf,
 };
 
-use anyhow::{bail, ensure, Result};
+use anyhow::{bail, Result};
 use rustix::fs::makedev;
-use tar::{Entry, EntryType, Header, PaxExtensions};
+use tar::{Entry, EntryType, Header};
 use tokio::io::{AsyncRead, AsyncReadExt};
 
 use crate::{
-    dumpfile, etrace,
+    dumpfile,
     fsverity::{FsVerityHashValue, Sha256HashValue},
     image::{LeafContent, Stat, StatXattrs},
-    splitstream::{SplitStreamData, SplitStreamReader, SplitStreamWriter},
+    splitstream::{SplitStreamReader, SplitStreamWriter},
     util::{read_exactish, read_exactish_async},
     INLINE_CONTENT_MAX,
 };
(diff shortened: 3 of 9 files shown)


src/image.rs
```diff
@@ -11,13 +11,43 @@ use anyhow::{bail, Context, Result};
...
+            st_gid: u32::MAX,
+            st_mtim_sec: -1,
+            xattrs: RefCell::new(BTreeMap::new()),
+        }
+    }
+}
+
+// Might be a bit weird as xattrs cannot be passed as a value here
+// TODO: Turn this into a simple `impl`
+impl TryFrom<&tar::Header> for Stat {
```
allisonkarlitskaya, line 39: This seems like a bit of a layering violation.  The knowledge about tar should probably be limited to the code in the `oci/` directory.  It probably makes sense to implement this there, and not as this trait.

src/image.rs
```diff
@@ -11,13 +11,43 @@ use anyhow::{bail, Context, Result};
...
     pub st_gid: u32,
     pub st_mtim_sec: i64,
-    pub xattrs: RefCell<BTreeMap<Box<OsStr>, Box<[u8]>>>,
+    pub xattrs: RefCell<StatXattrs>,
+}
+
+impl Default for Stat {
+    fn default() -> Self {
+        Self {
+            st_mode: u32::MAX,
```
allisonkarlitskaya, line 28: These are somewhat unfortunate choices for default values since they're all pretty much invalid.  The use of those values below was a giant hack to deal with not knowing the correct values to put on the root directory inode.  I wouldn't want to copy that hack into the default values for the struct — it would actually be nice to get rid of it entirely, if possible.

src/image.rs
```diff
@@ -28,6 +58,8 @@ pub enum LeafContent {
     CharacterDevice(u64),
     Fifo,
     Socket,
+    /// Contains the target of the link
+    /// The actual link path should be in TarEntry.path
```
allisonkarlitskaya, line 62: This comment seems a bit out of place (again: this code should have no knowledge about tar).

src/constants.rs
```diff
@@ -0,0 +1,8 @@
+/// All files that contain 64 or fewer bytes (size <= INLINE_CONTENT_MAX) should be stored inline
```
allisonkarlitskaya, line 1: Why did you create this file, just curious?  I'm not really against it, but I was wondering if this is some kind of a standard practice in rust?

src/oci/tar.rs
```diff
@@ -99,14 +100,17 @@ pub async fn split_async(
...
     Directory,
     Leaf(LeafContent),
+    /// Contains the target of the link
+    /// The actual link path should be in TarEntry.path
     Hardlink(OsString),
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct TarEntry {
```
allisonkarlitskaya, line 114: Was sort of hoping to see this gone ;)

src/oci/tar.rs
```diff
@@ -156,7 +160,211 @@ fn symlink_target_from_tar(pax: Option<Box<[u8]>>, gnu: Vec<u8>, short: &[u8]) -
...
+            tar_entry
+        };
+
+        return Ok(Some(tar_entry));
+    }
+
+    Ok(None)
+}
+
+pub fn get_entry_old<R: Read>(reader: &mut SplitStreamReader<R>) -> Result<Option<TarEntry>> {
```
allisonkarlitskaya, line 367: Does anything call that anymore?  Can we erase it?

allisonkarlitskaya (changes requested):
Thanks for working on this.  Some initial comments below, but let's try to schedule a meeting to discuss.

src/oci/tar.rs
Johan-Liebert1: yes, we can remove it. I forgot to do so

src/constants.rs
Johan-Liebert1: Nothing related to Rust coding practices, just thought having constants in a single file made sense.

src/image.rs
Johan-Liebert1: I was wondering if we could turn these into `Option<u32>`

src/oci/tar.rs
Johan-Liebert1: The `TarEntry` struct? A few tests were dependent on it, otherwise we can safely remove it as it only serves as an intermediate value

src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(missing_debug_implementations)]
+#![allow(clippy::needless_return)]
```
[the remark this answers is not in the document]
Johan-Liebert1: I feel like sometimes returning the result of an entire match statement gets a little confusing. We can of course standardize it across the repository. Should I remove this from this PR?

src/oci/tar.rs
```diff
@@ -10,17 +10,32 @@ use std::{
...
     splitstream::{SplitStreamData, SplitStreamReader, SplitStreamWriter},
     util::{read_exactish, read_exactish_async},
     INLINE_CONTENT_MAX,
 };
 
+// Constants related to tar archives
+const TAR_BLOCK_SIZE: usize = 512;
+const PAX_SCHILYXATTR: &str = "SCHILY.xattr.";
+
+fn stat_from_tar_header(header: &tar::Header) -> Result<Stat> {
```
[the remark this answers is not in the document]
Johan-Liebert1: Yes, standardizing errors would be nice

src/oci/tar.rs
```diff
@@ -217,39 +442,51 @@ pub fn get_entry<R: Read>(reader: &mut SplitStreamReader<R>) -> Result<Option<Ta
                     }
                     continue;
                 }
+
```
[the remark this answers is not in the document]
Johan-Liebert1: Ah, my bad. I removed other whitespace additions; seems like I forgot to remove this one.

PR #73, opened 2025-03-10
examples/bls: Add integration with Cockpit CI

allisonkarlitskaya:
This makes some changes to the bls example to make its image compatible with the Cockpit testing library:
 - installs an ssh server
 - installs cockpit's public key as `~root/.ssh/authorized_keys`
 - renames the image to have a more descriptive name and end with `-efi` for EFI booting
 - enables compatibility with cockpit's `vm-run` command
 - adds a trivial test case based on Cockpit's testing library
   - make sure / is read-only
   - make sure /sysroot is present and contains the expected files
   - make sure files written to /var persist across reboots
 - adds integration for cockpit CI's runners (not connected yet)


allisonkarlitskaya:
So a fairly serious issue here with my plan to run this from Cockpit CI: despite my best (and many) efforts, we can't build containers inside of containers :(

The most obvious idea for me was to use a VM to build the container, which led me to think "hey, isn't that what podman-machine is?" but frustratingly, it doesn't seem to like running in containers either (due to no systemd)...

allisonkarlitskaya:
So the actual testcase is already quite useful and can be run locally.  Let's merge that part, at least.  I dropped the cockpit-ci integration...

PR #75, opened 2025-03-14
.github/workflows: enable KVM

allisonkarlitskaya:


allisonkarlitskaya:
That was a test

.github/workflows/examples.yml
```diff
@@ -26,6 +26,13 @@ jobs:
     - name: Enable fs-verity on /
       run: sudo tune2fs -O verity $(findmnt -vno SOURCE /)
 
+    - name: Setup /dev/kvm
+      run: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm.rules
```
cgwalters, line 31: Offhand this all looks correct, but in practice isn't just a really complex way to do `chmod a+rx /dev/kvm`? I mean, the chance that we somehow race with udev probing from kvm here in GHA runners seems...extraordinarily small if nonexistent entirely.

But yeah, might as well do it the udev way.

allisonkarlitskaya: I copied this from the PR you linked me to (and transliterated it from GitHub Actions JS into shell).  I think the issue is that the device node doesn't exist at all in the first place, and this will cause it to be created, no?  We could maybe mknod it as well, but that also seems kinda dicey...

PR #89, opened 2025-04-03
examples: pull directly from containers storage

allisonkarlitskaya:
Instead of having podman save our container images to a oci-archive and pulling from that, save ourselves a bunch of work (including a gzip compress/decompress cycle) by pulling directly from `containers-storage:`.

In order to accomplish this, the patch series also adds support for uncompressed and zstd-compressed layers and includes a workaround for https://github.com/containers/skopeo/issues/2563

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -12,6 +12,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-24.04
+    timeout-minutes: 15
 
     strategy:
       matrix:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,6 +12,7 @@ env:
 jobs:
   fedora:
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     container:
       image: quay.io/fedora/fedora:41
       options: "--privileged --pid=host -v /var/tmp:/var/tmp -v /:/run/host"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ rhel9 = ['pre-6.15']
 
 [dependencies]
 anyhow = { version = "1.0.97", default-features = false }
-async-compression = { version = "0.4.22", default-features = false, features = ["tokio", "gzip"] }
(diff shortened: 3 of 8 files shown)


allisonkarlitskaya:
Adding this to the bootc milestone because the zstd+uncompressed layer part of this work is pretty important...

allisonkarlitskaya:
todo: let's set a reasonable timeout on our workflows...

cgwalters:
> This isn't working right now.  The pull of the first layer stalls out like this:

It's probably because of the exact same thing that motivated https://github.com/bootc-dev/bootc/pull/1248

(Again, a lot of dedup/alignment with bootc code to do here)

cgwalters:
Though there's zstd:chunked issue lingering too here, see https://github.com/ostreedev/ostree-rs-ext/pull/622

allisonkarlitskaya:
> > This isn't working right now.  The pull of the first layer stalls out like this:
> 
> It's probably because of the exact same thing that motivated [bootc-dev/bootc#1248](https://github.com/bootc-dev/bootc/pull/1248)
> 
> (Again, a lot of dedup/alignment with bootc code to do here)
> 
> EDIT: More info in [bootc-dev/bootc#1228](https://github.com/bootc-dev/bootc/pull/1228)

The workaround in bootc looks a lot like a workaround.  Couldn't we try to do a common fix for this in containers-image-proxy-rs?

allisonkarlitskaya:
I think I have a functioning workaround for this and it looks like it's the opposite of the "broken pipe" issue: our problem is that we're reading too much.  If we manually abort after the "end of stream" headers in the tar stream then this will actually work.

I'm not really satisfied with this workaround, though, at least not without a proper upstream report.  I'm gonna try to get a reproducer on containers-image-proxy-rs.

Johan-Liebert1 (approved):
LGTM

allisonkarlitskaya:
Oh.  This is obviously just https://github.com/containers/containers-image-proxy-rs/issues/71 back to haunt us again.  I'm glad I went to the effort of making a reproducer :rofl:

cgwalters (approved):

PR #94, opened 2025-04-04
repository: open_path: include path in error context

jeckersb:
It was not immediately obvious to me where the default repository was
expected to be located.

### DIFF
--- a/src/repository.rs
+++ b/src/repository.rs
@@ -47,9 +47,11 @@ impl Repository {
     }
 
     pub fn open_path(dirfd: impl AsFd, path: impl AsRef<Path>) -> Result<Repository> {
+        let path = path.as_ref();
+
         // O_PATH isn't enough because flock()
-        let repository = openat(dirfd, path.as_ref(), OFlags::RDONLY, Mode::empty())
-            .context("Cannot open composefs repository")?;
+        let repository = openat(dirfd, path, OFlags::RDONLY, Mode::empty())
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

src/repository.rs
```diff
@@ -47,9 +47,11 @@ impl Repository {
     }
 
     pub fn open_path(dirfd: impl AsFd, path: impl AsRef<Path>) -> Result<Repository> {
+        let path = path.as_ref();
+
         // O_PATH isn't enough because flock()
-        let repository = openat(dirfd, path.as_ref(), OFlags::RDONLY, Mode::empty())
-            .context("Cannot open composefs repository")?;
+        let repository = openat(dirfd, path, OFlags::RDONLY, Mode::empty())
+            .with_context(|| format!("Cannot open composefs repository at {}", path.display()))?;
```
allisonkarlitskaya, line 54: I actually dropped that recently because it doesn't make completely sense in context of the `dirfd` parameter and I was trying to slowly remove use of `anyhow`.  Admittedly, `dirfd` is basically always `CWD` currently...

My idea here was that the caller should add the context because it was more likely to know the full path.  Of course I never added that to any of the callers :)

Is there a usual rule for this sort of thing in Rust?

cgwalters: In this case I'd push the error prefixing to the CLI, right?

Issue #62, opened 2025-01-27
rfe: multithread splitstream creation


allisonkarlitskaya:
`cfsctl oci pull` is kinda slow right now.

When converting the container image `.tar.gz` into splitstreams it does the gzip decompression, sha256 verification (of the underlying `.tar`), fs-verity calculations (of the split objects), `fdatasync()` (of the split objects), zstd compression of the splitstream, and finally fs-verity and `fdatasync()` of the splitstream, all in a single thread.

`mkcomposefs` is smarter here: it spawns a bunch of worker threads to do the fs-verity and fdatasync() on the split objects.

This would be pretty easy for us to do, and Rust would help us make sure we're doing it right, except the way that splitstreams are currently constructed is pretty linear: you need to get the object ID of each object to write into the stream before you can proceed to the next one.  But: we're just writing to a `Vec<u8>` anyway, so we can easily go back in fill those parts in, or come up with some kind of a "chunks" model where we merge all the chunks into the linear stream at the end, or any other design.  We could then make some of the chunks be promises or something — I'm really not sure how this works in Rust...

allisonkarlitskaya:
This also needs to work for `cfsctl create-image` which scans the filesystem.  We might also try to opportunistically use reflinking in that case.

allisonkarlitskaya:
For the splitstream case: I think some sort of model like this:
 - we have two queues:
   - some kind of a "work queue" which is handled by workers 
   - some kind of an "output" queue which the splitstream segments get written to
 - one thread does the download/decompress/tar decoding
 - for each inline chunk, write it directly to the output queue
 - for each external chunk:
   - create a new ephemeral queue (as a sort of promise mechanism)
   - add an item to the worker queue containing:
     - the data
     - a promise queue Sender side
   - write the promise queue Receiver side to the "output" queue
 - the workers receive (data, Sender) pairs and write the data to the repo, sending the result (the fs-verity digest) to the Sender.
 - finally, have another thread responsible for composing the output.  It reads inline chunks and promises and writes the splitstream.

allisonkarlitskaya:
For the `cfsctl create-image` case it's going to be harder.  We could imagine a filesystem tree where external files contain promises for fs-verity digests instead of the fs-verity digests but this is a serious complication of the existing structure.  We could model it, instead, as a queue of (filename, fd) and move the read/checksum/write of the fd to a worker, producing another queue of (filename, checksum).  Ensuring ordering would be annoying in this case, though...

[...]

PR #95, opened 2025-04-07
Multithreaded SplitStream creation

Johan-Liebert1:
Related to - https://github.com/containers/composefs-rs/issues/62

With this we spawn multiple threads for external object writing and Zstd encoding + rolling SHA256sum update. Currently this is only supported for `Oci::Pull` operation and currently it doesn't handle multiple layers in parallel.

Communication between threads are handled by crossbeam mpmc channels.  For now we only have one channel for external object writing, but multiple for Encoders. The reasoning for this is Encoder threads are usually CPU bound while the object writer threads are more IO bound, so the back-pressure on the writer channels is negligible.

Performance gain ~3x


allisonkarlitskaya:
Suspicious test failures, timing out on pull...

```
+ ./cfsctl --repo tmp/sysroot/composefs oci pull containers-storage:c22ddccc8dcd7efeee38efdae72773ceff54eb70e47cdb32357c30aca5bfd7c2
Error: The operation was canceled.
```

src/oci/mod.rs
```diff
@@ -154,32 +161,162 @@ impl<'repo> ImageOp<'repo> {
         } else {
             // We need to add the config to the repo.  We need to parse the config and make sure we
             // have all of the layers first.
-            //
             self.progress
                 .println(format!("Fetching config {}", hex::encode(config_sha256)))?;
             let raw_config = self.proxy.fetch_config_raw(&self.img).await?;
             let config = ImageConfiguration::from_reader(raw_config.as_slice())?;
 
+            let (done_chan_sender, done_chan_recver, object_sender) = self.spawn_threads(&config);
```
cgwalters, line 169: Having played with things related to this in the past, my recommendation is to try instead to use tokio's [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) instead.

Here's some example code that does this, and uses a JoinSet as a way to both keep track of the tasks, but also act as a cap on maximum concurrency.

See
https://github.com/bootc-dev/bootc/blob/2de8e0d23fb89bed76722ff1466614afacec64b3/lib/src/fsck.rs#L195

Among the major benefits of this model is that these are just async tasks, which means cancellation works naturally; dropping the future holding the joinset.

cgwalters: I mean this is a super complex topic; I would say we should turn to rayon if we happened to be doing a lot of computation on in-memory array-like structures and could just use their super high level `.par_iter()` and `.reduce()` style APIs. But this is far from that.


(Nothing wrong with crossbeam either, but I am not sure we need it versus just the dedicated [tokio mpsc](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#communicating-between-sync-and-async-code) channels either? But this concern goes away too if we just use `spawn_blocking` as our communication mechanism between synchronous threads and async)

allisonkarlitskaya: > Having played with things related to this in the past, my recommendation is to try instead to use tokio's [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) instead.

This was also my first instinct.  I feel like this would make things a lot easier to generalize to the "multiple layers at the same time" level, as well...

Johan-Liebert1: I would agree this became a bit more complicated than anticipated. Converting these into a bunch of async operations was my first approach as well, but there were some issues with some values not being sendable. 

Thanks for the suggestion, I'll try out the async approach again

.gitignore
```diff
@@ -1,2 +1,3 @@
 /Cargo.lock
 /target
```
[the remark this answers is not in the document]
Johan-Liebert1: Thanks, moved it to a separate commit

src/oci/mod.rs
cgwalters: > but there were some issues with some values not being sendable.

Right, this is https://tmandry.gitlab.io/blog/posts/2023-03-01-scoped-tasks/ (see also https://without.boats/blog/the-scoped-task-trilemma/ )

The simplest thing to do here is drop the borrowing (making things `Send`) most usually by operating on owned data. For example in the case of fetching a container image layer, while the proxy exposes `get_blob() -> impl AsyncBufRead + Send + Unpin` honestly in the end it's just a file descriptor, and we should change the proxy to have a `get_blob_fd()` or so that gives you that underlying pipe file descriptor. Once we have that, it's a `File/OwnedFd` that we can just directly move ownership of into a worker thread.

cgwalters: > but there were some issues with some values not being sendable.

Right, this is https://tmandry.gitlab.io/blog/posts/2023-03-01-scoped-tasks/ (see also https://without.boats/blog/the-scoped-task-trilemma/ )

The simplest thing to do here is drop the borrowing (making things `Send`) most usually by operating on owned data. For example in the case of fetching a container image layer, while the proxy exposes `get_blob() -> impl AsyncBufRead + Send + Unpin` honestly in the end it's just a file descriptor, and we should change the proxy to have a `get_blob_fd()` or so that gives you that underlying pipe file descriptor. Once we have that, it's a `File/OwnedFd` that we can just directly move ownership of into a worker thread.

Johan-Liebert1: So, trying out an async approach using tokio's JoinSet results in no performance gains. Weird. I can see the layers being worked on in parallel, as I can see multiple progress bars at once. 

https://github.com/Johan-Liebert1/composefs-rs/blob/splitstream-tokio-mt/src/oci/mod.rs#L168

The main bottleneck in this case would be writing the external objects, which has a lot of synchronous syscalls. Wrapping them in their own `spawn_blocking` call also doesn't seem to have much of an affect on the performance.

https://github.com/Johan-Liebert1/composefs-rs/blob/splitstream-tokio-mt/src/repository.rs#L85

Johan-Liebert1: Running a quick and dirty test, it does seem like the layers that are smaller are being processed first while other layers are being processed in the background. So there does seem to be some sort of parallelism going on here. What I find really unintuitive here is that it's not better than the vanilla approach

allisonkarlitskaya:
#111 lays some groundwork that would probably make this easier to do, particularly if you decide to go with the tokio-based approach (which makes more and more sense to me)...

PR #96, opened 2025-04-07
splitstream: Drop an unsafe

cgwalters:
This one didn't have a `SAFETY` comment; looking through git log here I do see:

> It no longer does internal buffer allocations when reading tar headers, saving
> another ~10% off the total time to print a merged dumpfile.

However, it looks like that was going from allocation-per-read to "read into a global buffer", which we're still doing here. I didn't measure performance here, but I think
we have other performance worries and it's not worth carrying an `unsafe` for this.

In very hot I/O paths one *can* measure the impact of zero-initializing, but I'm doubtful of that here.

Just looking at the bigger picture here I think
instead of a generic `R: Read` we should take
a `BufReader` and then use `fill_buf` on it,
which is intended for this use case.

Anyways for now just drop the `unsafe`.

Also add tests.

### DIFF
--- a/src/splitstream.rs
+++ b/src/splitstream.rs
@@ -198,13 +198,13 @@ fn read_u64_le<R: Read>(reader: &mut R) -> Result<Option<usize>> {
     }
 }
 
+/// Using the provided [`vec`] as a buffer, read exactly [`size`]
+/// bytes of content from [`reader`] into it. Any existing content
+/// in [`vec`] will be discarded; however its capacity will be reused,
+/// making this function suitable for use in loops.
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
I agree that I'd prefer to drop the unsafe here, even if it costs us a bit.

Rustix has a really cool new buffer approach that I'd love to use here, but unfortunately it's not for the Read trait.  It seems like Rust is still having trouble making up its mind about what to do here: https://github.com/rust-lang/rust/issues/117693

Thanks for the tests!

PR #97, opened 2025-04-07
various src/image.rs cleanups

allisonkarlitskaya:
I think this is a net win... it seems a whole lot less ad-hoc to me.  The last commit could maybe do with some more splitting up...

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -1,4 +1,10 @@
-use std::{cell::RefCell, collections::BTreeMap, ffi::OsStr, path::Path, rc::Rc};
+use std::{
+    cell::RefCell,
+    collections::BTreeMap,
+    ffi::OsStr,
+    path::{Component, Path},
+    rc::Rc,
+};
 
 use thiserror::Error;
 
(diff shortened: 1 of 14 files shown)


src/image.rs
```diff
@@ -47,6 +47,18 @@ pub enum Inode {
     Leaf(Rc<Leaf>),
 }
 
+#[derive(Error, Debug)]
+pub enum ImageError {
+    #[error("Directory entry {0:?} does not exist")]
```
cgwalters, line 60: Minor style thing but I think where possible it's best to have any user-injected content at the end instead of the middle; it's safer to "visually parse", like:

`#[error("Directory entry does not exist: {0:?}")]`

(Also different tangent on filename display, in bootc I added https://github.com/bootc-dev/bootc/blob/2de8e0d23fb89bed76722ff1466614afacec64b3/utils/src/path.rs#L30 which I think is nicer than the default debug output)

cgwalters (approved):

src/image.rs
allisonkarlitskaya: To be honest, I'm not a huge fan of how the error reporting on filenames is working generally: most of these functions are going to end up throwing an error for a partial pathname instead of the whole thing, which is kinda weak...  I'll take a todo to clean this up later.

allisonkarlitskaya:
hihi new clippy...

some pretty nice suggestions in there to be honest.  I'll add a commit.

allisonkarlitskaya:
Nice.  Snapshot testing caught a bug introduced by my botched application of the clippy fixes...

cgwalters (approved):

Issue #91, opened 2025-04-04
Error: Unable to pull container image docker://alpine


allisonkarlitskaya:
Pretty straightforward.  `cfsctl oci pull docker://alpine` fails.  This is an old issue, but we need to fix it.

jeckersb:
For completeness-sake, the error I get is:

```
$ cargo run -- --repo repo oci pull docker://alpine
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/cfsctl --repo repo oci pull 'docker://alpine'`
Fetching config aded1e1a5b3705116fa0a92ba074a5e0b0031647d9c315983ccba2ee5428ec8b
Already have layer 08000c18d16dadf9553d747a58cf44023423a9ab010aab96cf263d2216b8b350
Error: Unable to pull container image docker://alpine

Caused by:
    0: Failed to pull config Descriptor { media_type: ImageConfig, digest: Digest { algorithm: Sha256, value: "sha256:aded1e1a5b3705116fa0a92ba074a5e0b0031647d9c315983ccba2ee5428ec8b", split: 6 }, size: 581, urls: None, annotations: None, platform: None, artifact_type: None, data: None }
    1: Content doesn't have expected SHA256 hash value!
```

cgwalters:
It's because alpine is docker format, not OCI:
```
$ skopeo inspect --raw -n docker://alpine | jq -r .mediaType
application/vnd.oci.image.index.v1+json
$
```

And currently skopeo experimental-image-proxy intentionally only serves back OCI types to avoid callers needing to understand both formats, but it does invalidate the checksums.

We probably want a method that errors out instead of converting.

jeckersb:
@cgwalters helped me run this down today, the problem is basically here in skopeo:

https://github.com/containers/skopeo/blob/119816b17ce93274910c1e7d5914869582e5545e/cmd/skopeo/proxy.go#L506-L514

We end up converting the config to an `OCIConfig` and then re-marshaling it, and the end result is the config comes out with the same content, except the map keys are sorted alphabetically, so the checksum no longer matches the original as listed in the manifest.

I noticed that `skopeo inspect --config --raw` does the right thing though, which is here:

https://github.com/containers/skopeo/blob/119816b17ce93274910c1e7d5914869582e5545e/cmd/skopeo/inspect.go#L131-L139

[...]

PR #98, opened 2025-04-07
oci: ensure_config: get raw config via `get_descriptor`

jeckersb:
`fetch_config_raw` may reorder map keys in the config json, which
means the sha256sum of the config does not match what is expected from
the image manifest.

Resolves: #91


cgwalters (approved):
Probably we should change `fetch_config_raw` in the proxy to do this by default.

PR #100, opened 2025-04-08
uki: new file for querying UKI boot labels

allisonkarlitskaya:
This contains a single function to query an appropriate boot label for a UKI image, which might be interesting for supporting bootloaders which don't implement the bootloader specification and is also likely to be used for showing installed bootloader entries in the future.

Nothing in the crate uses this right now, but it's a public API.

### DIFF
--- a/src/uki.rs
+++ b/src/uki.rs
@@ -1,0 +1,434 @@
+use std::collections::HashMap;
+
+use thiserror::Error;
+use zerocopy::{
+    little_endian::{U16, U32},
+    FromBytes, Immutable, KnownLayout,
+};
+
+// https://learn.microsoft.com/en-us/windows/win32/debug/pe-format
+#[derive(Debug, FromBytes, Immutable, KnownLayout)]
+#[cfg_attr(test, derive(zerocopy::IntoBytes, Default))]
+#[repr(C)]
+struct DosStub {
+    _unused1: [u8; 0x20],
+    _unused2: [u8; 0x1c],
+    pe_offset: U32,
+}
+
+#[derive(Debug, FromBytes, Immutable, KnownLayout)]
+#[cfg_attr(test, derive(zerocopy::IntoBytes, Default))]
+#[repr(C)]
+struct CoffFileHeader {
+    machine: U16,
+    number_of_sections: U16,
+    time_date_stamp: U32,
+    pointer_to_symbol_table: U32,
+    number_of_symbols: U32,
+    size_of_optional_header: U16,
+    characteristics: U16,
+}
+
+#[derive(Debug, FromBytes, Immutable, KnownLayout)]
+#[cfg_attr(test, derive(zerocopy::IntoBytes, Default))]
+#[repr(C)]
+struct PeHeader {
+    pe_magic: [u8; 4], // P E \0 \0
+    coff_file_header: CoffFileHeader,
+}
+const PE_MAGIC: [u8; 4] = *b"PE\0\0";
+
+#[derive(Debug, FromBytes, Immutable, KnownLayout)]
+#[cfg_attr(test, derive(zerocopy::IntoBytes, Default))]
+#[repr(C)]
+struct SectionHeader {
+    name: [u8; 8],
+    virtual_size: U32,
+    virtual_address: U32,
+    size_of_raw_data: U32,
+    pointer_to_raw_data: U32,
+    pointer_to_relocations: U32,
+    pointer_to_line_numbers: U32,
+    number_of_relocations: U16,
+    number_of_line_numbers: U16,
+    characteristics: U32,
+}
+const OSREL_SECTION: [u8; 8] = *b".osrel\0\0";
+
+#[derive(Debug, Error, PartialEq)]
+pub enum UkiError {
+    #[error("UKI is not valid EFI executable")]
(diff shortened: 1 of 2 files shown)


src/uki.rs
```diff
@@ -0,0 +1,183 @@
...
+    // https://www.freedesktop.org/software/systemd/man/latest/os-release.html
+    let os_release = std::str::from_utf8(get_pe_section(image, b".osrel\0\0")?)?;
+    // We're intentionally dumb about the parsing and this may snap up comments and stuff, but we
+    // won't ever do lookups for keys starting with '#'...
+    Ok(HashMap::from_iter(
+        os_release.lines().filter_map(|line| line.split_once('=')),
+    ))
+}
+
+fn dequote(value: &str) -> Option<String> {
```
cgwalters, line 111: This function seems very unit-testable

Also, there's the https://docs.rs/shlex/latest/shlex/ crate

(And for that matter https://github.com/pop-os/os-release but that thing is...really old and not maintained)

src/uki.rs
```diff
@@ -0,0 +1,183 @@
...
+/// ".osrel" section in it, return a string to use as the boootloader entry.  If we were unable to
+/// find any meaningful content in the os-release information this will be "Unknown 0".
+///
+/// If we couldn't parse the PE file or couldn't find an ".osrel" section then an error will be
+/// returned.
+pub fn get_boot_label(image: &[u8]) -> Result<String, UkiError> {
+    let os_release = get_os_release(image)?;
+    let name = get_value(&os_release, &["PRETTY_NAME", "ID"], "Unknown");
+    let version = get_value(&os_release, &["VERSION_ID", "VERSION"], "0");
+    Ok(format!("{name} {version}"))
```
cgwalters, line 182: Obviously this is a degenerate case but it seems clearer to omit the version if it's empty than to make it 0

src/uki.rs
```diff
@@ -0,0 +1,183 @@
...
+            return Ok(&image[start..end]);
+        }
+    }
+
+    Err(UkiError::MissingOsrelSection)
+}
+
+fn get_os_release(image: &[u8]) -> Result<HashMap<&str, &str>, UkiError> {
+    // https://www.freedesktop.org/software/systemd/man/latest/os-release.html
+    let os_release = std::str::from_utf8(get_pe_section(image, b".osrel\0\0")?)?;
```
cgwalters, line 103: I don't know much about this but it seems like the double-NUL (and leading `.` maybe)  should be implementation details of `get_pe_section`?

src/uki.rs
```diff
@@ -0,0 +1,183 @@
...
+    // We're intentionally dumb about the parsing and this may snap up comments and stuff, but we
+    // won't ever do lookups for keys starting with '#'...
```
cgwalters, line 105: Fair, but it's actually less typing to do `.filter(|line| !line.starts_with('#'))` than to explain why we're not handling comments right?

cgwalters (approved):
Looks sane to me

src/uki.rs
allisonkarlitskaya: You know me: I hate taking dependencies :)

I think the subset of shell dequoting that we need to understand here is simple enough to just do it ourselves...  I agree that some tests would be good though.

src/uki.rs
allisonkarlitskaya: Agreed, although it makes the thing with the space slightly more annoying.  Maybe I'll just leave the space in :)

src/uki.rs
allisonkarlitskaya: the `.` I'd leave in for sure.  It's written that way in all the specs.  The double-nul we could talk about... it kinda just seems like a way to complicate the implementation, though... and this is a private function after all.  I think I'll leave this one as-is.

src/uki.rs
allisonkarlitskaya: Then we'd also have to trim whitespace...  but okay, sure :)

allisonkarlitskaya (commented):
One big minus from me: I use `[x..y]` here in the PE parsing code.  It blows my mind that this can panic by default... I really prefer the Python behaviour of returning an empty slice if it's out of bounds...

src/uki.rs
allisonkarlitskaya: This adds 5(!!) lines because it changes how the wrapping works with `rustfmt`...

allisonkarlitskaya:
(working on addressing the comments, adding tests, etc.)

src/uki.rs
```diff
@@ -0,0 +1,434 @@
...
+mod test {
+    use core::mem::size_of;
+
+    use similar_asserts::assert_eq;
+    use zerocopy::IntoBytes;
+
+    use super::*;
+
+    #[test]
+    fn test_dequote() {
```
cgwalters, line 200: Optional followup: I think this would make sense as a separate `mod shell_parser` or something?

cgwalters (approved):

src/uki.rs
allisonkarlitskaya: Maybe, but it really does parse a subset of shell as is present in os-release files.  If we need some general-purpose shell parsing stuff in the future I'd consider `shlex`...

PR #102, opened 2025-04-11
fsverity: SHA-512 support, variable block sizes, lots of cleanups, docs, tests

allisonkarlitskaya:


allisonkarlitskaya:
Outstanding issues:
 - EnableVerityError wraps rustix Errno instead of std::io:Error which exposes an implementation detail.  We might want to avoid that (although it would come at the cost of losing PartialEq).
 - EnableVerityError could maybe use a separate "already enabled" error case

src/fsverity/mod.rs
```diff
@@ -1,27 +1,35 @@
...
-pub mod ioctl;
+pub use ioctl::{EnableVerityError, MeasureVerityError};
 
-pub trait FsVerityHashValue: Eq + AsRef<[u8]> {
+pub trait FsVerityHashValue
+where
+    Self: Eq + AsRef<[u8]> + Clone,
+    Self: From<Output<Self::Digest>>,
+{
+    type Digest: Digest + FixedOutputReset + std::fmt::Debug;
```
allisonkarlitskaya, line 16: This belongs in the next commit.

src/fsverity/mod.rs
```diff
@@ -142,21 +142,204 @@ pub fn ensure_verity_equal(
...
+        test::tempfile,
+        util::{parse_sha256, proc_self_fd},
+    };
+    use rustix::{
+        fd::OwnedFd,
+        fs::{open, Mode, OFlags},
+    };
+    use tempfile::tempfile_in;
+
+    use super::{compute_verity, *};
```
allisonkarlitskaya, line 157: This is an accidental inclusion.

src/fsverity/mod.rs
```diff
@@ -142,21 +142,204 @@ pub fn ensure_verity_equal(
 
 #[cfg(test)]
 mod tests {
-    use crate::fsverity::Sha256HashValue;
```
allisonkarlitskaya: The removal of use Sha256HashValue should have been handled in the previous test cleanups commit.

PR #103, opened 2025-04-12
Cargo.toml: list actual minimum dependencies

allisonkarlitskaya:
In 72d8b71465c6 ("Cargo.toml: update all the things") we randomly raised all of our dependencies to the latest versions, mostly because I don't know what I'm doing when it comes to these things.  That was a mistake: trying to use the latest version of composefs-rs from bootc causes all kinds of dependency conflicts now.

So let's try and do the opposite and try to list the minimum version that:
 - is semver compatible with the latest version (so we get updates)
 - doesn't break

I tried playing with automated tooling for this but it wasn't easy to get it working without a nightly toolchain (which I don't have setup), and I was also interested in exploring the actual reasons for the requirements, so I went with a simple manual process: I dropped all of the version numbers in [dependencies] (but not [dev-dependencies]) to their lowest semver-compatible version and added "=" in front.

Then I adjusted upwards for yanked releases:
 - sha2 0.10.0 -> 0.10.1
 - tempfile 3.0.0 -> 3.0.2
 - tokio 1.0.0 -> 1.0.1
 - xxhash-rust 0.8.0 -> 0.8.2

Then (because of '=' dependencies) I needed to make adjustments for conflicts with dependencies specified by other packages:
 - log 0.4.0 -> log 0.4.8 to satisfy an env_logger dependency
 - regex-automata 0.4.0 -> 0.4.4 for regex (via oci-spec)
 - serde 1.0.0 -> 1.0.145 for toml
 - tokio 1.0.1 -> 1.24.2 for async-compression

Then I actually tried to build and run unit tests to determine which versions were actually required in this crate, and bumped some more things:
 - anyhow 1.0.0 -> 1.0.87 to support the automatic conversions we require via use of `?`
 - clap 4.0.0 -> 4.0.1 was the first version with #[clap(version)]
 - tar 0.4.0 -> 0.4.38 for PaxExtensions::new()
 - tempfile 3.0.2 -> 3.8.0 for tempfile_in()

After that, I ran the examples and everything there seemed to work properly without additional changes, so I dropped the '=' signs and finished by checking that it was now possible to once again consume composefs-rs from bootc without a bunch of unnecessary version changes on that side.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,27 +17,27 @@ rhel9 = ['pre-6.15']
 'pre-6.15' = []
 
 [dependencies]
-anyhow = { version = "1.0.97", default-features = false }
-async-compression = { version = "0.4.22", default-features = false, features = ["tokio", "zstd", "gzip"] }
-clap = { version = "4.5.32", default-features = false, features = ["std", "help", "usage", "derive"] }
+anyhow = { version = "1.0.87", default-features = false }
+async-compression = { version = "0.4.0", default-features = false, features = ["tokio", "zstd", "gzip"] }
+clap = { version = "4.0.1", default-features = false, features = ["std", "help", "usage", "derive"] }
 containers-image-proxy = "0.7.0"
-env_logger = "0.11.7"
-hex = "0.4.3"
-indicatif = { version = "0.17.11", features = ["tokio"] }
-log = "0.4.27"
-oci-spec = "0.7.1"
-regex-automata = { version = "0.4.9", default-features = false }
-rustix = { version = "1.0.3", features = ["fs", "mount", "process"] }
-serde = "1.0.219"
-sha2 = "0.10.8"
-tar = { version = "0.4.44", default-features = false }
-tempfile = "3.19.1"
-thiserror = "2.0.12"
-tokio = "1.44.1"
-toml = "0.8.20"
-xxhash-rust = { version = "0.8.15", features = ["xxh32"] }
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
There's a big difference between well-established crates that have been published for years (e.g. `anyhow, `hex`, `serde`) vs ones that definitely change often (e.g. `zerocopy`, `clap`) etc.

In practice I think *most* crates only ad-hoc bump on demand and don't try to aggressively test minimal versions for the former case. Take as one reference in rustix: https://github.com/bytecodealliance/rustix/blame/main/Cargo.toml

Several of those crate requirements in that first group were last bumped years ago, like `tempfile`.

allisonkarlitskaya:
> There's a big difference between well-established crates that have been published for years (e.g. `anyhow`, `hex`, `serde`) vs ones that definitely change often (e.g. `zerocopy`, `clap`) etc.
> 
> In practice I think _most_ crates only ad-hoc bump on demand and don't try to aggressively test minimal versions for the former case. Take as one reference in rustix: https://github.com/bytecodealliance/rustix/blame/main/Cargo.toml
> 
> Several of those crate requirements in that first group were last bumped years ago, like `tempfile`.

I'd definitely be interested in improving our situation here, bumping depends when it makes sense and also actually verifying that we work with the minimal depends that we advertise.  I mostly did this because I wanted to get the composefs-rs work landed in bootc...

PR #105, opened 2025-04-15
add complete (library) support for sha-512

allisonkarlitskaya:
This rests upon a massive pile of generics, but it seems to be working.

With these changes the library is now fully agnostic to which hash algorithm we use.  I've tested this and it successfully produces erofs images with the metacopy attributes correctly set for verifying the fs-verity validity of files using sha512.  The examples also correctly boot with a sha512 in the `composefs=` cmdline.

The changes to the bin/ commands to support this will come soon: we need to thoughtfully consider what we want to do here at a higher level....

 - [x] remove the last two commits (which are here just for demo purposes)
 - [x] decide what we want to do about sha256/sha512 or both: for now we stick with SHA-256 by default.  *edit*: We can enable support for SHA-512 in the commandline tools in future commits.
 - [x] consider that we're essentially producing an incompatible repository format (with differently named objects, images, different image content, different splitstream content, everything) and figure out how we want to handle that from a compat standpoint.  *edit*: punt.

### DIFF
--- a/src/bin/cfsctl.rs
+++ b/src/bin/cfsctl.rs
@@ -5,7 +5,12 @@ use clap::{Parser, Subcommand};
 
 use rustix::fs::CWD;
 
-use composefs::{oci, repository::Repository, util::parse_sha256};
+use composefs::{
+    fsverity::{FsVerityHashValue, Sha256HashValue},
+    oci,
+    repository::Repository,
+    util::parse_sha256,
+};
 
 /// cfsctl
 #[derive(Debug, Parser)]
@@ -130,7 +135,7 @@ fn main() -> Result<()> {
         }
         Command::ImportImage { reference } => {
             let image_id = repo.import_image(&reference, &mut std::io::stdin())?;
-            println!("{}", hex::encode(image_id));
+            println!("{}", image_id.to_hex());
         }
         Command::Oci { cmd: oci_cmd } => match oci_cmd {
             OciCommand::ImportLayer { name, sha256 } => {
@@ -140,7 +145,7 @@ fn main() -> Result<()> {
                     name.as_deref(),
                     &mut std::io::stdin(),
                 )?;
-                println!("{}", hex::encode(object_id));
+                println!("{}", object_id.to_hex());
             }
             OciCommand::LsLayer { name } => {
                 oci::ls_layer(&repo, &name)?;
@@ -150,7 +155,7 @@ fn main() -> Result<()> {
             }
             OciCommand::CreateImage { config, name } => {
                 let image_id = oci::image::create_image(&repo, &config, name.as_deref(), None)?;
-                println!("{}", hex::encode(image_id));
+                println!("{}", image_id.to_hex());
             }
             OciCommand::Pull { ref image, name } => {
                 let runtime = tokio::runtime::Builder::new_current_thread()
@@ -161,10 +166,13 @@ fn main() -> Result<()> {
                 runtime.block_on(async move { oci::pull(&repo, image, name.as_deref()).await })?;
             }
             OciCommand::Seal { verity, ref name } => {
-                let (sha256, verity) =
-                    oci::seal(&repo, name, verity.map(parse_sha256).transpose()?.as_ref())?;
+                let (sha256, verity) = oci::seal(
+                    &repo,
+                    name,
+                    verity.map(Sha256HashValue::from_hex).transpose()?.as_ref(),
+                )?;
                 println!("sha256 {}", hex::encode(sha256));
-                println!("verity {}", hex::encode(verity));
+                println!("verity {}", verity.to_hex());
             }
             OciCommand::Mount {
                 ref name,
@@ -182,7 +190,7 @@ fn main() -> Result<()> {
         },
         Command::CreateImage { ref path } => {
             let image_id = composefs::fs::create_image(path, Some(&repo))?;
-            println!("{}", hex::encode(image_id));
+            println!("{}", image_id.to_hex());
         }
         Command::CreateDumpfile { ref path } => {
             composefs::fs::create_dumpfile(path)?;
@@ -193,7 +201,7 @@ fn main() -> Result<()> {
         Command::ImageObjects { name } => {
             let objects = repo.objects_for_image(&name)?;
             for object in objects {
-                println!("{}", hex::encode(object));
+                println!("{}", object.to_hex());
             }
         }
         Command::GC => {
--- a/src/bin/cfsctl.rs
+++ b/src/bin/cfsctl.rs
@@ -111,7 +111,7 @@ fn main() -> Result<()> {
 
     let args = App::parse();
 
-    let repo = (if let Some(path) = args.repo {
+    let repo: Repository<Sha256HashValue> = (if let Some(path) = args.repo {
         Repository::open_path(CWD, path)
     } else if args.system {
         Repository::open_system()
@@ -193,7 +193,7 @@ fn main() -> Result<()> {
             println!("{}", image_id.to_hex());
         }
         Command::CreateDumpfile { ref path } => {
-            composefs::fs::create_dumpfile(path)?;
+            composefs::fs::create_dumpfile::<Sha256HashValue>(path)?;
         }
         Command::Mount { name, mountpoint } => {
             repo.mount(&name, &mountpoint)?;
--- a/src/fsverity/hashvalue.rs
+++ b/src/fsverity/hashvalue.rs
@@ -1,0 +1,27 @@
+use sha2::{digest::FixedOutputReset, digest::Output, Digest, Sha256, Sha512};
+
+pub trait FsVerityHashValue
+where
+    Self: Eq + AsRef<[u8]> + Clone,
+    Self: From<Output<Self::Digest>>,
+{
+    type Digest: Digest + FixedOutputReset + std::fmt::Debug;
+    const ALGORITHM: u8;
+    const EMPTY: Self;
+}
+
+pub type Sha256HashValue = [u8; 32];
+
+impl FsVerityHashValue for Sha256HashValue {
+    type Digest = Sha256;
+    const ALGORITHM: u8 = 1;
+    const EMPTY: Self = [0; 32];
+}
+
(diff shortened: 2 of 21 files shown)


allisonkarlitskaya:
Some information about this change: for each of composefs-setup-root and cfsctl I tried building them with SHA-256 support (status quo), SHA-512 support (as per this PR currently) and then a hacked up version with support for both.  For `cfsctl` for example I renamed `main()` to `cfsctl()` and added a generic type parameter and instantiated it twice (once for sha256, once for sha512) from the "real" main(), just to get an idea.

For `composefs-setup-root`, it's just the fs-verity ioctls and mounting code involved, which gets us the following results in terms of binary size with `--release`:
```
sha256 = 1573712,
sha512 = 1573744
both = 1578272
```

so there it's basically no difference at all.  ~4.5k increase, ~0.2%.

For `cfsctl` (which pulls in basically all of the code, including the heavily genericified in-memory filesystem image stuff) the numbers are more interesting:

```
sha256 = 6952272
sha512 = 6974072
both = 7228032
```

which is an increase of some ~275k or ~4%.  That's generics for you....

The size of cfsctl doesn't really worry me too much, to be honest, and the size of composefs-setup-root is barely affected.

I'd probably play this out by doing something like adding a `--hash-alg=[sha256|sha512]` flag to `cfsctl`, paying the 4% tax, and then having `composefs-setup-root` expect to see the `composefs=` kernel cmdline arg prefixed with either `sha256:` or sha512:`.

src/fsverity/hashvalue.rs
```diff
@@ -1,27 +1,121 @@
...
+        if bytes.len() != expected_size {
+            return Err(FromHexError::InvalidStringLength);
+        }
+        let mut result = Self::EMPTY;
+        result.as_mut_bytes()[0] = dirnum;
+        hex::decode_to_slice(bytes, &mut result.as_mut_bytes()[1..])?;
+        Ok(result)
+    }
+
+    fn from_object_pathname(pathname: impl AsRef<[u8]>) -> Result<Self, FromHexError> {
```
cgwalters, line 41: Not new but this one looks unit testable.

Also here and elsewhere it's more efficient to do the "helper function" pattern where there's a generic wrapper for a concrete function that takes a concrete `&[u8]` does the real work.

src/fsverity/hashvalue.rs
```diff
@@ -1,27 +1,121 @@
...
+        // We want to the trailing part of "....../xx/yyyyyy" where xxyyyyyy is our hex length
+        let min_size = 2 * size_of::<Self>() + 1;
+        let bytes = pathname.as_ref();
+        if bytes.len() < min_size {
+            return Err(FromHexError::InvalidStringLength);
+        }
+
+        let trailing = &bytes[bytes.len() - min_size..];
```
cgwalters, line 49: I think using https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_checked would be cleaner

cgwalters (commented):
We could probably also consider avoiding some of the runtime generics and doing dynamic dispatch too...

There's already a dynamic type here https://docs.rs/oci-spec/latest/oci_spec/image/struct.Digest.html in the depchain even.

On the ostree side personally I ended up feeling it was just cleaner to pass around values as hex strings instead of binary actually.

cgwalters (approved):
Anyways looks sane to me but I obviously only skimmed things

src/fsverity/hashvalue.rs
allisonkarlitskaya: I was looking for this one!

allisonkarlitskaya:
I might slowly be coming around to the "just pass strings around" idea except that it involves a lot of extra allocations...

allisonkarlitskaya:
One thing worth thinking about, though: the different hash types aren't just different lengths of string, but encode some extra information like the name of the algorithm and the fs-verity digest algorithm number... not that this would be a *huge* problem to overcome, mind you...

src/oci/mod.rs
```diff
@@ -48,21 +48,21 @@ struct ImageOp<'repo> {
     progress: MultiProgress,
 }
 
-fn sha256_from_descriptor(descriptor: &Descriptor) -> Result<Sha256HashValue> {
+fn sha256_from_descriptor(descriptor: &Descriptor) -> Result<Sha256Digest> {
```
allisonkarlitskaya, line 54: wrong commit, x4

src/image.rs
```diff
@@ -8,7 +8,9 @@ use std::{
 
 use thiserror::Error;
 
-use crate::fsverity::Sha256HashValue;
+use crate::fsverity::FsVerityHashValue;
+
+// canhaz https://github.com/rust-lang/rfcs/issues/424?
```
allisonkarlitskaya, line 13: This is mentioned in the commit message already, let's drop it in the code.

src/bin/cfsctl.rs
```diff
@@ -6,7 +6,7 @@ use clap::{Parser, Subcommand};
 use rustix::fs::CWD;
 
 use composefs::{
-    fsverity::{FsVerityHashValue, Sha256HashValue},
+    fsverity::{FsVerityHashValue, Sha512HashValue},
```
allisonkarlitskaya, line 9: drop this commit, obviously...

allisonkarlitskaya (commented):
A few fixups here but on second reading I'm pretty happy with it, to be honest.

We might indeed to go for a "hex string" sort of approach in the future, or at least some more dynamic sort of thing with a boxed enum, but the code changes in this commit (particularly the untangling of the different uses of SHA-256) will make it a whole lot easier for us to approach that in the future... we could even try a FsVerityHashValue implementation as a string, with some adjustments...

cgwalters:
> I might slowly be coming around to the "just pass strings around" idea except that it involves a lot of extra allocations...

Passing `String` always involves a heap allocation yes but it's possible to stack allocate hex strings too though doing so via `[char; 64]` is obviously extra expensive and I think maybe we want an array of https://docs.rs/ascii/latest/ascii/ or so? There's a bit of a crate-interoperability thing here too, at one point ascii types were going to be in std, but they're not. In theory `hex::encode` should promise to output ascii (at a minimum).

allisonkarlitskaya:
Also worth mentioning, the way this works now, we end up with a lot of hash values in memory stored inline inside of our LeafContent structs.  Those structs are quite big because of this, and we pay that price for non-regular/external leaf files: internal files, chardevs, blockdevs, fifos, sockets, symlinks.  But all of those things (with the exception of symlinks) are exceedingly rare in container images.... and for symlinks maybe we could optimize small inline links or something *shrug*.  By all of this I mean: switching to boxes would be less efficient at runtime.

Brings me to another point, though: we need to adjust (double) our "max inline file size" when we're in SHA-512 mode...

src/fsverity/hashvalue.rs
allisonkarlitskaya: Adding unit tests found an actual bug, thanks!

src/fsverity/hashvalue.rs
allisonkarlitskaya: Ah this doesn't really help that much because it can't do calculations relative to the end, which means that we need to first check that the length is sufficient before doing the subtraction anyway...

src/oci/image.rs
```diff
@@ -16,7 +16,10 @@ use crate::{
     selabel::selabel,
 };
 
-pub fn process_entry(filesystem: &mut FileSystem, entry: TarEntry) -> Result<()> {
+pub fn process_entry<H: FsVerityHashValue>(
```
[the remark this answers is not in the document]
allisonkarlitskaya: Changed some `H:` cases... but also kept it for some lower-level APIs where it's not clear that we're using it for an object ID.

src/util.rs
```diff
@@ -83,14 +81,17 @@ pub(crate) async fn read_exactish_async(
     Ok(true)
 }
 
-/// Parse a string containing a SHA256 digest in hexidecimal form into a Sha256HashValue.
+/// A utility type representing a SHA-256 digest in binary.
+pub type Sha256Digest = [u8; 32];
```
jeckersb, line 85: This should probably be a proper type instead of an alias, especially since it's exposed in the public API.

The somewhat-silly reasoning:  I am selfish and my lsp with rust-analyzer (probably for others too not using insane emacs setups?) will just erase `Sha256Digest` everywhere in type annotations and show `[u8; 32]` in its place, which is annoying and loses context.

The more constructive reasoning: exposing it publicly as `[u8; 32]` is going to limit flexibility in the future.  Maybe you want to attach extra metadata to the digest.  Or maybe you want to `impl` some traits for `Sha256Digest`.  As is, you couldn't for example `impl std::fmt::Display` for it.

It's mildly more annoying since you'd have to do something like `Sha256Digest::from([0u8; 32])` and `digest.bytes()` or `digest.0` or similar all over the place, but it gives you a lot more flexibility in the future without needing to break semver.

Edit:  As I read the whole PR some more, I see you did basically exactly this in `fsverity::hashvalue::{Sha256HashValue,Sha512HashValue}`.

src/oci/mod.rs
```diff
@@ -41,31 +44,31 @@ pub fn ls_layer(repo: &Repository, name: &str) -> Result<()> {
...
-fn sha256_from_digest(digest: &str) -> Result<Sha256HashValue> {
+fn sha256_from_digest(digest: &str) -> Result<Sha256Digest> {
     match digest.strip_prefix("sha256:") {
         Some(rest) => Ok(parse_sha256(rest)?),
         None => bail!("Manifest has non-sha256 digest"),
     }
 }
 
-type ContentAndVerity = (Sha256HashValue, Sha256HashValue);
+type ContentAndVerity<ObjectID> = (Sha256Digest, ObjectID);
```
jeckersb, line 68: This one isn't as big of a deal being a type alias since it's not public, but worth consideration still.

jeckersb (commented):
I'll probably look at this a bit more tomorrow, but wanted to post what I have thus far before I take off for the day.  This on balance looks super good to me.  Lots of code-motion things that makes me go cross-eyed so I might miss some stuff in there.  Just one real concern so far with public API exposure.

PR #108, opened 2025-04-17
fsverity: call sync_data (fdatasync) in test code

### LINKED ISSUE #106: fsverity enablement can fail if a forked process temporarily cloned the writable fd
(edited by @cgwalters - original content below)

# enabling fsverity is racy when there's a possibility for concurrent forked processes

TL;DR: If a forked process captures a writable copy of the fd, we will fail to enable fsverity after reopening it readonly.

Our test suite today forks a process in one of the tests, but in general (as a library crate) we can't know what calling processes may do, and we should be robust against

## Option: Loop and retry

This will work just fine as long as a long-running `fork()` is not used. This is my preferred solution, as I think basically no one does a long running `fork()` (without a relatively-quickly-following `exec`) in Rust because we have threads and async.

## Option: Copy-and-retry

Here instead of looping, we create a *new* FD (using `copy_file_range` from the old one which should reflink if possible). The advantage of this is it's robust against long running `fork()`.

## Option: Hybrid

Here we implement both of the above; loop a few times, and if that fails fall back to a copying loop

## Option: fsverity helper process

We could use something like [procspawn](https://docs.rs/procspawn/latest/procspawn/) to create our own helper process, and do fsverity enablement there.

## Option (long term): close-on-fork

https://www.austingroupbugs.net/view.php?id=1318

## Option: (long term): io-uring and fixed files

https://github.com/containers/composefs-rs/issues/106#issuecomment-2877753944

This doesn't yet work but could be made to do so.

Original issue follows:

--- 

If you run the tests a bunch you sometimes get this:

```
---- fsverity::tests::crosscheck_interesting_cases stdout ----

thread 'fsverity::tests::crosscheck_interesting_cases' panicked at src/fsverity/mod.rs:343:37:
called `Result::unwrap()` on an `Err` value: Io(Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
---- fsverity::tests::crosscheck_interesting_cases stdout ----

thread 'fsverity::tests::crosscheck_interesting_cases' panicked at src/fsverity/mod.rs:343:37:
called `Result::unwrap()` on an `Err` value: Io(Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" })
stack backtrace:
   0: rust_begin_unwind
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/core/src/result.rs:1704:5
   3: core::result::Result<T,E>::unwrap
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/core/src/result.rs:1109:23
   4: composefs::fsverity::tests::crosscheck_interesting_cases::assert_kernel_equal
             at ./src/fsverity/mod.rs:343:13
   5: composefs::fsverity::tests::crosscheck_interesting_cases
             at ./src/fsverity/mod.rs:350:13
   6: composefs::fsverity::tests::crosscheck_interesting_cases::{{closure}}
             at ./src/fsverity/mod.rs:312:38
   7: core::ops::function::FnOnce::call_once
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/core/src/ops/function.rs:250:5
   8: core::ops::function::FnOnce::call_once
             at /builddir/build/BUILD/rust-1.85.1-build/rustc-1.85.1-src/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    fsverity::tests::crosscheck_interesting_cases
```

That seems like some kind of a race between dropping the writable file descriptor and enabling verity, but we're running single threaded, so we presumably always do things in the same order.  I wonder if there's a race in the kernel there where we might have to wait for the fd to be dropped "internally" or so?

This might then also impact the Repository code, but maybe not: we use `O_TMPFILE` there instead of the `tempfile` module...

jeckersb:
This partially fixes #106.

There's a race between the read-write file descriptor being closed,
and the subsequent ioctl to enable fsverify on the read-only copy.
Sometimes the ioctl will return ETXTBSY.  Syncing the data here seems
to be sufficient to resolve this race, but *only* on ext4.  On btrfs,
this is not sufficient.  I *can* fix this on btrfs by doing a
`syncfs()`, but we are trying to avoid doing that here.

### DIFF
--- a/src/fsverity/mod.rs
+++ b/src/fsverity/mod.rs
@@ -181,6 +181,7 @@ mod tests {
     fn rdonly_file_with(data: &[u8]) -> OwnedFd {
         let mut file = tempfile();
         file.write_all(data).unwrap();
+        file.sync_data().unwrap();
         let fd = open(
             proc_self_fd(&file),
             OFlags::RDONLY | OFlags::CLOEXEC,


allisonkarlitskaya (approved):
thanks!

cgwalters:
Hmm I don't think we've seen this in any other codebases using fsverity. Could this be because we're trying to enable fsverity on an unlinked file in this testing code path? That's a pretty weird thing to do.

cgwalters:
Yeah just skimming the kernel code quickly I think that's the problem, the fsverity calls end up in `filemap_write_and_wait_range()` which may not succeed on such an unlinked file.

PR #109, opened 2025-04-17
Use automatic links in docstrings

jeckersb:
Otherwise `cargo doc` will complain with `warning: this URL is not a hyperlink`


allisonkarlitskaya (approved):
I've seen warnings like that go by before.  Thanks!

PR #111, opened 2025-04-21
Add multithread `cfsctl oci pull`

allisonkarlitskaya:
For images with a large number of layers (as is the case for most bootc images) this is a massive win.  I used this as a test:

```sh
skopeo copy docker://quay.io/fedora/fedora-silverblue:42 oci:$HOME/oci/silverblue-42
```

and then

```sh
cfsctl oci pull oci:$HOME/oci/silverblue-42
```

And noticed an improvement from ~90s to ~9s.

Unfortunately the situation isn't much improved for images with a small number of layers (as is the case in our own `examples/` directory).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ hex = "0.4.0"
 indicatif = { version = "0.17.0", features = ["tokio"] }
 log = "0.4.8"
 oci-spec = "0.7.0"
+once_cell = { version = "1.21.3", default-features = false }
 regex-automata = { version = "0.4.4", default-features = false }
 rustix = { version = "1.0.0", features = ["fs", "mount", "process"] }
 serde = "1.0.145"
@@ -41,7 +42,6 @@ zstd = "0.13.0"
 
 [dev-dependencies]
 insta = "1.42.2"
-once_cell = "1.21.3"
 similar-asserts = "1.7.0"
 test-with = { version = "0.14", default-features = false, features = ["executable", "runtime"] }
 tokio-test = "0.4.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ rhel9 = ['pre-6.15']
 anyhow = { version = "1.0.87", default-features = false }
 async-compression = { version = "0.4.0", default-features = false, features = ["tokio", "zstd", "gzip"] }
 clap = { version = "4.0.1", default-features = false, features = ["std", "help", "usage", "derive"] }
-containers-image-proxy = "0.7.0"
+containers-image-proxy = "0.7.1"
 env_logger = "0.11.0"
 hex = "0.4.0"
 indicatif = { version = "0.17.0", features = ["tokio"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ sha2 = "0.10.1"
 tar = { version = "0.4.38", default-features = false }
 tempfile = "3.8.0"
 thiserror = "2.0.0"
-tokio = "1.24.2"
+tokio = { version = "1.24.2", features = ["rt-multi-thread"] }
 toml = "0.8.0"
 xxhash-rust = { version = "0.8.2", features = ["xxh32"] }
 zerocopy = { version = "0.8.0", features = ["derive", "std"] }
--- a/src/oci/mod.rs
+++ b/src/oci/mod.rs
@@ -1,4 +1,4 @@
-use std::process::Command;
+use std::{cmp::Reverse, process::Command, thread::available_parallelism};
 
 pub mod image;
 pub mod tar;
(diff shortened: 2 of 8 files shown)


allisonkarlitskaya:
An earlier version of this PR had a `objects: OnceCell<Arc<OwnedFd>`` field and an internal separate method for ensure_objects_in_dir() that took an `Arc<OwnedFd>` of the objects directory and did its work inside of that... I wonder if this approach is nicer in general (although it doesn't help solve the weird lifetime issues associated with the stream writer or the oci imageop).

This is somewhere where my lack of experience is definitely showing...

src/oci/mod.rs
```diff
@@ -172,14 +172,31 @@ impl<'repo, ObjectID: FsVerityHashValue> ImageOp<'repo, ObjectID> {
             let raw_config = config?;
             let config = ImageConfiguration::from_reader(&raw_config[..])?;
 
+            // We want to sort the layers based on size so we can get started on the big layers
+            // first.  The last thing we want is to start on the biggest layer right at the end.
+            let mut layers: Vec<_> = zip(manifest_layers, config.rootfs().diff_ids()).collect();
+            layers.sort_by_key(|(mld, ..)| Reverse(mld.size()));
+
+            // Bound the number of tasks to the available parallelism.
+            let threads = available_parallelism()?;
```
cgwalters, line 181: This case is tricky because we're intermixing CPU and I/O work; one problem I've seen in the past is running on large CPU count servers (e.g. 64+) but with more limited I/O bandwidth. In those cases we end up with 64 threads competing pointlessly for more limited I/O.

There's no convenient way to get any estimate for I/O parallelism unfortunately that I know of but in some equivalent places I've capped at an arbitrary number like 4.

src/oci/mod.rs
```diff
@@ -172,14 +172,31 @@ impl<'repo, ObjectID: FsVerityHashValue> ImageOp<'repo, ObjectID> {
...
+            let sem = Arc::new(Semaphore::new(threads.into()));
+            let mut entries = vec![];
+            for (mld, diff_id) in layers {
+                let self_ = Arc::clone(self);
+                let permit = Arc::clone(&sem).acquire_owned().await?;
```
cgwalters, line 186: Ah yeah using a semaphore for this makes sense.

src/oci/mod.rs
```diff
@@ -172,14 +172,31 @@ impl<'repo, ObjectID: FsVerityHashValue> ImageOp<'repo, ObjectID> {
...
+                let layer_sha256 = sha256_from_digest(diff_id)?;
+                let descriptor = mld.clone();
+                let future = tokio::spawn(async move {
```
cgwalters, line 189: See https://github.com/bootc-dev/bootc/blob/2de8e0d23fb89bed76722ff1466614afacec64b3/lib/src/fsck.rs#L195 which uses a JoinSet which is designed for this, especially that it enforces structured concurrency.

Cargo.toml
```diff
@@ -20,28 +20,28 @@ rhel9 = ['pre-6.15']
...
 async-compression = { version = "0.4.0", default-features = false, features = ["tokio", "zstd", "gzip"] }
 clap = { version = "4.0.1", default-features = false, features = ["std", "help", "usage", "derive"] }
-containers-image-proxy = "0.7.0"
+containers-image-proxy = { git = "https://github.com/containers/containers-image-proxy-rs", rev = "0bb03cea8468c24bae47bfcfe16c44973cc2d3a6" }
 env_logger = "0.11.0"
 hex = "0.4.0"
 indicatif = { version = "0.17.0", features = ["tokio"] }
 log = "0.4.8"
 oci-spec = "0.7.0"
+once_cell = { version = "1.21.3", default-features = false }
```
cgwalters, line 29: This change looks spurious? Also there's no reason to use the external crate since the functionality got merged into std.

allisonkarlitskaya: I mentioned that in the commit message: https://github.com/containers/composefs-rs/pull/111/commits/f286208a104ee3350043dcff7201885deb49d3fd

> We use the external `once_cell::` crate (which we already had as a -devel dependency) because the .try() version of the API is not yet stable in the standard library.

src/oci/mod.rs
allisonkarlitskaya: Thanks for the pointer.  I used this for the HTTP downloader.

src/oci/mod.rs
allisonkarlitskaya: Ya, I think I'll stick with it here...

allisonkarlitskaya:
The only thing blocking this in my opinion is that it would be nice to have a new containers-image-proxy-rs release.  Also: I might drop the await on the driver as we discussed in https://github.com/containers/containers-image-proxy-rs/issues/80#issuecomment-2830692496

PR #112, opened 2025-04-30
new boot API

allisonkarlitskaya:
This should make composefs-rs a whole lot easier to use from bootc.

### DIFF
--- a/examples/bls/Containerfile.arch
+++ b/examples/bls/Containerfile.arch
@@ -22,9 +22,5 @@ RUN <<EOF
     systemctl enable systemd-networkd systemd-resolved sshd
     passwd -d root
     mkdir /sysroot
-    mkdir /composefs-meta
-    mv /boot /composefs-meta
-    mkdir /boot
 EOF
 COPY cfsctl /usr/bin
-RUN true
--- a/examples/bls/Containerfile.arch
+++ b/examples/bls/Containerfile.arch
@@ -1,18 +1,30 @@
 FROM archlinux AS base
 COPY extra /
-COPY cfsctl /usr/bin
 RUN <<EOF
     set -eux
+
     touch /etc/machine-id
     echo 'root=/dev/vda2' > /etc/kernel/cmdline
+
     pacman -Syu --noconfirm
-    pacman -Sy --noconfirm skopeo composefs strace dosfstools linux mkinitcpio btrfs-progs openssh
+    pacman -Sy --noconfirm \
+        btrfs-progs \
+        composefs \
+        dosfstools \
+        linux \
+        mkinitcpio \
+        openssh \
+        skopeo \
+        strace
+
+    kernel-install add "$(ls /usr/lib/modules)" /usr/lib/modules/"$(ls /usr/lib/modules)"/vmlinuz
+
     systemctl enable systemd-networkd systemd-resolved sshd
     passwd -d root
     mkdir /sysroot
-    kernel-install add "$(ls /usr/lib/modules)" /usr/lib/modules/"$(ls /usr/lib/modules)"/vmlinuz
     mkdir /composefs-meta
     mv /boot /composefs-meta
     mkdir /boot
 EOF
+COPY cfsctl /usr/bin
 RUN true
--- a/src/bin/cfsctl.rs
+++ b/src/bin/cfsctl.rs
@@ -257,7 +257,7 @@ async fn main() -> Result<()> {
             bootable,
             stat_root,
         } => {
-            let mut fs = composefs::fs::read_from_path(path, Some(&repo), stat_root)?;
+            let mut fs = composefs::fs::read_filesystem(CWD, path, Some(&repo), stat_root)?;
             if bootable {
                 fs.transform_for_boot(&repo)?;
             }
@@ -270,7 +270,7 @@ async fn main() -> Result<()> {
             stat_root,
             ref image_name,
         } => {
-            let mut fs = composefs::fs::read_from_path(path, Some(&repo), stat_root)?;
+            let mut fs = composefs::fs::read_filesystem(CWD, path, Some(&repo), stat_root)?;
             if bootable {
                 fs.transform_for_boot(&repo)?;
             }
@@ -282,7 +282,7 @@ async fn main() -> Result<()> {
             bootable,
             stat_root,
         } => {
-            let mut fs = composefs::fs::read_from_path(path, Some(&repo), stat_root)?;
+            let mut fs = composefs::fs::read_filesystem(CWD, path, Some(&repo), stat_root)?;
             if bootable {
                 fs.transform_for_boot(&repo)?;
             }
--- a/src/bin/cfsctl.rs
+++ b/src/bin/cfsctl.rs
@@ -39,20 +39,31 @@ enum OciCommand {
         /// the name of the stream
         name: String,
     },
-    CreateDumpfile {
-        layers: Vec<String>,
+    Dump {
+        config_name: String,
+        config_verity: Option<String>,
     },
     Pull {
         image: String,
         name: Option<String>,
     },
+    ComputeId {
+        config_name: String,
+        config_verity: Option<String>,
+        #[clap(long)]
+        bootable: bool,
+    },
     CreateImage {
-        config: String,
-        name: Option<String>,
+        config_name: String,
+        config_verity: Option<String>,
+        #[clap(long)]
+        bootable: bool,
+        #[clap(long)]
+        image_name: Option<String>,
     },
     Seal {
-        name: String,
-        verity: Option<String>,
+        config_name: String,
+        config_verity: Option<String>,
     },
     Mount {
         name: String,
@@ -62,8 +73,10 @@ enum OciCommand {
         name: String,
     },
     PrepareBoot {
-        name: String,
-        bootdir: Option<PathBuf>,
+        config_name: String,
+        config_verity: Option<String>,
+        #[clap(long, default_value = "/boot")]
+        bootdir: PathBuf,
     },
 }
 
@@ -97,13 +110,36 @@ enum Command {
     },
     CreateImage {
         path: PathBuf,
+        #[clap(long)]
+        bootable: bool,
+        #[clap(long)]
+        stat_root: bool,
+        image_name: Option<String>,
+    },
+    ComputeId {
+        path: PathBuf,
+        #[clap(long)]
+        bootable: bool,
+        #[clap(long)]
+        stat_root: bool,
     },
     CreateDumpfile {
         path: PathBuf,
+        #[clap(long)]
+        bootable: bool,
+        #[clap(long)]
+        stat_root: bool,
     },
     ImageObjects {
         name: String,
     },
+}
+
+fn verity_opt(opt: &Option<String>) -> Result<Option<Sha256HashValue>> {
+    Ok(match opt {
+        Some(value) => Some(FsVerityHashValue::from_hex(value)?),
+        None => None,
+    })
 }
 
 #[tokio::main]
@@ -112,7 +148,7 @@ async fn main() -> Result<()> {
 
     let args = App::parse();
 
-    let repo: Repository<Sha256HashValue> = (if let Some(path) = args.repo {
+    let repo: Repository<Sha256HashValue> = (if let Some(path) = &args.repo {
         Repository::open_path(CWD, path)
     } else if args.system {
         Repository::open_system()
@@ -136,7 +172,7 @@ async fn main() -> Result<()> {
         }
         Command::ImportImage { reference } => {
             let image_id = repo.import_image(&reference, &mut std::io::stdin())?;
-            println!("{}", image_id.to_hex());
+            println!("{}", image_id.to_id());
         }
         Command::Oci { cmd: oci_cmd } => match oci_cmd {
             OciCommand::ImportLayer { name, sha256 } => {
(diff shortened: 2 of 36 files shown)


src/bin/cfsctl.rs
```diff
@@ -179,25 +251,88 @@ async fn main() -> Result<()> {
...
+                ref bootdir,
+                ref entry_id,
+                ref cmdline,
+            } => {
+                let verity = verity_opt(config_verity)?;
+                let mut fs = oci::image::create_filesystem(&repo, config_name, verity.as_ref())?;
+                let entries = fs.transform_for_boot(&repo)?;
+                let id = fs.commit_image(&repo, None)?;
+
+                let Some(entry) = entries.into_iter().next() else {
```
cgwalters, line 266: We discussed this in a live chat, my preference here is that when we create a sealed image we also by default prune the old `vmlinuz+initramfs.img`. Yes, the RPM database would say it's missing, but I don't see that as a serious problem. We already have other related issues with moving `/var/lib/selinux` back to `/etc`.

The advantage of pruning is we don't duplicate the kernel binary, and it's very clear - there's one kernel in the image.

In theory of course we could have the official `prepare-boot` flow run through `rpmbuild` + `rpm -Uvh` (for fedora-bootc derivatives); actually really on this topic it's important to note there's already `kernel-uki-virt` and so we could suggest installing `kernel-uki-stub` or something that we then mutate?

src/bin/cfsctl.rs
```diff
@@ -179,25 +251,88 @@ async fn main() -> Result<()> {
...
+                    entry_id.as_deref(),
+                    &cmdline_refs,
+                )?;
+
+                let state = args
+                    .repo
+                    .as_ref()
+                    .map(|p: &PathBuf| p.parent().unwrap())
+                    .unwrap_or(Path::new("/sysroot"))
+                    .join("state")
```
cgwalters, line 285: Surprised there isn't a const for this somewhere

src/bin/cfsctl.rs
```diff
@@ -179,25 +251,88 @@ async fn main() -> Result<()> {
...
+                    .join(id.to_hex());
+
+                create_dir_all(state.join("var"))?;
+                create_dir_all(state.join("etc/upper"))?;
```
cgwalters, line 289: Ditto

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+    repo: &Repository<ObjectID>,
+) -> Result<Box<[u8]>> {
+    match file {
+        RegularFile::Inline(data) => Ok(data.clone()),
+        RegularFile::External(id, size) => {
+            // TODO: mmap niceness?
+            // TODO: verify size: it could be faked
+            let mut data = Vec::with_capacity(*size as usize);
+            // TODO: read loop...
+            read(repo.open_object(id)?, spare_capacity(&mut data)).unwrap();
```
cgwalters, line 152: Why the unwrap?

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+
+        // 3. Append new "options" line with our argument
+        self.lines.push(format!("options {arg}"));
+    }
+
+    /// Adjusts the kernel command-line arguments by adding a composefs= parameter (if appropriate)
+    /// and adding additional arguments, as requested.
+    pub fn adjust_cmdline(&mut self, composefs: Option<&str>, extra: &[&str]) {
+        if let Some(id) = composefs {
+            self.add_cmdline(&format!("composefs={id}"));
```
cgwalters, line 128: Not a new thing obviously but per discussion I think we should move the linked digest to a file in the initramfs by default, with a kernel argument as override.

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+use crate::{
+    fsverity::FsVerityHashValue,
+    repository::Repository,
+    tree::{Directory, FileSystem, ImageError, Inode, LeafContent, RegularFile},
+};
+
+fn split_cmdline(cmdline: &str) -> impl Iterator<Item = &str> {
+    let mut in_quotes = false;
+
+    cmdline.split(move |c: char| {
```
cgwalters, line 24: Huh I hadn't seen the model of passing a stateful closure to split, but it does look pretty clean in this case.

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+
+        match root.get_directory("/usr/lib/modules".as_ref()) {
+            Ok(modules_dir) => {
+                for (kver, inode) in modules_dir.entries() {
+                    let Inode::Directory(dir) = inode else {
+                        continue;
+                    };
+
+                    for (filename, inode) in dir.entries() {
+                        if !filename.as_bytes().ends_with(b"b.efi") {
```
cgwalters, line 336: Is the extra `b` here a typo?

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+            self.add_cmdline(item);
+        }
+    }
+}
+
+// TODO: put this in a better spot
+// Reminder: it only makes sense to do boot stuff on a legit repository because we need to read the
+// boot loader entry content.  But: wouldn't it be nice to use the trait convenience wrappers
+// anyway?
+fn read_file<ObjectID: FsVerityHashValue>(
```
cgwalters, line 141: One question to ask too is why don't we just mount the composefs and read from it via standard file APIs?

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+    // Relocates boot resources.
+    //
+    // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
+    // the file into the given 'entry_id' pathname and rename the entry file itself to
+    // "{entry_id}.conf".
+    pub fn relocate(&mut self, entry_id: &str) {
+        self.filename = Box::from(format!("{entry_id}.conf").as_ref());
+        for line in &mut self.entry.lines {
+            for key in ["linux", "initrd", "efi"] {
+                if let Some(value) = strip_ble_key(line, key) {
```
cgwalters, line 176: The nesting is pretty deep, around this point I tend to do e.g.

`let Some(value) = strip_ble_key(line, key) else { continue };`

src/bootloader.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// This is the basename of the bootloader entry .conf file
+    pub filename: Box<OsStr>,
+    pub entry: BootLoaderEntryFile,
+    pub files: HashMap<Box<str>, RegularFile<ObjectID>>,
+}
+
+impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
+    // Relocates boot resources.
+    //
+    // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
```
cgwalters, line 169: And just to record this, we discussed this in person but there's basically two possible models here:

- Have bootloader entries in `/boot` in the image which we parse/extract (this code), then we can mask out `/boot` entirely when computing the final digest
- Just use `/usr/lib/modules/$kver`, the container image's `/boot` is empty (which aligns with how bootc encourages it today, *and* AFAIK how the uapi-spec is intending it for DDIs) - but for the sealed UKI path we need to mask out `/usr/lib/modules/$kver/foo.efi` (ref https://uapi-group.org/specifications/specs/unified_kernel_image/#locations-for-distribution-built-ukis-installed-by-package-managers )

I guess the bigger picture question to me here is under what scenarios will we have e.g. type1 non-sealed, which this code is trying to handle? Why wouldn't we just encourage people to use the `/usr/lib/modules/$kver` directory in "unsealed" cases?

The sealed UKI+composefs case is special, no doubt about that and I think it's certainly legitimate to say "we'll just mask off all of `/boot`". Although, since we're creating the thing we could just as easily say it's `/usr/lib/composefs-uki` or `/composefs-uki` or whatever too.

src/bin/cfsctl.rs
allisonkarlitskaya: My 2¢: we should modify the image as little as possible.

If the person creating the image wants to delete the files then they absolutely can.  If the files are in the image that we've already downloaded them and have them in the repository.   Having them in the image is no extra overhead.

src/bin/cfsctl.rs
allisonkarlitskaya: For "state"?  I suppose we could add this.  To be honest, I consider this chunk of the code to be little more than an example...

src/bootloader.rs
allisonkarlitskaya: "Move fast and do broken things".

I'll definitely fix this.  Thanks for the catch :)

src/bootloader.rs
allisonkarlitskaya: Which would be fine for sealed images but complicates the story for people who want to derive an image to add a package or whatever.  They need a mechanism to change the root filesystem without rebuilding the initramfs.

src/bootloader.rs
allisonkarlitskaya: Ya!  This was a really nice discovery for me too, thanks for noticing.  I may have been a tiny bit proud of this code after having written it. :)

src/bootloader.rs
allisonkarlitskaya: At this point the image doesn't exist yet.  I am working on a fuse backend that can make the in-memory filesystem tree visible though and it would indeed be interesting to use for cases like this...

In general, though, I prefer to keep these operations as "pure" as possible.  Fewer roundtrips through multiple processes running via the kernel means fewer chances for non-determinism.  Ideally we'd be able to support a build flow that involves going from a given container sha ID to a deterministic qcow every time the same...

I actually think a lot about how `cap-std` could be nice for some of these things if `Dir` were a trait rather than a struct.  I'd love it if we could backend something like an `impl Dir` (with a similar -- but smaller -- set of operations to `cap-std::fs::Dir`) onto either the real filesystem or an in-memory tree backed with an object store.  I plan to do a bit of work in this area in the medium-term...

src/bootloader.rs
allisonkarlitskaya: Cleaned this up by just using File::from().read_to_end() on the fd and comparing the length after.  It's a lot neater.

src/bootloader.rs
```diff
@@ -0,0 +1,411 @@
...
+            self.add_cmdline(&format!("composefs={id}"));
+        }
+
+        for item in extra {
+            self.add_cmdline(item);
+        }
+    }
+}
+
+pub(crate) fn read_file<ObjectID: FsVerityHashValue>(
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's a bit weird because `Repository` doesn't know about `tree::*` and vice versa.  At one point I made a trait impl over `(&Repository, &tree::Directory)`....

Also: I have some vague future plans to have non-Repository object stores supported (in memory, "null", etc)...

I also have some other plans to eventually add an mmap interface at the Repository level.

in short: i think this is follow-up stuff...

src/bootloader.rs
cgwalters: > It's also the medium of interchange for systemd-boot which we're going to want to support anyway. In fact, the /usr/lib/modules/kver/vmlinuz case is supported by converting it into a type 1 entry and writing it out...

Yes of course! But that operation happens on the client. That's the main focus of the debate here right? As opposed to embedding it in containers.

cgwalters (approved):
There's a lot of preparatory cleanups and such here that probably would have made sense as a distinct PR that could easily have gone in first, before the later, deeper bootloader stuff.

I am still uncertain about some of that but, we also don't need to block on it from my PoV.

examples/bls/Containerfile.arch
```diff
@@ -1,18 +1,26 @@
...
+        btrfs-progs \
+        composefs \
+        dosfstools \
+        linux \
+        mkinitcpio \
+        openssh \
+        skopeo \
+        strace
+
+    kernel-install add "$(ls /usr/lib/modules)" /usr/lib/modules/"$(ls /usr/lib/modules)"/vmlinuz
```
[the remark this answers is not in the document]
allisonkarlitskaya: Interesting.  I mean, this particular commit was just moving around something that was already there.

But you're right.  Now that we have the vmlinuz/initramfs.img support we might theoretically look at getting rid of this.  Or since this is the "bls" directory, consider introducing another examples dir that does the vmlinuz/initramfs.img way.

My ultimate plan is to add some examples that directly pull the bootc Fedora image and run that in CI, btw.  After the fixes in this PR, the only thing that's missing there at present (I think) is having composefs-setup-root in the initramfs... but I have some very evil ideas about that......

src/bootloader.rs
cgwalters: @keszybz for the record the debate here is roughly: should container images have BLS entries directly in `/boot` or should they just have `/usr/lib/modules/$kver` (with out of band formats for kargs in the generic vendor OS case).

examples/bls/Containerfile.arch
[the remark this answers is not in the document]
allisonkarlitskaya: My idea is more evil :)

src/bootloader.rs
keszybz: The driving idea for `/boot` in the upstream systemd model is that it is a shared resource. There is one ESP partition that is used and managed by one or more Linux systems and the firmware and Microsoft Windows. This means that we cannot use fixed naming of files and directories under `/boot`, but we have per-installation directories and all that stuff, so we carefully copy files over into `/boot`, possibly doing cleanup of older files. Making `/boot` part of the image is largely incompatible with this approach.

One drawback of the approach of shipping files in `/usr/lib/modules` is that they are duplicated and they need to be copied. But in practice this duplication doesn't seem to be a big problem, because the kernel is not big enough compared to the full installation to matter. The modules and the firmware are bigger, but we want those to be present in the installation, because they can be loaded dynamically.

In final installation, we don't want the images to ship stuff directly in `/boot`. So I *think* we don't want them to be there in the image either, but maybe some more complicated scheme where the files are there in the image but then are masked in the installation would be possible. I don't have an opinion on this, because I don't know the bootc internals.

PR #113, opened 2025-04-30
.github/workflows: drop podman install hack

allisonkarlitskaya:
Ubuntu 24.04 has gotten an updated version of podman (supporting "here docs") since we added this hack.  Let's drop it.


allisonkarlitskaya:
Nope.  Not yet, apparently.

PR #115, opened 2025-05-06
Update return type of `oci::pull`

Johan-Liebert1:
Move the printing of SHA256 and verity from the library to cfsctl binary.

### DIFF
--- a/src/oci/mod.rs
+++ b/src/oci/mod.rs
@@ -238,7 +238,7 @@ pub async fn pull(
     repo: &Arc<Repository<impl FsVerityHashValue>>,
     imgref: &str,
     reference: Option<&str>,
-) -> Result<()> {
+) -> Result<(Sha256Digest, impl FsVerityHashValue)> {
     let op = Arc::new(ImageOp::new(repo, imgref).await?);
     let (sha256, id) = op
         .pull()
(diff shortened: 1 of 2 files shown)


src/fsverity/hashvalue.rs
```diff
@@ -95,7 +95,7 @@ impl fmt::Debug for Sha512HashValue {
 
 #[derive(Clone, Eq, FromBytes, Hash, Immutable, IntoBytes, KnownLayout, PartialEq, Unaligned)]
 #[repr(C)]
-pub struct Sha256HashValue([u8; 32]);
+pub struct Sha256HashValue(pub(crate) [u8; 32]);
```
allisonkarlitskaya, line 98: Please drop this change.  See below.

src/oci/mod.rs
```diff
@@ -238,7 +239,7 @@ pub async fn pull(
     repo: &Arc<Repository<impl FsVerityHashValue>>,
     imgref: &str,
     reference: Option<&str>,
-) -> Result<()> {
+) -> Result<(Sha256HashValue, impl FsVerityHashValue)> {
```
allisonkarlitskaya, line 242: This should be `util::Sha256Digest` now.  We changed this a while ago: the values from `fsverity::` are now only used for verity digests.

allisonkarlitskaya (changes requested):

src/oci/mod.rs
Johan-Liebert1: Ah, I see. Updated

allisonkarlitskaya (approved):
Thanks!

PR #116, opened 2025-05-06
examples: add bootc/ directory

allisonkarlitskaya:
This builds a bootable image from a stock bootc image.

It works by appending composefs-setup-root to the existing initramfs by using a second initrd line in the boot loader entry.

This is an evil little hack but it was kinda fun and maybe we have some sort of use for this sort of thing in the short-term.  I'm not sure this is worth merging....

### DIFF
--- a/examples/bootc/build
+++ b/examples/bootc/build
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+set -eux
+
+cd "${0%/*}"
+image="${1:-docker://quay.io/fedora/fedora-bootc:42}"
+
+cargo build --release
+cp ../../target/release/cfsctl .
+mkdir -p initrd/usr/bin
+cp ../../target/release/composefs-setup-root initrd/usr/bin
+
+# rm -rf tmp
+../common/install-systemd-boot
+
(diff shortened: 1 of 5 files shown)


cgwalters (approved):
Looks ok to me but I hope we can make faster progress on the inverse side of doing this from the bootc git repo and integration there.

In particular continuing looming issues like delivery of cfsctl or whether that gets wrapped, overall integration with bootc install (vs the kargs here) etc.

travier:
This is indeed a workaround but I don't think that's were we should go as this notably won't work for the UKI use case. We should require the image to include all the needed bits.

cgwalters:
> We should require the image to include all the needed bits.

OK I think the simplest thing is:

- Just keep updating the vendored composefs-rs in bootc (happening in https://github.com/bootc-dev/bootc/pull/1314 at least which can be split out)
- Add a feature flag to bootc which also installs composefs-prepare-root and such
- Enable that feature flag for https://gitlab.com/fedora/bootc/base-images-dev which also tracks git main

or so?

PR #117, opened 2025-05-07
Only use `podman unshare` outside of a container

Johan-Liebert1:
Bootc runs inside a container and pulling from `containers-storage` transport fails.

This ensures that for pulling `containers-storage` images, we only use `podman unshare skopeo` if we're not already running inside a container.

### DIFF
--- a/src/oci/mod.rs
+++ b/src/oci/mod.rs
@@ -13,6 +13,7 @@ use oci_spec::image::{Descriptor, ImageConfiguration, ImageManifest, MediaType};
 use sha2::{Digest, Sha256};
 use tokio::{io::AsyncReadExt, sync::Semaphore};
 
+use crate::util::is_podman_container;
 use crate::{
     fsverity::FsVerityHashValue,
     oci::tar::{get_entry, split_async},
@@ -69,7 +70,8 @@ type ContentAndVerity<ObjectID> = (Sha256Digest, ObjectID);
 impl<ObjectID: FsVerityHashValue> ImageOp<ObjectID> {
     async fn new(repo: &Arc<Repository<ObjectID>>, imgref: &str) -> Result<Self> {
         // See https://github.com/containers/skopeo/issues/2563
-        let skopeo_cmd = if imgref.starts_with("containers-storage:") {
+        // TODO: Also check here for Docker container...
+        let skopeo_cmd = if imgref.starts_with("containers-storage:") && !is_podman_container() {
             let mut cmd = Command::new("podman");
             cmd.args(["unshare", "skopeo"]);
             Some(cmd)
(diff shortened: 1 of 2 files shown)


src/oci/mod.rs
```diff
@@ -69,7 +70,8 @@ type ContentAndVerity<ObjectID> = (Sha256Digest, ObjectID);
 impl<ObjectID: FsVerityHashValue> ImageOp<ObjectID> {
     async fn new(repo: &Arc<Repository<ObjectID>>, imgref: &str) -> Result<Self> {
         // See https://github.com/containers/skopeo/issues/2563
-        let skopeo_cmd = if imgref.starts_with("containers-storage:") {
+        // TODO: Also check here for Docker container...
+        let skopeo_cmd = if imgref.starts_with("containers-storage:") && !is_podman_container() {
```
allisonkarlitskaya, line 74: I don't think this check is completely correct.  It would break my usecase for sure.

Specifically: I'm running in a toolbox, as a non-root user, and when I want to pull skopeo `containers-storage:` URLs I want to use `podman unshare` to escape the container.  It's a very common practice to have some kind of a `flatpak-spawn --host`-based script for that in toolbox-type environments...

Checking if we were root would get a bit closer to a solution here, I guess...

What happened to the idea of mounting the container's rootfs inside of itself using `--mount=` and building the image from that?

cgwalters: > Checking if we were root would get a bit closer to a solution here, I guess...

Probably, but I think what we actually need anyways is to override the proxy configuration entirely anyways - ref https://github.com/bootc-dev/bootc/issues/1292#issuecomment-2841657355 right?

> What happened to the idea of mounting the container's rootfs inside of itself using --mount= and building the image from that?

A bit confused by that; `/` is already the merged rootfs, why would we mount it again?
But the real problem is that we don't want the merged root, we want the layers so we can store them separately and we want the manifest/config.

allisonkarlitskaya: Interesting.  I didn't think about the config, but I guess that has some useful stuff in it.

Why do we want to store the layers separately?

allisonkarlitskaya: Sure, but by the time we go to install the image, we've already downloaded the entire thing, no?

cgwalters: Yep absolutely, but we want to retain distinct knowledge of the layers so that the next `bootc upgrade` doesn't re-fetch all of the content again (xref https://github.com/bootc-dev/bootc/issues/1197 ) and we only have a hope of doing that if we retain knowledge of the layer structure.

allisonkarlitskaya: True.

allisonkarlitskaya (approved):
Thanks!

PR #119, opened 2025-05-08
directory: Add _opt helper

cgwalters:
This is like https://docs.rs/cap-std-ext/latest/cap_std_ext/dirext/trait.CapStdExtDirExt.html#tymethod.open_dir_optional etc.

The rationale here is that it's really common for filesystem operations to want to handle "not found" and that maps most cleanly to `Option`. We only have one use right now but there could easily be more.

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -437,3 +437,294 @@ impl<ObjectID: FsVerityHashValue> FileSystem<ObjectID> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::fsverity::Sha256HashValue;
+    use std::cell::RefCell;
+    use std::collections::BTreeMap;
+    use std::ffi::{OsStr, OsString};
+    use std::rc::Rc;
+
+    // Helper to create a Stat with a specific mtime
+    fn stat_with_mtime(mtime: i64) -> Stat {
+        Stat {
+            st_mode: 0o755,
+            st_uid: 1000,
+            st_gid: 1000,
+            st_mtim_sec: mtime,
+            xattrs: RefCell::new(BTreeMap::new()),
+        }
+    }
+
+    // Helper to create a simple Leaf (e.g., an empty inline file)
+    fn new_leaf_file(mtime: i64) -> Rc<Leaf<Sha256HashValue>> {
+        Rc::new(Leaf {
+            stat: stat_with_mtime(mtime),
+            content: LeafContent::Regular(RegularFile::Inline(Box::new([]))),
+        })
+    }
+
+    // Helper to create a simple Leaf (symlink)
+    fn new_leaf_symlink(target: &str, mtime: i64) -> Rc<Leaf<Sha256HashValue>> {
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
Thanks for this.  I find this generally quite nice and I've considered adding this as well.

I think you could improve the implementation by either just repeating it or switching it around: we synthesize the NotFoundError when we get None returned from the btreemap, so the `_opt()` case is actually the "more natural" case of the two...

allisonkarlitskaya (changes requested):
oh, and `cargo fmt` is unhappy.

cgwalters:
> I think you could improve the implementation by either just repeating it or switching it around: we synthesize the NotFoundError when we get None returned from the btreemap, so the _opt() case is actually the "more natural" case of the two...

I tried this but it's not that easy `NotFound` includes the filename right now, so we can't actually go `None` -> `NotFoundError`.

cgwalters:
> I tried this but it's not that easy NotFound includes the filename right now, so we can't actually go None -> NotFoundError.

Eh I was wrong, of course we have the original filename in the function anyways. Updated.

src/tree.rs
```diff
@@ -164,12 +175,12 @@ impl<ObjectID: FsVerityHashValue> Directory<ObjectID> {
                 Component::Normal(filename) => match dir.entries.get(filename) {
                     Some(Inode::Directory(subdir)) => subdir,
                     Some(_) => return Err(ImageError::NotADirectory(filename.into())),
-                    None => return Err(ImageError::NotFound(filename.into())),
+                    None => return Ok(None),
```
allisonkarlitskaya, line 178: aww.  I expected to see some fancy `.map().transpose()` work here ;)

src/tree.rs
```diff
@@ -417,3 +439,294 @@ impl<ObjectID: FsVerityHashValue> FileSystem<ObjectID> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
allisonkarlitskaya, line 444: :heart:

allisonkarlitskaya (approved):
Looks great, thanks!

Feel free to merge it or have fun with `.map().transpose()`, as you wish :)

src/tree.rs
allisonkarlitskaya: er.  I wrote this comment int he wrong place.  I meant it for the get_file_opt() case, sorry.

src/tree.rs
```diff
@@ -301,13 +312,24 @@ impl<ObjectID: FsVerityHashValue> Directory<ObjectID> {
         &'a self,
         filename: &OsStr,
     ) -> Result<&'a RegularFile<ObjectID>, ImageError> {
+        match self.get_file_opt(filename)? {
+            Some(r) => Ok(r),
+            None => Err(ImageError::NotFound(Box::from(filename))),
```
allisonkarlitskaya, line 317: ...and this looks like a `?.ok_or_else(||)` :)

allisonkarlitskaya (approved):

PR #120, opened 2025-05-08
enable_verity: syncfs and retry on ETXTBSY

jeckersb:
On btrfs, sometimes after the writable fd is closed, attempting to
enable verity will still result in ETXTBSY.  In this case it appears
to be sufficient to call `syncfs()` and retry enabling verity.  This
isn't ideal since it's a lot slower, but better than hard-erroring.

Resolves: #106

### DIFF
--- a/src/fsverity/mod.rs
+++ b/src/fsverity/mod.rs
@@ -30,6 +30,8 @@ pub enum EnableVerityError {
     FilesystemNotSupported,
     #[error("fs-verity is already enabled on file")]
     AlreadyEnabled,
+    #[error("File is opened for writing")]
+    FileOpenedForWrite,
 }
 
 /// A verity comparison failed.
@@ -69,7 +71,20 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
 /// It's possible to choose the hash algorithm (via the generic parameter) but the blocksize is
 /// currently hardcoded to 4096.  Salt is not supported.
 pub fn enable_verity<H: FsVerityHashValue>(fd: impl AsFd) -> Result<(), EnableVerityError> {
-    ioctl::fs_ioc_enable_verity::<H>(fd)
+    match ioctl::fs_ioc_enable_verity::<H>(&fd) {
+        Err(EnableVerityError::FileOpenedForWrite) => {
+            // https://github.com/containers/composefs-rs/issues/106
+            // On btrfs, sometimes after the writable fd is closed,
+            // attempting to enable verity will still result in
+            // ETXTBSY.  In this case it appears to be sufficient to
+            // call `syncfs()` and retry enabling verity.  This isn't
+            // ideal since it's a lot slower, but better than
+            // hard-erroring.
+            rustix::fs::syncfs(&fd).map_err(Into::<std::io::Error>::into)?;
+            ioctl::fs_ioc_enable_verity::<H>(&fd)
+        }
+        other => other,
+    }
 }
 
 /// Measures fs-verity on the given file.
--- a/src/fsverity/ioctl.rs
+++ b/src/fsverity/ioctl.rs
@@ -52,6 +52,7 @@ pub(super) fn fs_ioc_enable_verity<H: FsVerityHashValue>(
                 Err(EnableVerityError::FilesystemNotSupported)
             }
             Err(Errno::EXIST) => Err(EnableVerityError::AlreadyEnabled),
+            Err(Errno::TXTBSY) => Err(EnableVerityError::FileOpenedForWrite),
             Err(e) => Err(Error::from(e).into()),
             Ok(_) => Ok(()),
         }


src/fsverity/mod.rs
```diff
@@ -69,7 +71,20 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+    match ioctl::fs_ioc_enable_verity::<H>(&fd) {
+        Err(EnableVerityError::FileOpenedForWrite) => {
+            // https://github.com/containers/composefs-rs/issues/106
+            // On btrfs, sometimes after the writable fd is closed,
+            // attempting to enable verity will still result in
+            // ETXTBSY.  In this case it appears to be sufficient to
+            // call `syncfs()` and retry enabling verity.  This isn't
+            // ideal since it's a lot slower, but better than
+            // hard-erroring.
+            rustix::fs::syncfs(&fd).map_err(Into::<std::io::Error>::into)?;
```
cgwalters, line 83: Wait we need to sync *the entire filesystem*, and not just `fsync`?

cgwalters (commented):
I think we should probably report this as a kernel bug (against btrfs? have we ever seen it with ext4?)

jeckersb:
> I think we should probably report this as a kernel bug (against btrfs? have we ever seen it with ext4?)

Yeah I know @allisonkarlitskaya had already mentioned this (upstream?) somewhere, I don't know if there's an official tracker for it or not.  I'll let her chime in before I go off creating additional noise.

src/fsverity/mod.rs
jeckersb: Let me double check this since I went through a lot of iterations on this, but I think `fsync` wasn't sufficient.  It might be that `syncfs()` doesn't even really do anything other than introduce enough delay for things to settle.

jeckersb: Confirmed, `fsync` isn't enough, it still fails sometimes with that.

cgwalters: To clarify the problem with `syncfs` is that it penalizes every other process on the system.

Hmm, how about `calling `fsync` *and* calling `fsync` on the parent directory of the file? That might force flush of the inode data perhaps?

cgwalters: Actually just thinking this through more...maybe the way to handle this is to make it an explicit loop like:

```
  const TRIES = 60;
  for (i in 0..TRIES+1) {
     match enable_verity() {
       Err(opened) if (i < TRIES) => { sleep(1); continue; }
       o => return o
    }
   }
  
```

or so?

jeckersb: Brain dump before I take off for the day...

I implemented the above, sleeping 1 nanosecond between invocations, and then with some extra logging added I ran the tests 100x like:

```
$ for i in `seq 100`; do cargo test -- --show-output | grep -B1 Succeeded | grep size | cut -f2 -d'='; done | sort -n | uniq -c
```

Which shows us how often each size test case failed and needed to be retried.  The reason I did this is that I noticed in the big list of various sizes we try, the failures were always with the values at the smaller end.

(This is in relationship to [size](https://github.com/containers/composefs-rs/blob/main/src/fsverity/mod.rs#L348) in the `crosscheck_interesting_cases` test)

So end result I ended up with:

```
      2 1
      1 31
      5 32
     10 33
      8 63
      5 64
      8 65
      2 991
      1 992
```

Meaning, out of 100 invocations of cargo test, data of size `33` was the most common to fail, which happened 10 out of 100 times.

The larger data sizes seem to always succeed.

Also worth noting that when we do fail, it seems to take anywhere between 2 and 5 retries before it finally succeeds.

I might come back to this later this evening and tweak things a bit and gather a larger sample size.

allisonkarlitskaya:
> Yeah I know @allisonkarlitskaya had already mentioned this (upstream?) somewhere, I don't know if there's an official tracker for it or not. I'll let her chime in before I go off creating additional noise.

I just pinged @brauner about it and he gave some hints.  There's no official upstream report, and I agree that we should definitely do that.

I also think that `syncfs()` is one hell of a heavy hammer for this.

But, if we want to do that in the testcases, I'm open to it, but it's too much for the repository, which means that I'd prefer not to bake it into the ioctl code.  I see `crosscheck_interesting_cases()` fail locally a hell of a lot (>50%) and I've started to ignore it when it does, which is always a bit dangerous.  I'd welcome a timely fix to that one testcase, even if it's obscene.

So:
 - patch the testcase now
 - ask upstream for clarification
 - wait on any change that impacts repository, and (probably) make the change to the `Repository` code itself.  We already call `fdatasync()` there, for example, so a potential `sync_harder()` call would fit in nicely.  I have a feeling that the issue is highly useage-pattern dependent and it feels weird to put a workaround in the ioctl code....

src/fsverity/mod.rs
allisonkarlitskaya, line 83: This is definitely the wrong spot for this.

allisonkarlitskaya (changes requested):

src/fsverity/mod.rs
cgwalters: > Also worth noting that when we do fail, it seems to take anywhere between 2 and 5 retries before it finally succeeds.

With just one nanosecond between invocations? That's not much at all! Surely the syscall overhead here is really dwarfing that sleep. I'd personally sleep at least a millisecond here just because going faster than that is nearly a busy loop.

jeckersb:
I'm going to close this and resume discussion in #106 as I continue trying to figure this out.

PR #121, opened 2025-05-09
Allow passing absolute paths to initrd and vmlinuz

Johan-Liebert1:
Grub needs absolute paths to initrd and vmlinuz if we do not have `/boot` in a boot partition, which we do not in bootc.

### DIFF
--- a/crates/composefs-boot/src/bootloader.rs
+++ b/crates/composefs-boot/src/bootloader.rs
@@ -144,7 +144,7 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
-    pub fn relocate(&mut self, entry_id: &str) {
+    pub fn relocate(&mut self, boot_subdir: Option<&str>, entry_id: &str) {
         self.filename = Box::from(format!("{entry_id}.conf").as_ref());
         for line in &mut self.entry.lines {
             for key in ["linux", "initrd", "efi"] {
@@ -159,7 +159,14 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
 
                 let new = format!("/{entry_id}/{basename}");
                 let range = substr_range(line, value).unwrap();
-                line.replace_range(range, &new);
+
+                let final_entry_path = if let Some(boot_subdir) = boot_subdir {
+                    format!("/{boot_subdir}{new}")
+                } else {
+                    new.clone()
+                };
+
+                line.replace_range(range, &final_entry_path);
 
                 if let Some(file) = file {
                     self.files.insert(new.into_boxed_str(), file);
--- a/crates/composefs-boot/src/write_boot.rs
+++ b/crates/composefs-boot/src/write_boot.rs
@@ -16,10 +16,18 @@ use crate::{
 pub fn write_t1_simple<ObjectID: FsVerityHashValue>(
     mut t1: Type1Entry<ObjectID>,
     bootdir: &Path,
+    boot_subdir: Option<&str>,
     root_id: &ObjectID,
     cmdline_extra: &[&str],
     repo: &Repository<ObjectID>,
 ) -> Result<()> {
+    let bootdir = if let Some(subdir) = boot_subdir {
+        let subdir_path = Path::new(subdir);
+        bootdir.join(subdir_path.strip_prefix("/").unwrap_or(subdir_path))
+    } else {
+        bootdir.to_path_buf()
+    };
+
     t1.entry
         .adjust_cmdline(Some(&root_id.to_hex()), cmdline_extra);
 
@@ -63,35 +71,78 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
     Ok(())
 }
 
+/// Writes boot entry to the boot partition
+///
+/// # Arguments
+///
+/// * repo           - The composefs repository
+/// * entry          - Boot entry variant to be written
+/// * root_id        - The content hash of the generated EROFS image id
+/// * boot_partition - Path to the boot partition/directory
+/// * boot_subdir    - If `Some(path)`, the path is prepended to `initrd` and `linux` keys in the BLS entry
+///
+/// For example, if `boot_partition = "/boot"` and `boot_subdir = Some("1")` ,
+/// the BLS entry will contain
+///
+/// ```text
+/// linux /boot/1/<entry_id>/linux
+/// initrd /boot/1/<entry_id>/initrd
+/// ```
+///
+/// If `boot_partition = "/boot"` and `boot_subdir = None` , the BLS entry will contain
+///
+/// ```text
+/// linux /<entry_id>/linux
+/// initrd /<entry_id>/initrd
+/// ```
+///
+/// * entry_id       - In case of a BLS entry, the name of file to be generated in `loader/entries`
+/// * cmdline_extra  - Extra kernel command line arguments
+///
 pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
     repo: &Repository<ObjectID>,
     entry: BootEntry<ObjectID>,
     root_id: &ObjectID,
-    bootdir: &Path,
+    boot_partition: &Path,
+    boot_subdir: Option<&str>,
     entry_id: Option<&str>,
     cmdline_extra: &[&str],
 ) -> Result<()> {
     match entry {
         BootEntry::Type1(mut t1) => {
             if let Some(name) = entry_id {
-                t1.relocate(name);
+                t1.relocate(boot_subdir, name);
             }
-            write_t1_simple(t1, bootdir, root_id, cmdline_extra, repo)?;
+            write_t1_simple(
+                t1,
+                boot_partition,
+                boot_subdir,
+                root_id,
+                cmdline_extra,
+                repo,
+            )?;
         }
         BootEntry::Type2(mut t2) => {
             if let Some(name) = entry_id {
                 t2.rename(name);
             }
             ensure!(cmdline_extra.is_empty(), "Can't add --cmdline args to UKIs");
-            write_t2_simple(t2, bootdir, root_id, repo)?;
+            write_t2_simple(t2, boot_partition, root_id, repo)?;
         }
         BootEntry::UsrLibModulesUki(_entry) => todo!(),
         BootEntry::UsrLibModulesVmLinuz(entry) => {
             let mut t1 = entry.into_type1(entry_id);
             if let Some(name) = entry_id {
-                t1.relocate(name)?;
+                t1.relocate(boot_subdir, name);
             }
-            write_t1_simple(t1, bootdir, root_id, cmdline_extra, repo)?;
+            write_t1_simple(
+                t1,
+                boot_partition,
+                boot_subdir,
+                root_id,
+                cmdline_extra,
+                repo,
+            )?;
         }
     };
 
--- a/crates/composefs-boot/src/write_boot.rs
+++ b/crates/composefs-boot/src/write_boot.rs
@@ -87,7 +87,10 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
         }
         BootEntry::UsrLibModulesUki(_entry) => todo!(),
         BootEntry::UsrLibModulesVmLinuz(entry) => {
-            let t1 = entry.into_type1(entry_id);
+            let mut t1 = entry.into_type1(entry_id);
+            if let Some(name) = entry_id {
+                t1.relocate(name)?;
+            }
             write_t1_simple(t1, bootdir, root_id, cmdline_extra, repo)?;
         }
     };
--- a/crates/cfsctl/src/main.rs
+++ b/crates/cfsctl/src/main.rs
@@ -279,6 +279,7 @@ async fn main() -> Result<()> {
                     entry,
                     &id,
                     bootdir,
+                    None,
                     entry_id.as_deref(),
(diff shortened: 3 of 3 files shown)


src/bootloader.rs
```diff
@@ -143,7 +144,7 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
-    pub fn relocate(&mut self, entry_id: &str) {
+    pub fn relocate(&mut self, entry_id: &str, abs_entry_path: Option<&str>) -> Result<()> {
```
cgwalters, line 147: I think it'd be more consistent if this was `Option<&Path>`

src/bootloader.rs
```diff
@@ -157,14 +158,29 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
                 let file = self.files.remove(value);
 
                 let new = format!("/{entry_id}/{basename}");
+                let new_entry_path = Path::new(&new);
```
cgwalters, line 161: Except yeah, above we have a string and so...this is where camino makes sense. It's honestly also one of the reasons I just ended up strictly denying non-UTF8 filenames in ostree, dealing with it is just a giant pain and there's so many file formats which reference filenames that just don't support non-UTF8 filenames, or would be extremely awkward to do.

cgwalters (approved):
Per previous debate overall I just really do not like this idea of mutating bootloader entries shipped in the image and would prefer that we generate them client side. This logic would need to fundamentally change for that (and look more similar to what `kernel-install` and `ostree` to today).

But from my PoV that isn't blocking experimentation.

src/bootloader.rs
Johan-Liebert1: Yeah, `std::Path` would've been a better choice here. Unsure why we went with `&str`, maybe because of Rust shenanigans? @allisonkarlitskaya

allisonkarlitskaya (commented):
The first patch should absolutely just go in as-is.  That's clearly an oversight on my part.  We might even add the relocation step to the write_type1 function to save a bit of repetition (which is what ultimately led to the bug).

For the second one, I think the issue here is that it's not clear what "bootdir" refers to.  Up to now (in my brain, at least) bootdir was one thing: the root of the boot partition.  Now we have the possibility of not having a separate boot partition but rather /boot/ inside of another partition.  I think this is a violation of the bootloader spec, but let's set that aside.

How about we rather do something like having two parameters:
 - an absolute path to the partition which contains the boot stuff
 - an optional relative path to the boot directory within that partition

So for a bootc system this would look like `"/sysroot"` and `Some("boot")` respectively.  For the examples it would look like `"/boot"` and `None`.

That would allow us to:
 - properly locate our `loader/entries/` directory
 - properly place our boot resources inside of a subdir
 - properly refer to those boot resources in our loader entry file relative to the root of the partition
 - not have any redundancy in the path components we provide

allisonkarlitskaya:
>  I think this is a violation of the bootloader spec

That sort of thing requires a citation, so here we go:

> Note: In all cases the /loader/entries/ directory should be located directly in the root of the ESP or XBOOTLDR filesystem.

https://uapi-group.org/specifications/specs/boot_loader_specification/

cgwalters:
> I think this is a violation of the bootloader spec

Yes, the status quo with at least Fedora derivatives using grub and ostree (at least) is that it isn't set up matching the BLS. That's 100% something we'd like to fix (https://github.com/ostreedev/ostree/pull/3359 is one notable ingredient).

src/write_boot.rs
```diff
@@ -63,6 +63,34 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
...
+/// # Arguments
+///
+/// * repo           - The composefs repository
+/// * entry          - Boot entry variant to be written
+/// * root_id        - The content hash of the generated EROFS image id
+/// * boot_partition - Path to the boot partition/directory
+/// * entry_id       - In case of a BLS entry, the name of file to be generated in `loader/entries`
+/// * boot_subdir    - If `Some(path)`, the path is prepended to `initrd` and `linux` keys in the BLS entry
+///
+/// For example, if `boot_subdir = Some("/boot/1")` and `boot_partition = "/boot"`,
```
allisonkarlitskaya, line 77: I think this is what I was trying to avoid.

I rather think that in this case we'd have:
 - `boot_partition`: `/boot`
 - `boot_subdir`: `Some("1")`

So in the BLS-compliant case where the boot directory is rooted in its own filesystem, we'd pass `None` for `boot_subdir`.  We should think about this less in terms of mechanics like ("prepended to paths in the entry file") and more about what the user sees in terms of their config.

allisonkarlitskaya (changes requested):
I'm very sorry.  I thought I posted this review earlier today.  It's the same as we discussed on matrix, though...

src/write_boot.rs
Johan-Liebert1: But again, we want absolute "boot" path. If the "boot" dir is mounted at `/run/bootc/mounts/rootfs/boot` we have no way of knowing what the "final" absolute path is going to be. It could be `/rootfs/boot` or it could just be `/boot`, and this is the crux of the issue. 

I didn't like the idea of just assuming that the "final" directory in the `boot_partition` will be the actual one.

src/write_boot.rs
```diff
@@ -67,31 +67,32 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
     repo: &Repository<ObjectID>,
     entry: BootEntry<ObjectID>,
     root_id: &ObjectID,
-    bootdir: &Path,
+    boot_partition: &Path,
```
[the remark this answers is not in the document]
Johan-Liebert1: Updated

src/bootloader.rs
```diff
@@ -143,7 +144,7 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
-    pub fn relocate(&mut self, entry_id: &str) {
+    pub fn relocate(&mut self, entry_id: &str, boot_subdir: Option<&str>) -> Result<()> {
```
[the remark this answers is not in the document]
Johan-Liebert1: Updated

crates/composefs-boot/src/write_boot.rs
```diff
@@ -71,7 +71,6 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
 /// * entry          - Boot entry variant to be written
 /// * root_id        - The content hash of the generated EROFS image id
 /// * boot_partition - Path to the boot partition/directory
-/// * entry_id       - In case of a BLS entry, the name of file to be generated in `loader/entries`
 /// * boot_subdir    - If `Some(path)`, the path is prepended to `initrd` and `linux` keys in the BLS entry
 ///
 /// For example, if `boot_subdir = Some("/boot/1")` and `boot_partition = "/boot"`,
```
allisonkarlitskaya, line 76: The repetition of `/boot` is the thing that I'm objecting to.  The path to the files contained in the entry should consist of concatenating boot_partition + boot_subdir + entry_id.  I think we should talk about this in a call :)

crates/composefs-boot/src/bootloader.rs
```diff
@@ -144,7 +145,7 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
-    pub fn relocate(&mut self, entry_id: &str) {
+    pub fn relocate(&mut self, entry_id: &str, boot_subdir: Option<&str>) -> Result<()> {
```
allisonkarlitskaya: Same nags as before: try to keep these in the order that they get pasted back together... so when they appear, they are always in the order of: boot_partition, boot_subdir, entry_id.

crates/composefs-boot/src/bootloader.rs
```diff
@@ -158,14 +159,29 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
...
+                        .join(new_entry_path.strip_prefix("/").unwrap_or(new_entry_path))
+                } else {
+                    new_entry_path.to_path_buf()
+                };
+
+                let final_path = final_path
+                    .to_str()
+                    .ok_or_else(|| anyhow!("Failed to convert Path '{final_path:?}' to string"))?;
+
+                line.replace_range(range, final_path);
```
allisonkarlitskaya, line 176: This seems really too complicated.  Should just be something like:

```rust
let new = match boot_subdir {
    Some(subdir) => format!("/{subdir}/{entry_id}/{basename}"),
    None => format!("/{entry_id}/{basename}"),
};
```

or so, no?

crates/composefs-boot/src/bootloader.rs
allisonkarlitskaya: Also: with the arguments defined as they are now, I don't think this function needs to be failable anymore...

crates/composefs-boot/src/write_boot.rs
```diff
@@ -67,31 +67,32 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
     repo: &Repository<ObjectID>,
     entry: BootEntry<ObjectID>,
     root_id: &ObjectID,
-    bootdir: &Path,
+    boot_partition: &Path,
     entry_id: Option<&str>,
+    boot_subdir: Option<&str>,
```
allisonkarlitskaya, line 99: Order again: boot_partition, boot_subdir, entry_id.

crates/composefs-boot/src/write_boot.rs
```diff
@@ -67,31 +67,32 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
...
         }
         BootEntry::UsrLibModulesUki(_entry) => todo!(),
         BootEntry::UsrLibModulesVmLinuz(entry) => {
             let mut t1 = entry.into_type1(entry_id);
             if let Some(name) = entry_id {
-                t1.relocate(name)?;
+                t1.relocate(name, boot_subdir)?;
             }
-            write_t1_simple(t1, bootdir, root_id, cmdline_extra, repo)?;
+            write_t1_simple(t1, boot_partition, root_id, cmdline_extra, repo)?;
```
allisonkarlitskaya, line 123: Feels like that's gonna fail?  boot_partition.join(boot_subdir), no?  Or just pass both arguments through.

You might actually consider doing the relocation inside of the write code as well (it would reduce a small amount of duplication).

allisonkarlitskaya (commented):
Looking better thanks, but I think with the args defined the way that they are defined now, we don't need to make things so complicated...

crates/composefs-boot/src/bootloader.rs
Johan-Liebert1: So, `new` here is where the loader file is created relative to the /boot directory and final is the final concatenated path in the loader/entries/<entry_id>

They have to be separate

crates/composefs-boot/src/write_boot.rs
Johan-Liebert1: This is correct, no?

crates/composefs-boot/src/bootloader.rs
allisonkarlitskaya: ah sure.  I failed to notice that we use `new` for both things.  But even still, you could `let new_entry_path = match ...` as per the previous comment, no?

crates/composefs-boot/src/bootloader.rs
```diff
@@ -144,7 +144,7 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
-    pub fn relocate(&mut self, entry_id: &str) {
+    pub fn relocate(&mut self, boot_subdir: Option<&str>, entry_id: &str) -> Result<()> {
```
allisonkarlitskaya, line 147: Drop the Result<()> please

crates/composefs-boot/src/write_boot.rs
allisonkarlitskaya: Yes. I think I wrote that to a previous iteration.

crates/composefs-boot/src/bootloader.rs
```diff
@@ -159,13 +159,22 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
 
                 let new = format!("/{entry_id}/{basename}");
                 let range = substr_range(line, value).unwrap();
-                line.replace_range(range, &new);
+
+                let final_entry_path = if let Some(boot_subdir) = boot_subdir {
+                    format!("/{boot_subdir}{new}")
+                } else {
+                    new.clone()
```
allisonkarlitskaya, line 166: I guess you could just move (and duplicate) the `line.replace_range()` into the if to avoid the clone if it bothers you ;)

crates/composefs-boot/src/bootloader.rs
Johan-Liebert1: my bad, forgot to remove after refactor

crates/composefs-boot/src/bootloader.rs
Johan-Liebert1: duplication bothers me a bit more than a String clone :)

allisonkarlitskaya (approved):
Thanks for all the iterations.  Looks good now!

PR #122, opened 2025-05-12
Update `composefs-setup-root` to parse composefs-rs config

Johan-Liebert1:
Instead of parsing ostree's toml config, we will have and parse composefs-rs config located at `/usr/lib/composefs/setup-root-conf.toml`


allisonkarlitskaya (approved):
looks reasonable, thanks!

PR #133, opened 2025-05-20
src: experimental http client support

allisonkarlitskaya:
It turns out that the information contained in splitstreams to assist with garbage collection (ie: the list of things that we mustn't discard) is exactly the required information for downloading (ie: the list of things that we must acquire).

Use this fact to add support for fetching repository content from HTTP servers.  We only download the objects that are actually required, so incremental pulls are very fast.

This works with just about any HTTP server, so you can do something like

  python -m http.server -d ~/.var/lib/composefs

and download from that.  With a fast enough web server on localhost, pulling a complete image into an empty repository takes about as long as pulling an `oci:` directory via skopeo with `cfsctl oci pull`.

In practice, this is intended to be used with a webserver which supports static compression and pre-compressed objects stored on the server.  In particular, zstd support is enabled in the `reqwest` crate for this reason, and it's working with something like:

  find repo/objects/ -type f -name '*[0-9a-f]' -exec zstd -19 -v '{}' +
  static-web-server -p 8888 --compression-static -d repo

There's also an included s3-uploader.py in the examples/ directory which will upload a repository to an S3 bucket, with zstd compression.


jeckersb:
Super cool, I'll try to set aside some time to play with this soon :)

PR #137, opened 2025-06-03
wip: zstd:chunked support via oci-client

allisonkarlitskaya:


cgwalters:
There's some cool code here, and I think it's really useful actually to have this as an option even if it's hidden/experimental just to use as a comparison.

The main issue is really c/image (the important part of skopeo here) vs oci-client as expanded on in https://github.com/containers/containers-image-proxy-rs/?tab=readme-ov-file#why and it's for that reason I can't imagine using this by default in bootc at least anytime in the nearly visible future. Things like [ECR credential helpers](https://github.com/bootc-dev/bootc/pull/215#issuecomment-2935001160) are highly load bearing. OTOH, other folks could reasonably make other choices.

allisonkarlitskaya:
Ya.  My first instinct is to stick this behind a feature flag (and maybe also do the same with the skopeo and potential-future-oci-dir paths)... it could be useful for the kind of scenario where we don't need the skopeo features.

In terms of those features, I feel like:
 - depending on the registry alias stuff for use-cases like bootc is probably not a reasonable best practice anyway
 - the auth stuff is hard, but also not required by a large number of our existing users
 - signing is a whole other conversation: if we want to do things like validation of the images we mount based on container signatures we're going to need support for this in the higher layers anyway so we can check it at runtime.

I feel like we could maybe close some of these gaps by adding some code to skopeo to do the work of resolving registry aliases, taking care of the initial authentication, etc. and call out to that.

I'm also sort of of the (optimistic) opinion that maybe we can have a Rust-based stack for all of this stuff at some point, and honestly, the sooner the better... and although this piece of code here leaves a lot of the other questions unanswered, it's maybe one part of moving towards a world where we could image a feasible solution to that puzzle...

allisonkarlitskaya:
> The first commit here though seems sane to me, let's split it into its own PR and get it in?

→ https://github.com/containers/composefs-rs/pull/138

allisonkarlitskaya:
So the biggest unanswered question here is what to do about the sha256-to-fsverity mapping.  It's currently stored as raw `sha256:abc123 ->` symlinks directly in the objects directory, but that's fairly clearly not intended to stay.

My initial impulse was to store symlinks in the repository, but: this is just about the worst possible thing we can do.  There's no actual benefit to being able to follow sha256 symlinks into the object files and this is only used for zstd:chunked.  I'd actually prefer something more like an sqlite database storing sha256->fs-verity mappings.  Also: it might make sense to store that in `/var/cache` or `~/.cache` instead of directly in the repository.

@alexlarsson is currently trying to solve a very similar issue for the ostree mapping...

PR #142, opened 2025-06-13
erofs/writer: Print trace logs instead of debug logs

Johan-Liebert1:
Turning on debug logs clutters the output. IMO these should be trace logs

### DIFF
--- a/crates/composefs/src/erofs/writer.rs
+++ b/crates/composefs/src/erofs/writer.rs
@@ -6,7 +6,7 @@ use std::{
     rc::Rc,
 };
 
-use log::debug;
+use log::trace;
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
> I think the first three things could probably stay at debug level since we get at most one print per EROFS?

The write function is called for each inode iirc

Ah, rawhide build failed. Looking into it

Johan-Liebert1:
Seems like an issue with mount code in 6.16

```
move_mount(4, "", 7, "sysroot", MOVE_MOUNT_F_EMPTY_PATH) = -1 EINVAL (Invalid argument)
```

Issue #141, opened 2025-06-05
Proposal for ostree pull support (and object id maps)


alexlarsson:
Background
==========

Ostree updates are of two forms, deltas or generic:

For a delta update, you go from a specific commit to a new one. When downloading such a delta you address it by the source commit, so it is assumed by the delta that all the objects in the source commit are already available, and only (but *all*) the remaining objects are in the delta. Additionally, if some objects are new and/or large they can be stored as "fallback" objects which are downloaded by the generic mechanism. (In practice, fallbacks are only used for file objects.)

Deltas can also be "from nothing", which means we don't have a source, so the delta will contain everything (except the fallbacks).

For "generic" updates, we only have the commit id and we download each object separately CAS-style, recursing into other referenced objects. Normally we have existing objects stored locally, so we can stop recursing whenever we reach an object we already have locally available (modulo some .partialcommit bullshit that we need not care about here). This is in theory efficient, since we can stop early, but in practice it is very chatty over HTTP with many small connections, so it is not great.

Proposal for composefs-rs
=========================

For composefs-rs, the goal is to "support" everything, but focus on the delta case. I.e. the generic approach need to work, but doesn't have to perform optimally.

At a high level, the ideal effect of a "composefs-rs pull ostree" operation would look something like this:
 * Each file object is stored as a separate object, stored by fs-verity
 * One splitstream that has all the metadata for the commit (commit, dirmeta,
   dirtree, file metadata objects) and references to the content object for each file
   metadata.

From this we can efficiently generate an erofs/composefs image. Additionally, given such a splitstream, we have all that we need to implement a delta pull. We can parse the original splitstream to reconstruct the ostree objects from the source commit, and then reconstruct the remainder from the delta.

In this above operation we never download any objects that are in the previous version of the image, however for new files we still always download them, whereas a regular ostree pull operation would notice the cross-image sharing even at pull time and avoid downloads. This is not a critical, as everything still works and ends up with the same on-disk result, but is not ideal, so we should have some kind of way to optimize this that doesn't need to be 100% perfect but should be simple and robust.

To handle this, I propose adding objects to the composefs repo that are "object id maps". These would be used for file objects only (remember, those are the only fallback objects in deltas), and contain a mapping from the ostree file object checksum to the fs-verity digest of the content and some extra data, containing the file metadata (uid, gid, mode, xattrs). Such a file format should be efficient to mmap and perform object id lookups in. I have a more detailed proposal below.

[...]

Issue #123, opened 2025-05-12
todo: upstream our ioctls to rustix, forbid unsafe


allisonkarlitskaya:
We have a small number of custom ioctl implementations:
 - `FS_IOC_ENABLE_VERITY`
 - `FS_IOC_MEASURE_VERITY`
 - `LOOP_CTL_GET_FREE`
 - `LOOP_CONFIGURE`

these are all of the unsafe code in all of composefs-rs.

The reason we implement those ourselves is pretty simple: rustix doesn't contain support.  We should look into seeing if upstream would accept this so that we can forbid unsafe code in composefs-rs.

PR #144, opened 2025-06-16
Add composefs-ostree and some basic CLI tools     

alexlarsson:
Based on ideas from https://github.com/containers/composefs-rs/issues/141
    
This is an initial version of ostree support. This allows pulling
from local and remote ostree repos, which will create a set of
regular file content objects, as well as a blob containing all the
remaining ostree objects. From the blob we can create an image.
   
When pulling a commit, a base blob (i.e. "the previous version" can be
specified. Any objects in that base blob will not be downloaded. If a
name is given for the pulled commit, then pre-existing blobs with the
same name will automatically be used as a base blob.
    
This is an initial version and there are several things missing:
 * Pull operations are completely serial
 * There is no support for ostree summary files
 * There is no support for ostree delta files
 * There is no caching of local file availability (other than base blob)
 * Local ostree repos only support archive mode

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -42,7 +42,7 @@ use std::sync::Arc;
 
 use anyhow::{Context as _, Result};
 use clap::{Parser, Subcommand, ValueEnum};
-#[cfg(feature = "oci")]
+#[cfg(any(feature = "oci", feature = "ostree"))]
 use comfy_table::{Table, presets::UTF8_FULL};
 #[cfg(any(feature = "oci", feature = "http"))]
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
@@ -501,6 +501,76 @@ enum OciCommand {
     },
 }
 
+#[cfg(feature = "ostree")]
+#[derive(Debug, Subcommand)]
+enum OstreeCommand {
+    PullLocal {
+        ostree_repo_path: PathBuf,
+        /// Ostree ref name or commit ID (64-character hex)
+        ostree_ref: String,
+        #[clap(long)]
+        base_name: Option<String>,
+    },
+    Pull {
+        ostree_repo_url: String,
+        /// Ostree ref name or commit ID (64-character hex)
+        ostree_ref: String,
+        #[clap(long)]
+        base_name: Option<String>,
+    },
+    /// Mount an ostree commit's composefs EROFS at the given mountpoint
+    Mount {
+        /// Ostree commit ref or commit ID
+        commit: String,
+        /// Target mountpoint
+        mountpoint: String,
+        /// Writable upper layer directory for overlayfs
+        #[arg(long, requires = "workdir")]
+        upperdir: Option<PathBuf>,
+        /// Work directory for overlayfs (required with --upperdir)
+        #[arg(long, requires = "upperdir")]
+        workdir: Option<PathBuf>,
+        /// Mount read-write (requires --upperdir)
+        #[arg(long, requires = "upperdir")]
+        read_write: bool,
+    },
+    /// Dump the filesystem of an ostree commit as a composefs dumpfile to stdout
+    Dump {
+        /// Ostree commit ref name
+        commit_name: String,
+    },
+    /// Compute the composefs image ID of an ostree commit
+    ComputeId {
+        /// Ostree commit ref name
+        commit_name: String,
+    },
+    /// Show the contents of an ostree commit
+    Inspect {
+        /// Ostree ref name, commit ID, or commit ID prefix
+        source: String,
+        /// Print only the commit metadata key-value pairs
+        #[clap(long)]
+        metadata: bool,
+    },
+    /// Tag an ostree commit with a name
+    ///
+    /// The source can be an ostree commit checksum or an existing ref name.
+    Tag {
+        /// Ostree commit checksum (hex) or existing ref name
+        source: String,
+        /// Tag name to assign
+        name: String,
+    },
+    /// Remove a named ostree reference
+    Untag {
+        /// Tag name to remove
+        name: String,
+    },
+    /// List all ostree commits in the repository
+    #[clap(name = "images")]
+    ListCommits,
+}
+
 /// Common options for reading a filesystem from a path
 #[derive(Debug, Parser)]
 struct FsReadOptions {
@@ -569,6 +639,11 @@ enum Command {
     Oci {
         #[clap(subcommand)]
         cmd: OciCommand,
+    },
+    #[cfg(feature = "ostree")]
+    Ostree {
+        #[clap(subcommand)]
+        cmd: OstreeCommand,
     },
     /// Mounts a composefs image, possibly enforcing fsverity of the image
     Mount {
@@ -1564,6 +1639,111 @@ where
             }
             OciCommand::Varlink { .. } => {
                 unreachable!("oci varlink is handled before opening a repository");
+            }
+        },
+        #[cfg(feature = "ostree")]
+        Command::Ostree { cmd: ostree_cmd } => match ostree_cmd {
+            OstreeCommand::PullLocal {
+                ref ostree_repo_path,
+                ref ostree_ref,
+                base_name,
+            } => {
+                eprintln!("Fetching {ostree_ref}");
+                let (verity, stats) = composefs_ostree::pull_local(
+                    &repo,
+                    ostree_repo_path,
+                    ostree_ref,
+                    base_name.as_deref(),
+                )
+                .await?;
+
+                let image_id = composefs_ostree::get_image_ref(&repo, &stats.commit_id)?;
+                println!("commit  {}", stats.commit_id);
+                println!("verity  {}", verity.to_hex());
+                println!("image   {}", image_id.to_hex());
+                if !composefs_ostree::is_commit_id(ostree_ref) {
+                    println!("tagged  {ostree_ref}");
+                }
+                println!(
+                    "objects {} metadata + {} files fetched",
+                    stats.metadata_fetched, stats.files_fetched
+                );
+            }
+            OstreeCommand::Pull {
+                ref ostree_repo_url,
+                ref ostree_ref,
+                base_name,
+            } => {
+                eprintln!("Fetching {ostree_ref}");
+                let (verity, stats) = composefs_ostree::pull(
+                    &repo,
+                    ostree_repo_url,
+                    ostree_ref,
+                    base_name.as_deref(),
+                )
+                .await?;
+
+                let image_id = composefs_ostree::get_image_ref(&repo, &stats.commit_id)?;
+                println!("commit  {}", stats.commit_id);
+                println!("verity  {}", verity.to_hex());
+                println!("image   {}", image_id.to_hex());
+                if !composefs_ostree::is_commit_id(ostree_ref) {
+                    println!("tagged  {ostree_ref}");
+                }
+                println!(
+                    "objects {} metadata + {} files fetched",
+                    stats.metadata_fetched, stats.files_fetched
+                );
+            }
+            OstreeCommand::Mount {
+                ref commit,
+                ref mountpoint,
+                ref upperdir,
+                ref workdir,
+                read_write,
+            } => {
+                let mount_options =
+                    get_mount_options(upperdir.as_deref(), workdir.as_deref(), read_write)?;
+                let image_id = composefs_ostree::get_image_ref(&repo, commit)?;
+                repo.mount_at(&image_id.to_hex(), mountpoint.as_str(), &mount_options)?;
+            }
+            OstreeCommand::Dump { ref commit_name } => {
+                let fs = composefs_ostree::create_filesystem(&repo, commit_name)?;
+                fs.print_dumpfile()?;
+            }
+            OstreeCommand::ComputeId { ref commit_name } => {
+                let image_id = composefs_ostree::ensure_ostree_erofs(&repo, commit_name)?;
+                println!("{}", image_id.to_hex());
+            }
+            OstreeCommand::Inspect {
+                ref source,
+                metadata,
+            } => {
+                composefs_ostree::inspect(&repo, source, metadata)?;
+            }
+            OstreeCommand::Tag {
+                ref source,
+                ref name,
+            } => {
+                composefs_ostree::tag(&repo, source, name)?;
+                println!("Tagged {source} as {name}");
+            }
+            OstreeCommand::Untag { ref name } => {
+                composefs_ostree::untag(&repo, name)?;
+            }
+            OstreeCommand::ListCommits => {
+                let commits = composefs_ostree::list_commits(&repo)?;
+                if commits.is_empty() {
+                    println!("No ostree commits found");
+                } else {
+                    let mut table = Table::new();
+                    table.load_preset(UTF8_FULL);
+                    table.set_header(["NAME", "COMMIT"]);
+                    for c in commits {
+                        table.add_row([c.name.as_str(), &c.commit_id]);
+                    }
+                    println!("{table}");
+                }
             }
         },
         Command::CreateImage {
--- a/crates/composefs-ostree/src/commit.rs
+++ b/crates/composefs-ostree/src/commit.rs
@@ -1,0 +1,588 @@
+//! Ostree commit splitstream serialization and deserialization.
+//!
+//! Implements the binary format described in `doc/ostree.md`: a sorted array
+//! of ostree object IDs with bucket-indexed lookup, per-object data refs
+//! (with optional external content references), and an 8-byte-aligned content
+//! region.
+use anyhow::{Result, anyhow, bail, ensure};
+use gvariant::aligned_bytes::{A8, AlignedBuf, AlignedSlice, TryAsAligned};
+use std::{fmt, io::Read, mem::size_of, sync::Arc};
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+use sha2::{Digest, Sha256};
+use std::{
+    collections::{BTreeMap, HashMap},
+    ffi::OsStr,
+    os::unix::ffi::OsStrExt,
+};
+
+use composefs::{
+    fsverity::FsVerityHashValue,
+    generic_tree::LeafId,
+    repository::Repository,
+    tree::{Directory, FileSystem, Inode, LeafContent, RegularFile, Stat},
+    util::Sha256Digest,
+};
+
+use crate::ostree::{
+    OstreeCommit, OstreeDirMeta, OstreeDirTree, OstreeFileHeader, split_sized_variant,
+};
+
+const OSTREE_COMMIT_CONTENT_TYPE: u64 = 0xAFE138C18C463EF1;
+
+const S_IFMT: u32 = 0o170000;
+const S_IFLNK: u32 = 0o120000;
+
+fn xattrs_to_btreemap(xattrs: &[(Vec<u8>, Vec<u8>)]) -> BTreeMap<Box<OsStr>, Box<[u8]>> {
+    xattrs
+        .iter()
+        .map(|(k, v)| (OsStr::from_bytes(k).into(), Box::from(v.as_slice())))
+        .collect()
+}
+
+#[derive(Debug, FromBytes, Immutable, IntoBytes, KnownLayout)]
+#[repr(C)]
+struct CommitHeader {
+    commit_id: u32,
+    flags: u32,
+    bucket_ends: [u32; 256],
+}
+
+#[derive(Debug, FromBytes, Immutable, KnownLayout)]
+#[repr(C)]
+struct Sha256DigestArray {
+    ids: [Sha256Digest],
+}
+
+const NO_EXTERNAL_INDEX: u32 = u32::MAX;
+
+#[derive(Debug, FromBytes, Immutable, IntoBytes, KnownLayout, Clone)]
+#[repr(C)]
+struct DataRef {
+    offset: u32,
+    size: u32,
+    external_index: u32,
+}
+
+impl DataRef {
+    pub fn new(offset: usize, size: usize, external_index: Option<usize>) -> Self {
+        DataRef {
+            offset: u32::to_le(offset as u32),
+            size: u32::to_le(size as u32),
+            external_index: u32::to_le(match external_index {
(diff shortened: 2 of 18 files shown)


allisonkarlitskaya (dismissed):
I love this!  Thanks for working on it!

I made some comments on the first round of commits.  Feel free to adjust those and PR them separately: we can merge those now without further discussion.

The blobs thing is going to need a call.

I didn't review the crate addition in any detail at all.  That's probably also going to need a call :)

alexlarsson:
Ok. Reworked this to use splitstreams for object maps and commits. And, by using an object mapping to find the object map we make the content of the splitstream for the commit be *just* the commit data, and thus the sha256 of that splitstream matches the ostree commit id.

alexlarsson:
@allisonkarlitskaya There is still lots to do here. But have a look at this approach and see what you think.

alexlarsson:
I started working on the delta support, but it failed because of an [issue in gvariant-rs](https://github.com/ostreedev/gvariant-rs/issues/10).

allisonkarlitskaya (commented):
It occurs to me that it might be interesting not to sort the table of fs-verity references, and it might also be interesting to permit duplicate items.

On the topic of deferring writing of objects to a background thread, this would allow us to write "external object #123" based on a sequential index to the splitstream without actually knowing the hash value yet, and then fill in the actual values in the header at the end when we're writing: it helps there that the fs-verity references aren't compressed and therefore not part of the stream...

cgwalters:
Can you explain a bit what the medium-term is for this? Is it flatpak using this? Would we try to create a shim C API? Or is it for something like bootc (or rpm-ostree?) fetching from ostree repos? Something else?

---

Something else I'm thinking about here is the intersection with https://github.com/containers/composefs-rs/pull/216 - this is a half-baked thought but...maybe we could always map ostree to OCI (instead of mapping ostree to splitstream), though depending on how we do that there could be nontrivial metadata overhead.

allisonkarlitskaya:
I would actually like to see this merged.  It's good code and will almost definitely be useful to somebody at some point.  It also doesn't really hurt anyone since it's in a separate crate.

cgwalters:
> I also think the fact that it has nothing to do with OCI is great. 

I think unless we prove out that composefs can be a very good way to store OCI, then it is not worth investing in. Thankfully that's not the case - I think it is (and I believe you do too!).

So it's not that it has "nothing to do with OCI" (right?) - how about "has the capability to easily/natively store any type of content that one would want to represent as read-only immutable versioned filesystem trees".

For example, today Android as far as I know uses fsverity on single zip files, and they've made it work quite well, but it's harder to get deduplication across apps that way, and maybe someday they go to a composefs-like model.

crates/composefs-ostree/src/repo.rs
```diff
@@ -0,0 +1,544 @@
...
+pub(crate) fn ostree_zlib_file_header_to_regular(zlib_header_data: &AlignedSlice<A8>) -> Vec<u8> {
+    let data = gv!("(tuuuusa(ayay))").cast(zlib_header_data);
+    let (_size, uid, gid, mode, zero, symlink_target, xattrs_data) = data.to_tuple();
+    let mut s = Vec::<(&[u8], &[u8])>::new();
+    for x in xattrs_data.iter() {
+        let (key, value) = x.to_tuple();
+        s.push((key, value))
+    }
+
+    gv!("(uuuusa(ayay))").serialize_to_vec(&(*uid, *gid, *mode, *zero, symlink_target.to_str(), &s))
```
cgwalters: (Can we host some of these to nice documented `const` or is that not possible with the macro?)

crates/composefs-ostree/src/repo.rs
```diff
@@ -0,0 +1,544 @@
...
+        // Compute the checksum of (regular) header + data
+        let mut hasher = Sha256::new();
+        let regular_header = ostree_zlib_file_header_to_regular(&aligned_variant_data);
+        let sized_regular_header = size_prefix(&regular_header);
+        hasher.update(&*sized_regular_header);
+
+        // Decompress rest
+        let mut uncompressed = DeflateDecoder::new(compressed_data);
+
+        // TODO: Stream files into repo instead of reading it all
```
cgwalters: We just landed a lot of code for this for the OCI importer should be quite doable.

cgwalters (commented):
Just an initial pass

alexlarsson:
I rebased this and fixes some comments. Still some work to do though.

crates/composefs-ostree/src/pull.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        self.enqueue_fetch(commit_id, ObjectType::Commit);
+
+        // TODO: Support deltas
+
+        // TODO: At least for http we should make parallel fetches
+        while !self.outstanding.is_empty() {
+            let fetch = self.outstanding.pop_front().unwrap();
+            println!(
```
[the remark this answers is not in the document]
alexlarsson: added indicatif use

crates/composefs-ostree/src/pull.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    pub async fn pull_commit(&mut self, commit_id: &Sha256Digest) -> Result<ObjectID> {
+        let content_id = format!("ostree-commit-{}", hex::encode(commit_id));
+        if let Some(objid) = self.repo.has_stream(&content_id)? {
+            return Ok(objid);
+        }
+
+        self.enqueue_fetch(commit_id, ObjectType::Commit);
+
+        // TODO: Support deltas
```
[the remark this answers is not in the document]
alexlarsson: YAY!

crates/composefs-ostree/src/pull.rs
```diff
@@ -0,0 +1,272 @@
...
+        for f in files_data.iter() {
+            let (_name, checksum) = f.to_tuple();
+
+            self.maybe_fetch_file(checksum.try_into().unwrap());
+        }
+
+        for d in dirs_data.iter() {
+            let (_name, tree_checksum, meta_checksum) = d.to_tuple();
+
+            self.maybe_fetch_dirmeta(meta_checksum.try_into().unwrap());
```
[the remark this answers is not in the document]
alexlarsson: fixed

crates/composefs-ostree/src/repo.rs
alexlarsson: Unfortunately not, the macro needs a pure string

crates/composefs-ostree/src/repo.rs
alexlarsson: done

crates/composefs-ostree/src/repo.rs
```diff
@@ -0,0 +1,544 @@
...
+#[derive(Debug)]
+pub(crate) struct LocalRepo<ObjectID: FsVerityHashValue> {
+    repo: Arc<Repository<ObjectID>>,
+    mode: RepoMode,
+    dir: OwnedFd,
+    objects: OwnedFd,
+}
+
+impl<ObjectID: FsVerityHashValue> LocalRepo<ObjectID> {
+    pub fn open_path(
```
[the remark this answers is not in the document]
alexlarsson: This is not a public type atm though.

alexlarsson:
Ok, i updated this to the latest version and added streaming creation of repo files and parallelized fetching. Plus some other cleanups.

alexlarsson:
Ok, I sent some time on this, its now much more like the "cfsctl oci" commands and behavior, and it does parallel fetches. I also added various integration tests. I think this is pretty complete for what it does (i.e. imports ostree commits into composefs and lets you mount it). 

There are some TODOs for summary and delta support, but those are not necessarily super important for the basic functionallity.

crates/composefs-ctl/src/lib.rs
```diff
@@ -1566,6 +1641,111 @@ where
...
             }
         },
+        #[cfg(feature = "ostree")]
+        Command::Ostree { cmd: ostree_cmd } => match ostree_cmd {
+            OstreeCommand::PullLocal {
+                ref ostree_repo_path,
+                ref ostree_ref,
+                base_name,
+            } => {
+                eprintln!("Fetching {ostree_ref}");
```
cgwalters, line 1651: Don't log via `eprintln!` we have the progress API now.

Also on that topic...I think we should expose a varlink API for this now, right?

I guess neither of these need to strictly block merging though.

---

🤔 I guess actually...if we go down this varlink path, perhaps in theory we could have both the `oci` and `ostree` fetchers be extension binaries i.e. something like `/usr/libexec/composefs/ext/oci` is automatically `cfsctl oci`? That could be interesting...and would actually force us to have a good "core" varlink api.

crates/composefs-ostree/src/commit.rs
```diff
@@ -0,0 +1,637 @@
...
+            );
+        }
+        m.finish()
+    }
+}
+
+fn validate_buckets(buckets: &[u32; 256]) -> Result<()> {
+    for i in 1..256 {
+        // Bucket ends are (non-strictly) increasing
+        if buckets[i] < buckets[i - 1] {
```
cgwalters: In general in Rust many array accesses can be done more elegantly *and* more safely than just direct indexing. In this specific case I think https://doc.rust-lang.org/stable/std/primitive.slice.html#method.array_windows is what we want

crates/composefs-ostree/src/pull.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        let metadata_bar = ProgressBar::new_spinner().with_message("Fetching metadata");
+        let metadata_semaphore = Arc::new(Semaphore::new(MAX_CONCURRENT_METADATA_FETCHES));
+
+        // Phase 1: Fetch all metadata (commits, dirtrees, dirmetas) in parallel.
+        // Processing results may discover new metadata to fetch, so we loop
+        // until the queue is drained and all in-flight fetches have completed.
+        let mut join_set: JoinSet<Result<FetchResult<ObjectID>>> = JoinSet::new();
+
+        loop {
```
cgwalters, line 302: We *can* interleave metadata and data fetches, it's what libostree does. Is it worth the added complexity? Maybe not.

crates/composefs-ostree/src/commit.rs
```diff
@@ -0,0 +1,637 @@
+//! Ostree commit splitstream implementation
+
+/* Implementation of the ostree commit splitstream format
+ *
+ * Commit splitstreams are mappings from a set of ostree sha256
+ * digests into the content for that ostree object. The content is
+ * defined as some data, and an optional ObjectID referencing an
```
cgwalters: I read this thing twice and didn't fully understand it. "The content is defined as some data" is vague 😄 

---

Our goal is conceptually to define a serialization of an ostree commit into a single "stream", right? And then splitting out content objects as externals.

Hmm...why wouldn't it work to basically do what we do with tar, walk the commit in a depth-first manner, serializing metadata + externals as we go?

alexlarsson: Well, the "some data" is just generally the ostree object data for the digest.
We don't just want to have a serialization, because we also want to use the commit as a  way to efficiently look up ostree objects in the commit. We use this during pull to avoid pulling objects that was in the previous version of the commit.

alexlarsson: So, its more like a hash table from sha256 digests to objects that are optionally external object ids.

crates/composefs-ostree/src/pull.rs
alexlarsson: probably not. This thing is actually surprisingly fast as is:

```
$ time target/debug/cfsctl --repo repo ostree pull https://dl.flathub.org/repo runtime/org.gnome.Platform/x86_64/50
Fetching runtime/org.gnome.Platform/x86_64/50
█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 16288/16288commit  f6fb972824514aefc06b23d7d591192c9cba2ad72648bf0473d06a565c40c264
verity  a2e16123b2310d65b6b886e89e4fc45ec61c47efd0b4daac05bd3132ac3d8f78890fcbb22d0975411917742d17e86ad52709c8ec1ff33a62b97e94f38528242a
image   079ade7aba4e5fef51b30a80e3d1805fd5f79d9a48f240340c7a94f98497d8265edee857563308c00574a1bab792168caa0152af04ef47ac15d8a935ca291e15
tagged  runtime/org.gnome.Platform/x86_64/50
objects 2752 metadata + 16288 files fetched

real	0m13,006s
user	0m12,066s
sys	0m1,663s
$ du -csh repo
1,1G	repo
1,1G	total

```

crates/composefs-ostree/src/commit.rs
alexlarsson: And its not actually "inline data OR external refs", we sometimes have to have both, because we need to store metadata as well. So, more like we store the archive-z2 file header in the data, and then content in the external ref.

cgwalters: > We don't just want to have a serialization, because we also want to use the commit as a way to efficiently look up ostree objects in the commit.

Would parsing all of the metadata be expensive though?

alexlarsson: I mean, would it be impossible to change it to something else? For sure not, but what is the point? Allison and I spent a fair amount of time creating the new splitstream format specifically for this use. So it is the format we have, and its intended to be efficient for what we use it for.

cgwalters: OK, fair! But can you spend some tokens clarifying the docs a bit at least?

I get the efficiency idea, but one thing that seems odd to me right now is that because we store this ostree-specific thing in the split stream content, it ends up zstd compressed. So we're at least reading the whole thing into RAM, we can't mmap etc.

With putting tar in split stream, this all made sense because we basically don't look at the tar stream unless we're copying the image out.

Also, while I get that it was nontrivial to design the format, there's also the traditional "cost of maintenance > cost of writing" to consider. Splitstream is a good bit of complexity on its own, but I think it's turned out mostly OK because for the OCI case it basically is a wrapper for a very well known thing - tar (ok well tar is a mess too, but it's a well-known mess). This work here is combining split stream with two entirely different more bespoke formats (splitstream-ostree and ostree).

I guess one way I'd say this is if you have a data format, it should have the ability to be converted to JSON, have a "structure checker" like fsck etc.

I'm aware I ~lost this argument before but e.g. https://cbor.io is pretty widely used. Does pulling in cbor for just this secondary bespoke binary format have the right cost/benefit? Perhaps not. (But, since we already need it: why not gvariant?)

---

Dunno. This is a discussion, nothing I am saying here is blocking.

alexlarsson: I'll update the docs to be be more readable, comprehensive and documenting the final/current state of things. And, I agree that having them zstd compressed does make it a bit weird for this to claim "efficiency", although I sort of agree with Allisons more modern view of mmap and its problems. 

That said, I fundamentally think a "bucket of sha256 indexed objects" is the more correct format for an ostree thing. Serializing an ostree commit tree just feels wrong. Like, would you then duplicate things that were shared in many places (like dirmetas, or hardlinks)?

cgwalters:
@allisonkarlitskaya You have a "changes requested" here which blocks merges

crates/composefs-ostree/src/commit.rs
cgwalters: > I'll update the docs to be be more readable, comprehensive and documenting the final/current state of things. 

Thanks.

> That said, I fundamentally think a "bucket of sha256 indexed objects" is the more correct format for an ostree thing. Serializing an ostree commit tree just feels wrong. Like, would you then duplicate things that were shared in many places (like dirmetas, or hardlinks)?

No, I think the obvious flattened serialization would just have "each object is emitted once" semantics. Hardlinks are implicit in the ostree format - the data doesn't have `st_nlink`.

crates/composefs-ostree/src/commit.rs
alexlarsson: array_windows is unstable though, do we really want to use that?

alexlarsson: I used regular .windows() instead. Also, I spent some time in general rustifying the code and cleaning it up.

crates/composefs-ostree/src/commit.rs
alexlarsson: I added doc/ostree.md which has a more detailed documentation on the format, including some general faffing about ostree and how this is supposed to be used.

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
+
+A splitstream is normally a series of internal and external chunks,
+but the ostree commit uses only one inline chunk. This chunk is
+basically a serialized form of the "objects" directory of an OSTree
+repository. I.e. it has a mapping of sha256 to ostree object data.
+All objects except file objects are stored in the standard ostree
+object format.
+
+OSTree file objects are stored in the archive-z2 format, except not
+compressed, and optionally the file content part of it may be stored
```
cgwalters, line 66: The `-z2` is obsolete terminology, it's just `archive`.

Also: the core library calls the "metadata+data" an "object stream". It's really pretty similar to a tar header + content.

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
+as referencing the index of an external object. The z2 format is,
```
cgwalters, line 67: So not "z2 format" just "object format".

Like: when we do `ostree fsck` we reconstruct in memory that format and that's what we compare vs the digest - there's no compression involved.

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
+lookups in it using a binary search.  The buckets in the header can be
+used to quickly limit the binary search based on the first byte of a
+digest.
+
+Then, at the same index as the binary searched object you can look up
+the object info which gives you the offset/length of the object
+content data and optionally a reference to an external object.
+
+The exact form of the data looks like this, packed in order from the
+start of the splitstream content. All ints are in little endian.
```
cgwalters, line 97: How about "numbers" or "integers". "ints" is a bit too informal...

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
+| [u32; 256]: end index of bucket   |
++-----------------------------------+
+```
+
+The bucket list contains the end index (in the object ids table) of
+objects starting with that particular byte, and can be used to quickly
+limit the search.  We can also compute the total number of objects
+(n_objects) by looking in the last bucket.
+
+### Object ids
```
cgwalters, line 113: Should re-mention the sorting here

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
+```
+ n_objects x
++-----------------------------------+
+|  [u8; 32] ostree object id        |
```
cgwalters, line 117: (binary sha256) would help clarify

doc/ostree.md
```diff
@@ -0,0 +1,139 @@
...
++-----------------------------------+
+```
+
+### Object Info
+```
+ n_objects x
++-----------------------------------+
+| u32: Offset to per-object data    |
```
cgwalters, line 125: I'd *hope* no one ever wants to ship more than 4G of metadata...but perhaps worth calling out. Or...perhaps instead of raw "offset" since we have the alignment requirement, we could make it "block" offset with a 8 byte block. We'd then go up to 32G of metadata. But dunno.

crates/composefs-ostree/src/commit.rs
```diff
@@ -0,0 +1,588 @@
...
+        );
+
+        // Compute bucket ends
+        for e in self.map.iter() {
+            // Initially end is just the count
+            header.bucket_ends[e.ostree_id[0] as usize] += 1;
+        }
+        for i in 1..256 {
+            // Then we sum them up to the end
+            header.bucket_ends[i] += header.bucket_ends[i - 1];
```
cgwalters, line 227: Same comment here re `windows()`

crates/composefs-ostree/src/repo.rs
```diff
@@ -0,0 +1,509 @@
...
+            hex::encode(actual_checksum),
+            hex::encode(expected_checksum)
+        );
+    }
+
+    Ok((file_data, obj_id))
+}
+
+/// Abstraction over local and remote ostree repository access.
+pub(crate) trait OstreeRepo<ObjectID: FsVerityHashValue>: Send + Sync {
```
cgwalters, line 95: One operation I think we want here is `fsck` too at least

cgwalters (approved):
Overall: Nothing blocking, from my PoV we can get this in as a clearly experimental separate crate and iterate from there.

A lot of interesting stuff here, but I think we need to write up an even more concrete plan for how consumers (flatpak?) would use this - gets into the varlink (or a -capi crate?) discussion.

cgwalters:
> A lot of interesting stuff here, but I think we need to write up an even more concrete plan for how consumers (flatpak?) would use this - gets into the varlink (or a -capi crate?) discussion.

Or to say it a different way - I think we do need a tight focus on the "core" things to get to the [current milestones](https://github.com/composefs/composefs-rs/milestones). I didn't add this to a milestone, but if we did I think it would be e.g. 1.1 or so? I am happy to debate it of course, we could say that this is part of 1.0 too.

What I keep thinking about again here is making it really clear and obvious how to use composefs to store arbitrary content without writing a custom crate and glue. One idea here is: have a (varlink) API that accepts metadata + tarball and stores it as an OCI artifact, but where we do the splitstream stuff and support mounting it too. Think e.g. storing a dpkg/rpm in composefs.

alexlarsson:
@cgwalters I agree, we should have a way that lets you just store unstructured data from a tar or a directory, with no guarantees of "exact bitwise reproduceability" of the original data. That seems very useful, however its not quite the same as the oci/ostree case where we want to roundtrip both ways losslessly.

Anyway, lots of good comments here, but I'm gonna merge this for now and lets consider this unstable and then I'll do updates to it in separate PRs, because this one is getting unwieldy.

crates/composefs-ostree/src/repo.rs
alexlarsson: What would this do? fsck a local/remote ostree repo? When would you use that?

alexlarsson: Or, do you mean some fsck operation of a local commit splitstream?

crates/composefs-ostree/src/commit.rs
alexlarsson: Can't write to the window though.

alexlarsson:
I resolved all comments fixed in #340

Issue #87, opened 2025-04-02
investigate optional fs-verity


allisonkarlitskaya:
We want to use this stuff on systems without fs-verity support (like RHEL systems with xfs).

We need to teach composefs-setup-root (and the mount API in general) about how to deal with no fs-verity:
 - disable the fs-verity mount option on the composefs overlayfs
 - don't measure the erofs image before mounting it (or at least accept that the measurement might return ENOTTY, and don't fail)

We also need to investigate how to deal with things on the writer side.  We do our fs-verity merkle tree calculations in userspace before writing to the disk, so we're already independent of kernel fs-verity support there.  We could just skip the 'enable' call, or don't abort if it fails with a "not-supported"-type errno.

Then the largest unsolved problem is what images want to do.  They can hardcode the fs-verity requirement into the composefs-setup-root config in their initramfs, but they don't know if the user will be deploying to xfs or so.  We need to figure out a way to make sure nothing goes sideways here...

Also: in general, "verify fs-verity if it exists, but don't fail if it's unsupported" is useful for more casual data integrity purposes, even if it has very little value as a security measure (as it's trivial to circumvent).

cgwalters:
> Then the largest unsolved problem is what images want to do. They can hardcode the fs-verity requirement into the composefs-setup-root config in their initramfs, but they don't know if the user will be deploying to xfs or so. We need to figure out a way to make sure nothing goes sideways here...

Since I see this is part of the bootc milestone; from that PoV it seems quite simple, we have to have it disabled by default and enabled via opt-in using the config files that already exist, right?

But it'd make sense (since this is a new project) to default to requiring fsverity, we just need a flag to disable it, which could probably just be a `set_verity_not_required()` API call?

> even if it has very little value as a security measure (as it's trivial to circumvent).

I think the person who contributed the patch to ostree to require fsverity was planning to do something related to IMA to verify the composefs file (distinct from the kernel commandline/UKI binding) and in general I think that can make sense.

allisonkarlitskaya:
We're going to do this.  Rough plan is like this:

 - `Repository` will get a flag to enable/disable mandatory fs-verity checking
   - we can try for it opportunistically, though
 - composefs-setup-root will inspect `/proc/cmdline` for some indication that it is allowed to enable "best effort" mode.  maybe `composefs=?0123abcd...` or a `composefs.fsverity=0` or so
   - in this case, we'll check the fs-verity of the erofs image
[...]

PR #147, opened 2025-06-19
Make fs-verity optional

croissanne:
- adds "insecure" option to repository, allows images which do not have fs-verity enabled
- if `?` is prepended to composefs= cmdline, set repository to insecure
- opportunistically enable fs-verity where appropriate, even if repository is in insecure mode
- add `--insecure` to cfsctl

#87

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -26,6 +26,8 @@ jobs:
           - { dir: 'uki', os: 'fedora' }
           - { dir: 'unified', os: 'fedora' }
           - { dir: 'unified-secureboot', os: 'fedora' }
+          - { dir: 'bls', os: 'arch', fsfmt: 'ext4', verity: 'none' }
+          - { dir: 'bls', os: 'arch', fsfmt: 'xfs', verity: 'none' }
       fail-fast: false
 
     steps:
@@ -71,4 +73,6 @@ jobs:
     - name: Run example tests
       run: |
         export PATH="${HOME}/bin:${PATH}"
+        export FS_FORMAT=${{ matrix.example.fsfmt }}
+        export FS_VERITY_MODE=${{ matrix.example.verity }}
         examples/test/run ${{ matrix.example.dir }} ${{ matrix.example.os }}
--- a/crates/cfsctl/src/main.rs
+++ b/crates/cfsctl/src/main.rs
@@ -339,7 +339,7 @@ async fn main() -> Result<()> {
             fs.print_dumpfile()?;
         }
         Command::Mount { name, mountpoint } => {
-            repo.mount(&name, &mountpoint)?;
+            repo.mount_at(&name, &mountpoint)?;
         }
         Command::ImageObjects { name } => {
             let objects = repo.objects_for_image(&name)?;
--- a/crates/cfsctl/src/main.rs
+++ b/crates/cfsctl/src/main.rs
@@ -27,6 +27,12 @@ pub struct App {
     #[clap(long, group = "repopath")]
     system: bool,
 
+    /// Sets the repository to insecure before running any operation and
+    /// prepend '?' to the composefs kernel command line when writing
+    /// boot entry.
+    #[clap(long)]
+    insecure: bool,
+
     #[clap(subcommand)]
     cmd: Command,
 }
@@ -160,7 +166,7 @@ async fn main() -> Result<()> {
 
     let args = App::parse();
 
-    let repo: Repository<Sha256HashValue> = (if let Some(path) = &args.repo {
+    let mut repo: Repository<Sha256HashValue> = (if let Some(path) = &args.repo {
         Repository::open_path(CWD, path)
     } else if args.system {
         Repository::open_system()
@@ -171,6 +177,8 @@ async fn main() -> Result<()> {
     } else {
         Repository::open_user()
     })?;
+
+    repo.set_insecure(args.insecure);
 
     match args.cmd {
         Command::Transaction => {
--- a/crates/cfsctl/src/main.rs
+++ b/crates/cfsctl/src/main.rs
@@ -291,6 +291,7 @@ async fn main() -> Result<()> {
                     &repo,
                     entry,
                     &id,
+                    args.insecure,
                     bootdir,
                     None,
                     entry_id.as_deref(),
--- a/crates/composefs-boot/src/write_boot.rs
+++ b/crates/composefs-boot/src/write_boot.rs
@@ -3,13 +3,13 @@ use std::{
     path::Path,
 };
 
-use anyhow::{bail, ensure, Result};
+use anyhow::{ensure, Result};
 
 use composefs::{fsverity::FsVerityHashValue, repository::Repository};
 
 use crate::{
     bootloader::{read_file, BootEntry, Type1Entry, Type2Entry},
-    cmdline::get_cmdline_value,
+    cmdline::get_cmdline_composefs,
     uki,
 };
 
@@ -18,6 +18,7 @@ pub fn write_t1_simple<ObjectID: FsVerityHashValue>(
     bootdir: &Path,
     boot_subdir: Option<&str>,
     root_id: &ObjectID,
+    insecure: bool,
     cmdline_extra: &[&str],
     repo: &Repository<ObjectID>,
 ) -> Result<()> {
@@ -29,7 +30,7 @@ pub fn write_t1_simple<ObjectID: FsVerityHashValue>(
     };
 
     t1.entry
-        .adjust_cmdline(Some(&root_id.to_hex()), cmdline_extra);
+        .adjust_cmdline(Some(&root_id.to_hex()), insecure, cmdline_extra);
 
     // Write the content before we write the loader entry
     for (filename, file) in &t1.files {
@@ -59,13 +60,11 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
     create_dir_all(&efi_linux)?;
     let filename = efi_linux.join(t2.filename.as_ref());
     let content = read_file(&t2.file, repo)?;
-    let Some(composefs) = get_cmdline_value(uki::get_cmdline(&content)?, "composefs=") else {
-        bail!("The UKI is missing a composefs= commandline parameter");
-    };
-    let expected = root_id.to_hex();
+    let (composefs, _) = get_cmdline_composefs::<ObjectID>(uki::get_cmdline(&content)?)?;
+
     ensure!(
-        composefs == expected,
-        "The UKI has the wrong composefs= parameter (is '{composefs}', should be {expected})"
+        &composefs == root_id,
+        "The UKI has the wrong composefs= parameter (is '{composefs:?}', should be {root_id:?})"
     );
     write(filename, content)?;
     Ok(())
@@ -99,10 +98,12 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
 /// * entry_id       - In case of a BLS entry, the name of file to be generated in `loader/entries`
 /// * cmdline_extra  - Extra kernel command line arguments
 ///
+#[allow(clippy::too_many_arguments)]
 pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
     repo: &Repository<ObjectID>,
     entry: BootEntry<ObjectID>,
     root_id: &ObjectID,
+    insecure: bool,
     boot_partition: &Path,
     boot_subdir: Option<&str>,
     entry_id: Option<&str>,
@@ -118,6 +119,7 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
                 boot_partition,
                 boot_subdir,
                 root_id,
+                insecure,
                 cmdline_extra,
                 repo,
             )?;
@@ -140,6 +142,7 @@ pub fn write_boot_simple<ObjectID: FsVerityHashValue>(
                 boot_partition,
                 boot_subdir,
                 root_id,
+                insecure,
                 cmdline_extra,
                 repo,
             )?;
--- a/crates/composefs-oci/src/lib.rs
+++ b/crates/composefs-oci/src/lib.rs
@@ -152,7 +152,7 @@ pub fn mount<ObjectID: FsVerityHashValue>(
     let Some(id) = config.get_config_annotation("containers.composefs.fsverity") else {
         bail!("Can only mount sealed containers");
     };
-    repo.mount(id, mountpoint)
+    repo.mount_at(id, mountpoint)
 }
 
 #[cfg(test)]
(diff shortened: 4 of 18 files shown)


crates/composefs/src/repository.rs
```diff
@@ -58,6 +59,7 @@ pub struct Repository<ObjectID: FsVerityHashValue> {
     repository: OwnedFd,
     objects: OnceCell<OwnedFd>,
     _data: std::marker::PhantomData<ObjectID>,
+    insecure: bool,
```
allisonkarlitskaya, line 62: Probably move that above the `PhantomData`...

crates/composefs/src/repository.rs
```diff
@@ -86,6 +88,7 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
             repository,
             objects: OnceCell::new(),
             _data: std::marker::PhantomData,
+            insecure: false,
```
allisonkarlitskaya, line 91: ...ditto

crates/composefs/src/repository.rs
```diff
@@ -127,7 +130,9 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
             Ok(fd) => {
                 // measure the existing file to ensure that it's correct
                 // TODO: try to replace file if it's broken?
-                ensure_verity_equal(fd, &id)?;
+                if !self.insecure {
```
allisonkarlitskaya, line 133: As a general rule I think we always want to attempt to measure fs-verity even in insecure mode, but allow that it fails with "not supported" errors.

crates/composefs/src/repository.rs
```diff
@@ -151,8 +156,10 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         )?;
         drop(file);
 
-        enable_verity::<ObjectID>(&ro_fd).context("Enabling verity digest")?;
-        ensure_verity_equal(&ro_fd, &id).context("Double-checking verity digest")?;
+        if !self.insecure {
```
allisonkarlitskaya: I think we also want to try opportunistically to enable it...

crates/composefs/src/repository.rs
```diff
@@ -175,10 +182,17 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
 
     fn open_with_verity(&self, filename: &str, expected_verity: &ObjectID) -> Result<OwnedFd> {
         let fd = self.openat(filename, OFlags::RDONLY)?;
-        ensure_verity_equal(&fd, expected_verity)?;
+        if !self.insecure {
```
allisonkarlitskaya, line 198: ditto

crates/composefs/src/repository.rs
```diff
@@ -220,6 +234,10 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
 
     /// Basically the same as has_stream() except that it performs expensive verification
     pub fn check_stream(&self, sha256: &Sha256Digest) -> Result<Option<ObjectID>> {
+        if self.insecure {
+            return self.has_stream(sha256);
```
allisonkarlitskaya, line 251: This isn't sufficient, I'm afraid.  We want to iterate the full structure to check that it's consistent and matches the given sha256 body digest, even if we don't have fs-verity.

crates/composefs/src/repository.rs
```diff
@@ -383,15 +401,21 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
         let image = self.openat(&format!("images/{name}"), OFlags::RDONLY)?;
 
-        if !name.contains("/") {
+        if !name.contains("/") && !self.insecure {
             // A name with no slashes in it is taken to be a sha256 fs-verity digest
             ensure_verity_equal(&image, &ObjectID::from_hex(name)?)?;
         }
 
-        Ok(image)
+        match measure_verity::<ObjectID>(&image) {
```
allisonkarlitskaya, line 425: You're effectively measuring and comparing verity twice here... You left the check above...

crates/composefs/src/repository.rs
```diff
@@ -383,15 +401,21 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
+            Ok(found) if found == FsVerityHashValue::from_hex(name)? => Ok((image, true)),
```
allisonkarlitskaya, line 426: `name` isn't always hex: it might be a ref with `/` in it.  That's what the check above is about.

crates/composefs/src/repository.rs
```diff
@@ -383,15 +401,21 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         self.write_image(Some(name), &data)
     }
 
-    pub fn open_image(&self, name: &str) -> Result<OwnedFd> {
+    fn open_image(&self, name: &str) -> Result<(OwnedFd, bool)> {
```
allisonkarlitskaya, line 417: Depending on how many users this API has, it might make sense to inline it where it's used... I guess the GC thing is using it as well, though......

I'm glad it's no longer public, at least.  A short doc about what the 'bool' is would certainly be appreciated though.

crates/composefs/src/mount.rs
```diff
@@ -78,14 +78,21 @@ pub fn erofs_mount(image: OwnedFd) -> Result<OwnedFd> {
     )?)
 }
 
-pub fn composefs_fsmount(image: OwnedFd, name: &str, basedir: impl AsFd) -> Result<OwnedFd> {
+pub fn composefs_fsmount(
+    image: OwnedFd,
+    name: &str,
+    basedir: impl AsFd,
+    enable_verity: bool,
```
allisonkarlitskaya, line 85: Almost feels like this should be a flags field.... in which case disable-verity would be the opt-in.  Can we make this function `pub(crate)`?  Ideally this will only ever be called via `Repository` now, which will get this logic correct...

crates/composefs/src/repository.rs
```diff
@@ -416,15 +416,25 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
+            self.objects_dir()?,
+            enable_verity,
+        )?)
     }
 
-    pub fn mount(&self, name: &str, mountpoint: &str) -> Result<()> {
-        let image = self.open_image(name)?;
+    pub fn mount_at(&self, name: &str, mountpoint: &str) -> Result<()> {
+        let (image, enable_verity) = self.open_image(name)?;
         Ok(mount_composefs_at(
```
allisonkarlitskaya, line 445: Can we maybe get rid of `mount_composefs_at()` in `mount.rs` and use `move_mount(self.mount(name), mountpoint)` here instead (or whatever `mount_at` internal helper we have?)

crates/composefs-setup-root/src/main.rs
```diff
@@ -198,12 +198,19 @@ fn mount_subdir(
...
-fn parse_composefs_cmdline<H: FsVerityHashValue>(cmdline: &str) -> Result<H> {
-    let Some(digest) = get_cmdline_value(cmdline, "composefs=") else {
+fn parse_composefs_cmdline<H: FsVerityHashValue>(cmdline: &str) -> Result<(H, bool)> {
+    let Some(mut digest) = get_cmdline_value(cmdline, "composefs=") else {
         bail!("Unable to find composefs= cmdline parameter");
     };
 
-    H::from_hex(digest).context("Parsing composefs=")
+    let mut insecure = false;
+    if let Some(stripped) = digest.strip_prefix('?') {
```
allisonkarlitskaya, line 207: I think we were going to move this code into `cmdline.rs` in the `composefs-boot` crate?  We might also use `~` instead of `?`.

crates/cfsctl/src/main.rs
```diff
@@ -339,7 +339,7 @@ async fn main() -> Result<()> {
             fs.print_dumpfile()?;
         }
         Command::Mount { name, mountpoint } => {
-            repo.mount(&name, &mountpoint)?;
+            repo.mount_at(&name, &mountpoint)?;
```
allisonkarlitskaya, line 347: Please merge this into the commit that made this change.  Each commit ought to compile cleanly and pass tests.

crates/composefs-oci/src/lib.rs
```diff
@@ -152,7 +152,7 @@ pub fn mount<ObjectID: FsVerityHashValue>(
     let Some(id) = config.get_config_annotation("containers.composefs.fsverity") else {
         bail!("Can only mount sealed containers");
     };
-    repo.mount(id, mountpoint)
+    repo.mount_at(id, mountpoint)
```
allisonkarlitskaya, line 155: ditto.  Please merge that into the original commit.

crates/cfsctl/src/main.rs
```diff
@@ -27,6 +27,9 @@ pub struct App {
     #[clap(long, group = "repopath")]
     system: bool,
 
+    #[clap(long)]
```
allisonkarlitskaya, line 30: A small doc here might be nice.

crates/composefs/src/repository.rs
```diff
@@ -156,7 +156,20 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         )?;
         drop(file);
 
-        if !self.insecure {
+        if self.insecure {
+            match measure_verity::<ObjectID>(&ro_fd) {
```
allisonkarlitskaya, line 160: The measure here doesn't make much sense.  We just created this, so the answer will always be either that there's no fs-verity on the file, or that fs-verity is unsupported.  But you could get that information by just trying to enable it and deal with the errors.  I think you could probably also merge the code paths with the `else` (original) case... The only different you really need is what to do if the enabling fails with the not-supported error: in secure mode, that's fatal.  In insecure mode, you ignore it.

allisonkarlitskaya: Also please merge this into the first commit which adds `Repository.insecure` in the first place.

.github/workflows/examples.yml
```diff
@@ -26,6 +26,7 @@ jobs:
           - { dir: 'uki', os: 'fedora' }
           - { dir: 'unified', os: 'fedora' }
           - { dir: 'unified-secureboot', os: 'fedora' }
+          - { dir: 'nofsverity', os: 'arch' }
```
allisonkarlitskaya, line 29: I think you might have forgotten to also add the directory in this commit :)

crates/composefs-boot/src/write_boot.rs
```diff
@@ -59,9 +59,14 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
     create_dir_all(&efi_linux)?;
     let filename = efi_linux.join(t2.filename.as_ref());
     let content = read_file(&t2.file, repo)?;
-    let Some(composefs) = get_cmdline_value(uki::get_cmdline(&content)?, "composefs=") else {
+    let Some(mut composefs) = get_cmdline_value(uki::get_cmdline(&content)?, "composefs=") else {
         bail!("The UKI is missing a composefs= commandline parameter");
     };
+
+    if let Some(stripped) = composefs.strip_prefix('?') {
```
allisonkarlitskaya, line 66: Yes, please add a new function to the end of `cmdline.rs`... You should probably also make this change at the same time as you change `composefs-setup-root` with the same change.

examples/common/make-image
```diff
@@ -1,12 +1,47 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eux
 
-output="$1"
+FIX_VERITY=0
```
allisonkarlitskaya, line 5: The user should be able to specify `FIX_VERITY=1` from the outside...

examples/common/make-image
```diff
@@ -1,12 +1,47 @@
...
+SETUP_REPART=0
+FS_FORMAT=ext4
+parsecmd() {
+    while [ $# -gt 0 ]; do
+        case "$1" in
+        #CMDSTART
+        -fv|--fix-verity) #
```
allisonkarlitskaya, line 12: Please don't do two-character 'short' opts...

examples/common/run-repart
```diff
@@ -9,7 +9,7 @@ chcon system_u:object_r:etc_t:s0 tmp/sysroot/state/*/etc/*
 
 definitions="${0%/*}/repart.d"
 
-if [ "${FIX_VERITY:-}" = '1' ]; then
+if [ "$SETUP_REPART" = '1' ]; then
```
allisonkarlitskaya, line 12: This variable is hilariously poorly named :rofl:

allisonkarlitskaya: It might be interesting to make this setup a bit more 'neutral' by dynamically generating the `repart.d` directory in the `tmp/` dir on each run using some sort of a heredoc-based approach... It's only two small files and it would be a bit cleaner than sed-hacking, I guess.

examples/common/make-image
```diff
@@ -1,12 +1,47 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eux
 
-output="$1"
+FIX_VERITY=0
+SETUP_REPART=0
+FS_FORMAT=ext4
+parsecmd() {
```
allisonkarlitskaya, line 8: This commandline interface thing seems a bit overkill to be honest...

It might be nicer to just be able to set a couple of extra envvars to modify the test environment, from outside:
 - an envvar to disable fs-verity complete, impacting both systemd-repart and the fix-verity script
 - an envvar to let choosing a different type for the rootfs (xfs, etc.)

Maybe we could avoid having a whole separate test scenario...

examples/nofsverity/build
```diff
@@ -0,0 +1,56 @@
...
+    .
+
+BASE_ID="$(sed s/sha256:// tmp/base.iid)"
+${CFSCTL} oci pull containers-storage:"${BASE_ID}"
+BASE_IMAGE_FSVERITY="$(${CFSCTL} oci compute-id --bootable "${BASE_ID}")"
+
+${PODMAN_BUILD} \
+    --iidfile=tmp/final.iid \
+    --build-context=base="container-image://${BASE_ID}" \
+    --build-arg=COMPOSEFS_FSVERITY="?${BASE_IMAGE_FSVERITY}" \
```
allisonkarlitskaya, line 46: It seems like just about the only thing changed is that you add the `?` here and the `-sr` below when building the image, ya?  I think that's probably not worth a whole separate scenario....

allisonkarlitskaya: It might also be interesting to try building an image without fs-verity but with the `composefs=` cmdline specified without the `?`, ie: to make sure that it fails as expected.

allisonkarlitskaya (changes requested):

crates/composefs/src/repository.rs
croissanne: Right, if it does I guess we can just skip the entire thing and return (image, true)?

allisonkarlitskaya (changes requested):

allisonkarlitskaya (approved):
:green_heart:

PR #148, opened 2025-06-24
fsverity: add FdVerity trait and default impl for AsFd

jeckersb:
These bare module-level functions were all operating on file
descriptors, I think it's slightly more ergonomic/idiomatic to bundle
them together and have them implemented directly on the file
descriptor instead.

### DIFF
--- a/crates/composefs/src/fsverity/mod.rs
+++ b/crates/composefs/src/fsverity/mod.rs
@@ -63,86 +63,91 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
     digest::FsVerityHasher::<H, 12>::hash(data)
 }
 
-/// Enable fs-verity on the given file.
-///
-/// This essentially boils down to the FS_IOC_ENABLE_VERITY ioctl.
-///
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
First takeaways:
 - this looks nice when you use it
 - the implementation is maybe a bit magic.  I always hesitate on trait impls for all types that happen to implement another trait...
 - (taking rustix as a comparison) most APIs I've seen that operate on fds tend to take the fd as a `impl AsFd` as the (explicit) first argument.  This flies in the face of that convention.

cgwalters:
I don't have a *strong* opinion here, but I think I have a weak preference for the status quo just based on the rationale that there's other `impl AsFd` that don't support fsverity (directory fds) and we don't have *that* much code that does fsverity stuff in the end.

allisonkarlitskaya:
> I have a weak preference for the status quo ... and we don't have that much code that does fsverity stuff in the end.

:100:

jeckersb:
:+1: 

I'll close it, I mentioned in matrix I don't feel strongly about it, it was just something I had laying around in a branch and threw it up to see how others felt.

Issue #38, opened 2024-11-20
Suggested layout for sysroot for bootc integration


travier:
Talking with Allison, when flushed the following potential design for the sysroot layout which would allow us to transition existing ostree based bootc systems to pure composefs ones: 

```
.
├── composefs
│   ├── images
│   ├── objects
│   └── streams
├── ostree
│   └── deploy
│       └── fedora
│           ├── deploy
│           │   └── hash.0
│           └── var
└── state
    ├── deploy
    │   ├── composefshash0
    │   │   ├── etc
    │   │   └── var -> ../../../ostree/deploy/fedora/var
    │   └── composefshash1
    │       ├── etc
    │       └── var -> ../../os/fedora/var
    └── os
        └── fedora
            └── var
```

allisonkarlitskaya:
## Notes
 - we're not sure if `/state` should be toplevel or if it should live in `/composefs/`.  This is considered an open question.
 - we normally want the composefs ID alone to select the state as well.  Things get complicated with UKIs if we have the ability to have multiple sets of state that should be accessible from a single bootable image.
   - but, for unusual situations where we *can* modify the kernel commandline, we might consider a `state=` argument that explicitly selects a different state directory
 - we assume that a given composefs image will correspond to a single OS (in the ostree stateroot sense).  That lets us consolidate all deployments in a single namespace.
 - on boot, we open the `/state/` directory that corresponds to the `composefs=` argument (or maybe `state=` argument).  Inside of that directory might exist `etc` and `var` either as directories or symlinks.  If they are directories, we mount them.  If they are symlinks, we mount what they point at.  That supports flexibility over different scenarios:
   - the usual situation is that we want to share one `/var` per OS.  The diagram above illustrates how that would look (`/state/os/fedora`, with symlinks from each deployment pointing to it eg. `composefshash1`)
   - sharing a `/var` with a sysroot from ostree, very useful for migrating or sharing (eg. `composefshash0`)
   - per-deployment `/var` (possibly created using btrfs snapshots on the sysroot filesystem directly)
   - no `/var`: if `/var` is mounted from a separate partition using `/etc/fstab` or GPT or otherwise by systemd
   - we could even mount the entire `/sysroot` itself as `/var` if we wanted
 - the main point here is that this is decided/managed by deployment code.  As the system is booting it just looks in the state directory and does what it's told.

## About /etc
[...]

PR #149, opened 2025-06-25
Update state path

Johan-Liebert1:
1. Log errors in `composefs-setup-root` for `open_dir` and `open_tree` functions
2. Move `/etc` and `/var` into `/sysroot/state/deploy/<image_id>` from `/sysroot/state/<image_id>` according to https://github.com/containers/composefs-rs/issues/38#issuecomment-3003931710

### DIFF
--- a/crates/composefs-boot/src/write_boot.rs
+++ b/crates/composefs-boot/src/write_boot.rs
@@ -77,6 +77,8 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
 /// * repo           - The composefs repository
 /// * entry          - Boot entry variant to be written
 /// * root_id        - The content hash of the generated EROFS image id
+/// * insecure       - Make fs-verity validation optional in case the filesystem doesn't support
+///   it, indicated by `composefs=?hash` cmdline argument
 /// * boot_partition - Path to the boot partition/directory
 /// * boot_subdir    - If `Some(path)`, the path is prepended to `initrd` and `linux` keys in the BLS entry
 ///


crates/composefs-setup-root/src/main.rs
```diff
@@ -116,6 +120,9 @@ fn bind_mount(fd: impl AsFd, path: &str) -> rustix::io::Result<OwnedFd> {
             | OpenTreeFlags::OPEN_TREE_CLOEXEC
             | OpenTreeFlags::AT_EMPTY_PATH,
     )
+    .inspect_err(|_| {
+        eprintln!("Open tree failed for {path}");
```
allisonkarlitskaya, line 124: Interesting.  I don't know this pattern.  I guess it makes sense with this not being a library...

crates/composefs-setup-root/src/main.rs
```diff
@@ -257,7 +257,10 @@ fn setup_root(args: Args) -> Result<()> {
     }
 
     // etc + var
-    let state = open_dir(open_dir(&sysroot, "state")?, image.to_hex())?;
+    let state = open_dir(
+        open_dir(open_dir(&sysroot, "state")?, "deploy")?,
```
allisonkarlitskaya, line 261: Could probably just combine "state/deploy" here, right?

crates/cfsctl/src/main.rs
```diff
@@ -303,7 +303,7 @@ async fn main() -> Result<()> {
                     .as_ref()
                     .map(|p: &PathBuf| p.parent().unwrap())
                     .unwrap_or(Path::new("/sysroot"))
-                    .join("state")
+                    .join("state/deploy")
```
allisonkarlitskaya, line 306: Yup.  I agree: I noticed a while ago that I did the wrong thing here.  Thanks for fixing it.

crates/composefs/tests/test.sh
```diff
@@ -35,7 +35,7 @@ EOF
 blkdev="$(mkd "${top}/blkdev")"
 repo="$(mkd "${blkdev}/composefs")"
 state="$(mkd "${blkdev}/state")"
-deployment="$(mkd "${state}/${imageid}")"
+deployment="$(mkd "${state}/deploy/${imageid}")"
```
allisonkarlitskaya, line 38: Thanks for fixing this even though it's kinda abandonware (read: everything not executed by CI).  Maybe we should wire that up to CI somehow...

crates/composefs-boot/src/write_boot.rs
```diff
@@ -77,6 +77,7 @@ pub fn write_t2_simple<ObjectID: FsVerityHashValue>(
 /// * repo           - The composefs repository
 /// * entry          - Boot entry variant to be written
 /// * root_id        - The content hash of the generated EROFS image id
+/// * insecure       - If true, skips fs-verity validation for the EROFS image
```
allisonkarlitskaya, line 80: Thanks!

Is there some documentation linter we could be running here to catch stuff like this?

The description is slightly accurate: it doesn't disable fs-verity validation but it does make it optional if the filesystem doesn't support it.  If we find a fs-verity validated image then we'll still validate all the files at runtime.

It's also about what ends up in the composefs= cmdline.  We might make that a bit more clear.

crates/composefs/tests/test.sh
Johan-Liebert1: ah, I didn't know this wasn't used

crates/composefs-boot/src/write_boot.rs
Johan-Liebert1: > Is there some documentation linter we could be running here to catch stuff like this?

I'm not sure if there is. Will need to search around

Johan-Liebert1: Updated the description. Let me know if it still doesn't feel correct

PR #150, opened 2025-06-26
fsverity: Wrap enable_verity in test code to retry on ETXTBSY

jeckersb:

### DIFF
--- a/crates/composefs/src/fsverity/mod.rs
+++ b/crates/composefs/src/fsverity/mod.rs
@@ -172,6 +172,24 @@ mod tests {
         fd
     }
 
+    /// Wrap the real method inside of the tests and do a little
+    /// sleep/retry dance.  This is known to be race-prone due to the
+    /// way the tests are run in parallel across different runner
+    /// threads, see:
+    /// https://github.com/containers/composefs-rs/issues/106
+    fn enable_verity<H: FsVerityHashValue>(fd: impl AsFd) -> Result<(), EnableVerityError> {
+        for attempt in 1..=3 {
+            match super::enable_verity::<H>(&fd) {
+                Err(EnableVerityError::FileOpenedForWrite) if attempt != 3 => {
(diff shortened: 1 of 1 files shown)


jeckersb:
Somewhat confused why it's getting console-0.15.12 when we have 0.15.11 in `Cargo.lock`

crates/composefs/src/fsverity/mod.rs
```diff
@@ -172,6 +172,24 @@ mod tests {
         fd
     }
 
+    /// Wrap the real method inside of the tests and do a little
```
cgwalters, line 175: I'm still not a big fan of fixing this only in the tests.

crates/composefs/src/fsverity/mod.rs
```diff
@@ -172,6 +172,24 @@ mod tests {
...
+        for attempt in 1..=3 {
+            match super::enable_verity::<H>(&fd) {
+                Err(EnableVerityError::FileOpenedForWrite) if attempt != 3 => {
+                    std::thread::sleep(std::time::Duration::from_millis(1));
+                }
+                other => return other,
+            }
+        }
+
+        unreachable!("We always return on the last iteration above");
```
cgwalters, line 190: We can avoid this `unreachable!` with an explicit `loop`, something like this

```
fn enable_verity<H: FsVerityHashValue>(fd: impl AsFd) -> Result<(), EnableVerityError> {
    let mut attempt = 1;
    loop {
        match super::enable_verity::<H>(&fd) {
            Err(EnableVerityError::FileOpenedForWrite) if attempt < 3 => {
                std::thread::sleep(std::time::Duration::from_millis(1));
                attempt += 1;
            }
            other => return other,
        }
    }
}
```

But anyways I think it wouldn't really be too much more code to do the copy-based fallback right? And make it generic outside of the tests

PR #151, opened 2025-06-27
utils: Make filter_errno a method on the result instead of a function

allisonkarlitskaya:
I.e. instead of `filter(open(), ENOENT)`, you do `open().filter(ENOENT)`.


allisonkarlitskaya (commented):
Good by me, but I can't self-ack.

jeckersb (approved):

PR #152, opened 2025-07-03
fsverity: split `enable_verity` into multiple related functions

jeckersb:
`enable_verity_raw` is effectively the same as the previous `enable_verity`, it simply calls the underlying ioctl.

`enable_verity_with_retry` will attempt to enable verity an opinionated three times with a 1ms delay between tries when encountering ETXTBSY.  This is generally enough to work around the delay with CLOEXEC in conjunction with fork()->exec().

`enable_verity_maybe_copy` will take ownership of the file descriptor and attempt to enable verity using `enable_verity_with_retry`.  If successful, it returns `VerityFd::Orig(fd)` to designate that the file descriptor is the same one provided by the caller.  If unsuccessful, it will instead copy the data to a new O_TMPFILE and return that via `VerityFd::Copy(fd)`.  It is up to the caller to reconcile how this new file descriptor is to be persisted into the filesystem.

Resolves: https://github.com/containers/composefs-rs/issues/106

### DIFF
--- a/crates/composefs/src/fsverity/mod.rs
+++ b/crates/composefs/src/fsverity/mod.rs
@@ -288,7 +288,15 @@ pub fn ensure_verity_equal(
 
 #[cfg(test)]
 mod tests {
-    use std::{collections::BTreeSet, io::Write};
+    use std::{
+        collections::BTreeSet,
+        io::Write,
+        sync::{
+            atomic::{AtomicU64, Ordering},
+            Arc,
+        },
+        time::{self, Duration},
+    };
 
     use rustix::{
         fd::OwnedFd,
@@ -395,6 +403,71 @@ mod tests {
             found,
             "1e2eaa4202d750a41174ee454970b92c1bc2f925b1e35076d8c7d5f56362ba64"
         );
+    }
+
+    #[allow(unsafe_code)]
+    #[tokio::test]
+    async fn test_verity_forking() {
+        const DELAY: Duration = time::Duration::from_millis(10);
+
+        let fork_iterations = Arc::new(AtomicU64::new(0));
+        let fi = fork_iterations.clone();
+        let forker = tokio::task::spawn(async move {
+            loop {
+                fi.fetch_add(1, Ordering::AcqRel);
+                unsafe {
+                    tokio::process::Command::new("true")
+                        .pre_exec(|| {
+                            std::thread::sleep(DELAY);
+                            Ok(())
+                        })
+                        .status()
+                        .await
+                        .unwrap();
+                }
+            }
+        });
+        tokio::pin!(forker);
+        let verity_iterations = Arc::new(AtomicU64::new(0));
+        let vi = verity_iterations.clone();
+        let verity_enabler = tokio::task::spawn(async move {
+            loop {
+                vi.fetch_add(1, Ordering::AcqRel);
+                let r = tokio::task::spawn_blocking(move || {
+                    let tf = rdonly_file_with(b"hello world");
+                    enable_verity::<Sha256HashValue>(&tf)
+                })
+                .await
+                .unwrap();
+                if r.is_ok() {
+                    return;
+                }
+            }
+        });
+        tokio::pin!(verity_enabler);
+        let timer = tokio::time::interval(time::Duration::from_secs(1));
+        tokio::pin!(timer);
+        let ts = tokio::time::Instant::now();
+        loop {
+            tokio::select! {
+                _ = timer.tick() => {
+                    let forks = fork_iterations.load(Ordering::Acquire);
+                    let verity = verity_iterations.load(Ordering::Acquire);
+                    eprintln!("forks={forks} verity={verity}");
+                    continue
+                },
+                _ = forker.as_mut() => {
+                    unreachable!()
+                },
+                _ = verity_enabler.as_mut() => {
+                    let elapsed = ts.elapsed().as_secs();
+                    let forks = fork_iterations.load(Ordering::Acquire);
+                    let verity = verity_iterations.load(Ordering::Acquire);
+                    eprintln!("forks={forks} verity={verity} in {elapsed}s");
+                    break
+                }
+            }
+        }
     }
 
     #[test_with::path(/dev/shm)]
--- a/crates/composefs/src/fsverity/mod.rs
+++ b/crates/composefs/src/fsverity/mod.rs
@@ -288,16 +288,9 @@ pub fn ensure_verity_equal(
 
 #[cfg(test)]
 mod tests {
-    use std::{
-        collections::BTreeSet,
-        io::Write,
-        sync::{
-            atomic::{AtomicU64, Ordering},
-            Arc,
-        },
-        time::{self, Duration},
-    };
+    use std::{collections::BTreeSet, io::Write, os::unix::process::CommandExt, time::Duration};
 
+    use rand::Rng;
     use rustix::{
         fd::OwnedFd,
         fs::{open, Mode, OFlags},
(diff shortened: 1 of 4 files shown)


crates/composefs/src/repository.rs
```diff
@@ -135,7 +135,7 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
                     Err(CompareVerityError::Measure(MeasureVerityError::VerityMissing))
                         if self.insecure =>
                     {
-                        match enable_verity::<ObjectID>(&fd) {
+                        match enable_verity_raw::<ObjectID>(&fd) {
```
jeckersb, line 138: Note I very intentionally left this as the `_raw` variant for now so it's effectively a no-op as far as `Repository` is concerned, but should probably be updated to use one of the new APIs.

jeckersb:
classic case of "works on my machine" for the tests... unsure why we get EPERM here :eyes:

allisonkarlitskaya (commented):
I don't love these changes... they're very invasive and very public.

I think we need to treat this all as a workaround.  My preference would be to add the looping code into Repository, but I'd also be okay if it got added to the fs-verity module, but probably not as a separate API?  It should just "DTRT" when you call the main API...

One thing that's missing here is a large comment discussing the fact that the loop is a workaround and a link to a thread on fsdevel where we reported the issue and asked for a kernel fix.  You discovered the cause of the issue in the first place, but I'm also happy to write that email, if you prefer...

cgwalters:
> I don't love these changes... they're very invasive 

Hmm, invasive how? Lines of code?

> and very public.

As in you don't think the "retry" APIs should be public, but we'd keep the non-retry versions public?

> My preference would be to add the looping code into Repository

Yes, that was my initial thought as well, but then it makes it annoying to have unit tests for the low level fsverity APIs.

> One thing that's missing here is a large comment discussing the fact that the loop is a workaround and a link to a thread on fsdevel where we reported the issue and asked for a kernel fix. 

Fair, though the commit will include a link to the issue.

I guess one way to handle this (and perhaps help with the above other concerns would be to have a new module `fsverity_retry` or something where we put these "mid level retry" APIs and that module could have the doc comment or so?

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +113,116 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+    }
+}
+
+/// Enable fs-verity on a new copy of `fd`, consuming `fd` and
+/// returning the new copy.  The copy is created via O_TMPFILE
+/// relative to `dirfd`.
+fn enable_verity_on_copy<H: FsVerityHashValue>(
+    dirfd: impl AsFd,
+    fd: OwnedFd,
+    mode: Mode,
```
cgwalters, line 195: I think we could default to inheriting the mode of the input fd instead.

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +113,116 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+) -> Result<OwnedFd, EnableVerityError> {
+    let mut fd = File::from(fd);
+
+    loop {
+        fd.rewind().map_err(EnableVerityError::Rewind)?;
```
cgwalters, line 200: This seems unlikely enough to fail that I think we could just map to `EnableVerityError::Io`

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +113,116 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+        let new_ro_fd = open(
+            proc_self_fd(&new_rw_fd),
+            OFlags::RDONLY | OFlags::CLOEXEC,
+            Mode::empty(),
+        )
+        .map_err(|e| EnableVerityError::Open(e.into()))?;
+        drop(new_rw_fd);
+        if enable_verity_with_retry::<H>(&new_ro_fd).is_ok() {
+            return Ok(new_ro_fd);
+        }
```
cgwalters, line 222:
```suggestion
        }
        fd = new_ro_fd;
```

I think we need that to be robust against long running forks.

jeckersb: At this point we've already determined that we have to make a copy (`enable_verity_on_copy` is internal, called only when we have exhausted all other efforts and we are pretty sure there *is* a long running fork, otherwise the `enable_verity_with_retry` variant would have caught it).  The original `fd` doesn't matter other than providing us a source to copy the data from.  In other words, it doesn't matter if a long running fork has a rw copy of it, we're not trying to enable verity on that file.

jeckersb:
Hm apparently the `EPERM` error has magically gone away now?  That's mildly terrifying...

crates/composefs/src/fsverity/mod.rs
```diff
@@ -45,6 +55,33 @@ pub enum CompareVerityError {
...
+    Orig(OwnedFd),
+    Copy(OwnedFd),
+}
+
+impl VerityFd {
+    pub fn is_orig(&self) -> bool {
+        matches!(self, Self::Orig(_))
+    }
+
+    pub fn is_copy(&self) -> bool {
```
cgwalters, line 73: I think we should have docstrings for public APIs. But it's not clear to me these need to be public at all given it's just used in the tests.

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +109,115 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+///
+/// # Arguments:
+/// * `dirfd`: A directory file descriptor, used to determine the placement (via O_TMPFILE) of the new file (if necessary).
+/// * `fd`: The file decriptor to enable verity on
+pub fn enable_verity_maybe_copy<H: FsVerityHashValue>(
+    dirfd: impl AsFd,
+    fd: OwnedFd,
+) -> Result<VerityFd, EnableVerityError> {
+    match enable_verity_with_retry::<H>(&fd) {
+        Ok(_) => Ok(VerityFd::Orig(fd)),
```
cgwalters, line 176: I think a best practice is to spell this `Ok(())` - it looks weirder but ensures that we're only discarding a unit return value.

jeckersb:
> As in you don't think the "retry" APIs should be public, but we'd keep the non-retry versions public?

Yeah that's a reasonable approach.  I made them public just because I figured sooner than later somebody else would discover this same trap and want a way out.  Totally fine to make these private and we can always revisit in the future.  In a perfect world we get a proper fix before anybody notices then the problem goes away :smile:
 
> > My preference would be to add the looping code into Repository
> 
> Yes, that was my initial thought as well, but then it makes it annoying to have unit tests for the low level fsverity APIs.

I diverged from Colin's original proposal here precisely because I thought it made more sense to do this looping nonsense in the fsverity module.  The fact that we have to do this at all is entirely an fsverity problem and not something that Repository should have to care about.  Repository would need to accommodate re-linking things if it uses the `maybe_copy` variant, but I think that's the right place to separate the concerns.  Repository needs a file in a place with fsverity enabled.  It doesn't really care how that happens.

---

@allisonkarlitskaya re: fsdevel, I don't have a strong preference on who emails them.  I don't know anybody in kernel land and have never sent an email to any of the kernel mailing lists.  So on one hand, it's a good opportunity for me to join as a fresh face and get involved, but on the other hand you have earned some amount of trust and equity with those folks so it has more weight coming from you.

cgwalters:
>  Repository would need to accommodate re-linking things if it uses the maybe_copy variant,

Hmm wait would it? How would the behavior need to differ? We link into place *after* enabling verity. So we can drop the `VerityFd` enum and just accept an `OwnedFd` and return one, right?

(In unit tests, we'd want to detect the difference, but my suggestion for doing that is to check the file ctime or so)

cgwalters:
> I'm going to defer updates re: pub API docs until we collectively decide if we want to make any of this public at all.

Well, bootc uses the verity enable APIs today optionally in the ostree paths, and I think it makes sense to switch those over, so the APIs would need to be public.

jeckersb:
On a related note, we also [previously discussed](https://github.com/containers/composefs-rs/discussions/125) moving the fsverity stuff into its own crate since that's all generally useful even without composefs.  Somethinig to keep in mind as we decide where things should ultimately go.

cgwalters:
> (not copying — just the simple "wait and try again" logic)

What problem do you have with the copying API? The rationale here is that a wait/retry is inherently racy.

I was curious so I whipped this up:

```

    #[allow(unsafe_code)]
    #[tokio::test]
    async fn test_verity_forking() {
        const DELAY: Duration = time::Duration::from_millis(10);

        let fork_iterations = Arc::new(AtomicU64::new(0));
        let fi = fork_iterations.clone();
        let forker = tokio::task::spawn(async move {
            loop {
                fi.fetch_add(1, Ordering::Relaxed);
                unsafe {
                    tokio::process::Command::new("true")
                        .pre_exec(|| {
                            std::thread::sleep(DELAY);
                            Ok(())
                        })
                        .status()
                        .await
                        .unwrap();
                }
            }
        });
        tokio::pin!(forker);
        let verity_iterations = Arc::new(AtomicU64::new(0));
        let vi = verity_iterations.clone();
        let verity_enabler = tokio::task::spawn(async move {
            loop {
                vi.fetch_add(1, Ordering::Relaxed);
                let r = tokio::task::spawn_blocking(move || {
                    let tf = rdonly_file_with(b"hello world");
                    enable_verity::<Sha256HashValue>(&tf)
                })
                .await
                .unwrap();
                if r.is_ok() {
                    return;
                }
            }
        });
        tokio::pin!(verity_enabler);
        let timer = tokio::time::interval(time::Duration::from_secs(1));
        tokio::pin!(timer);
        let ts = tokio::time::Instant::now();
        loop {
            tokio::select! {
                _ = timer.tick() => {
                    let forks = fork_iterations.load(Ordering::Relaxed);
                    let verity = verity_iterations.load(Ordering::Relaxed);
                    eprintln!("forks={forks} verity={verity}");
                    continue
                },
                _ = forker.as_mut() => {
                    unreachable!()
                },
                _ = verity_enabler.as_mut() => {
                    let elapsed = ts.elapsed().as_secs();
                    let forks = fork_iterations.load(Ordering::Relaxed);
                    let verity = verity_iterations.load(Ordering::Relaxed);
                    eprintln!("forks={forks} verity={verity} in {elapsed}s");
                    break
                }
            }
        }
    }
```

At a `DELAY` of 0, things reliably complete for me in a second, with about 10 retries. But going to a delay of `time::Duration::from_millis(10);`, things spiral *very* fast into really human-visible delays:

```
---- fsverity::tests::test_verity_forking stdout ----
forks=1 verity=1
temporary directory from ~/.var/tmp: "/var/home/walters/.var/tmp"
forks=82 verity=82
forks=162 verity=162
forks=241 verity=241
forks=321 verity=321
forks=402 verity=402
forks=482 verity=482
forks=561 verity=561
forks=641 verity=641
forks=680 verity=679 in 8s
```

One time it completed in 0s, another time I've been waiting over a minute and it still isn't done.

The thing that's good about a copy fallback is while it's still racy, it's *insensitive to the delay between fork/exec* in general (but especially where that delay = infinity for a long-running fork).

jeckersb:
Still fiddling with this, but a few observations with your patch:

- It's not nearly as bad on my system as what you're seeing, I suspect your system is just so much better that it's more likely to hit the timing than my slow laptop.
- The `std::thread::sleep` call in the pre_exec hook is blocking up the eventloop, which makes even the successful case take longer than it should (it seems to always take a little more than DELAY*2) for me.  I rewrote that bit to use `spawn_blocking` and just use a `std::process::Command` instead.  I'm experimenting now with using the `biased` option with `tokio::select!` to somewhat force the ordering (I want the forker to generally start/poll first). 
- This test in parallel with the rest of the test suite *really* shows this problem, because it almost always blows up all of the other verity tests :laughing:

cgwalters:
> The std::thread::sleep call in the pre_exec hook is blocking up the eventloop,

Hmm wait how can that be? `pre_exec` should be running after `fork()` in a new process right?

crates/composefs/src/fsverity/mod.rs
```diff
@@ -45,6 +55,33 @@ pub enum CompareVerityError {
...
 }
 
+/// An owned file descriptor with fsverity enabled.  Used in contexts
+/// where a user-supplied file descriptor may be returned back to the
+/// user directly (`Orig`), or a distinct copy of the file descriptor
+/// (`Copy`) may be returned in place of the original.
+#[derive(Debug)]
+pub enum VerityFd {
+    Orig(OwnedFd),
+    Copy(OwnedFd),
```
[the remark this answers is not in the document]
cgwalters: We can still drop this one right?

cgwalters:
I think if we drop `VerityFd` (it's not needed) and also squash the commits this should overall be good to go from my PoV

jeckersb:
> > But I need to know whether it's the same file or a new copy so I can rearrange things accordingly.
> 
> OK. I verified that this type checks at least
> 
> ```
> #[test]
> fn t() {
>     let v: std::borrow::Cow<'_, BorrowedFd> = std::borrow::Cow::Owned(std::io::stdin().as_fd().clone().into());
> }
> ```

That goes back to a [comment](https://github.com/containers/composefs-rs/issues/106#issuecomment-3009928629) I made on the issue though.  I'd like the API to enforce the caller giving up ownership so it's not possible to keep using the original file in the case where it gets copied.  If we use `Cow` we'd always just give back a `Cow::Owned` but the caller has no (easy) way to determine if a copy was made or not.

jeckersb:
Ok I added your patch, rebased, removed the CI debug commit, cleaned up a WIP commit msg.  Still need to go back and do a few more cleanups (API docs mostly).

jeckersb:
Added a commit to hide the fork testing behind a feature.  I don't love it, but I also didn't really want to force every test to use the `maybe_copy` API.

jeckersb:
Hm EPERM again...

cgwalters:
IOW let's change every caller in the codebase to not use `enable_verity_raw` including repository.rs especially

jeckersb:
> > But I need to know whether it's the same file or a new copy so I can rearrange things accordingly.
> 
> OK. I verified that this type checks at least
> 
> ```
> #[test]
> fn t() {
>     let v: std::borrow::Cow<'_, BorrowedFd> = std::borrow::Cow::Owned(std::io::stdin().as_fd().clone().into());
> }
> ```

I am coming around to this idea, mainly because in `Repository` there are circumstances where we try to enable verity, but if the `insecure` option is set we ignore it and link it anyway.  But with the API requiring an `OwnedFd` we end up losing the fd in the process of returning `Err(EnableVerityError::FilesystemNotSupported)`.

cgwalters:
I should have looked more carefully, see [the docs](https://doc.rust-lang.org/std/os/fd/struct.BorrowedFd.html):

> This type does not have a [ToOwned](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) implementation. Calling .to_owned() on a variable of this type will call it on &BorrowedFd and use Clone::clone() like ToOwned does for all types implementing Clone. The result will be descriptor borrowed under the same lifetime.

>  But with the API requiring an OwnedFd we end up losing the fd in the process of returning Err(EnableVerityError::FilesystemNotSupported).

Ah yeah. Only borrowing the input fd does seem by far simpler here. In the end really it's not like we have 1000 use cases of this that need to be right, it should only basically boil down to the one caller in repository getting it right.

jeckersb:
So what I've landed on here is we take a `BorrowedFd`, and return `Option<OwnedFd>`.  It's `Some(fd)` if it makes a copy, and `None` if it used the original.  I like this the most of anything so far.

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +82,120 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+/// enable verity three times, pausing for one millisecond between
+/// attempts.
+pub fn enable_verity_with_retry<H: FsVerityHashValue>(
+    fd: impl AsFd,
+) -> Result<(), EnableVerityError> {
+    let mut attempt = 1;
+    loop {
+        match enable_verity_raw::<H>(&fd) {
+            Err(EnableVerityError::FileOpenedForWrite) if attempt < 3 => {
+                std::thread::sleep(std::time::Duration::from_millis(1));
```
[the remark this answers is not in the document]
jeckersb: That was 100% just finger-in-the-wind making up numbers.  Absolutely nothing backing that up, other than it seemed enough to "fix" the problem (until we added the pathological forker).

crates/composefs/src/fsverity/mod.rs
```diff
@@ -72,10 +82,120 @@ pub fn compute_verity<H: FsVerityHashValue>(data: &[u8]) -> H {
...
+            let fd = enable_verity_on_copy::<H>(dirfd, fd)?;
+            Ok(Some(fd))
+        }
+        Err(other) => Err(other),
+    }
+}
+
+/// Enable fs-verity on a new copy of `fd`, consuming `fd` and
+/// returning the new copy.  The copy is created via O_TMPFILE
+/// relative to `dirfd`.
```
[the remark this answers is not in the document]
cgwalters: I think if she wants to additionally review again that's obviously fine, but I'd like to cross this one off the board.

PR #153, opened 2025-07-09
features: drop pre-6.15 feature by default

allisonkarlitskaya:
Now that 6.15 is widely available (Arch, Fedora stable) we can drop this compatibility code by default.  We leave it on for Ubuntu.

We also migrate our compatibility test case to Fedora stable now that it has the new kernel, and enable it in CI.


allisonkarlitskaya:
The Ubuntu fail here is nothing new: it's the mount-related regressions in 6.15 that have been fixed in more-recent versions.

cgwalters:
Thanks yes, I adjusted the bootc PR.

PR #155, opened 2025-07-12
mountcompat: Handle 32 vs 64 bit ioctls

cgwalters:
See all the conditionals under
https://github.com/bytecodealliance/rustix/blob/94a0c5170b16ff3c43831ca9c44eca37cd3090e3/src/ioctl/mod.rs#L291

Basically on s390x and ppc64le we end up using glibc by default which uses u64 by default.


allisonkarlitskaya (approved):
Right!  I had to fix this in the fs-verity code as well.  Thanks for catching those!

PR #156, opened 2025-07-15
Port tests to `test.thing`

allisonkarlitskaya:
...probably still broken, but let's see how it goes.

---

Edited by @cgwalters 

Link: https://codeberg.org/lis/test.thing


cgwalters:
Just skimming test.thing looks quite cool but definitely overlaps with several other efforts/projects this space.

allisonkarlitskaya:
> Just skimming test.thing looks quite cool but definitely overlaps with several other efforts/projects this space.

This work is really being done mostly for the benefit of Cockpit.  I've looked into some of the other solutions (`mkosi qemu` was the closest fit) but after a chat with Daan at devconf and offline after, our usecase was different enough that I decided to make a try at my own thing...

The sheer amount of enablement work that I've been having to do in the distros is a bit of a hint to me that this approach is fairly new...

allisonkarlitskaya:
> Just skimming test.thing looks quite cool but definitely overlaps with several other efforts/projects this space.

Did you have a suggestion for a comparable project I should be looking into?

allisonkarlitskaya:
Facts:
 - the Ubuntu image (and only the Ubuntu image) is crashing
 - the crash happens very early during kernel startup before any console output appears.  I've added timeout support and console logging to `test.thing` to confirm that: the only thing you see is `systemd-boot` telling us that it'll boot the image in 1s.  I've also confirmed that during manual testing.
 - I've never seen the crash locally
 - there is definitely some random element to the crash.  I've seen the exact same commit ID work (ie: fail with the expected failure that we've been seeing on Ubuntu for ages) and then not work in a separate run:
   - attempt 1 (working): https://github.com/containers/composefs-rs/actions/runs/16357795305/job/46219910738?pr=156
   - attempt 2 (failing): https://github.com/containers/composefs-rs/actions/runs/16357795305/job/46220342374?pr=156

Observations:
 - the crash seems to go away when disabling ephemeral ssh key generation.  This happens regardless of if the cockpit-test key is still in the image.  ie: ephemeral ssh keys with a static key in the image will crash, but not generating the keys avoids the crash.  I have a feeling this is related to the smbios strings we send in for the ephemeral key.
 - manual inspection also seems to change things.  The crash is potentially timing-sensitive.  See the above comment about "definitely some random element"
 - on a hunch that qemu was producing bad smbios tables that caused a crash in the early kernel DMI parsing, I upgraded the qemu installed on the runners to the version from plucky, but it didn't fix the problem.  If qemu is involved here and the reason that I've never seen it locally is because my local version was fixed, then the fix happened between plucky and what I'm using locally.  That's a lot of "if".

allisonkarlitskaya:
So it seems to be getting stuck in the `security/integrity/ima` code at this point which does a lot of platformy kinda stuff with measurements and TPM and such... that passes a gut-check, I guess...

allisonkarlitskaya:
The globalness of vsock on the host bugs me as well, but for a different reason: allocating a unique CID is seriously annoying.  I'm waiting for qemu to add unix socket support for that like firecracker does: https://gitlab.com/qemu-project/qemu/-/issues/2095 and would definitely switch to that when it exists.

The mitigation we take for this is the use of ephemeral ssh keys: you can't get in unless you have that key.  We could also write our own virtio backend for the host side of vsock, and there even exists a Rust library for doing that, but it's decidedly non-trivial.

Also: this is testing-oriented... the status quo here was a globally-accessible TCP port forward with a well-known private key...

Thanks very much for the issue on `test.thing`.  I'll look into that.  I love feedback like that.

cgwalters:
> Also: this is testing-oriented... the status quo here was a globally-accessible TCP port forward with a well-known private key...

Network namespaces exist, so it doesn't have to be global in that sense. But really it's not just global to the *host* it's global to *all other virtual machines* which is just a whole other ball of wax.

PR #157, opened 2025-07-17
Initial cleanups for #156

allisonkarlitskaya:
I'm having trouble getting #156 across the line, so let's see if CI works on the first few commits here...


allisonkarlitskaya:
I ended up just doing this all in #156.  Without the Ubuntu image we can get that one merged now.

PR #160, opened 2025-07-18
tar: Strip trailing null characters from filenames

Johan-Liebert1:
Some file paths end up with trailing null characters which causes issues while building the filesystem tree.

Ex. `/usr/../grpc\0` adds `grpc\0` as the directory name. When it comes to directory `/usr/../grpc/otelgrpc\0` we get the error "Directory 'grpc' not found"

### DIFF
--- a/crates/composefs-oci/src/tar.rs
+++ b/crates/composefs-oci/src/tar.rs
@@ -147,6 +147,10 @@ fn path_from_tar(pax: Option<Box<[u8]>>, gnu: Vec<u8>, short: &[u8]) -> PathBuf 
         path.pop(); // this is Vec<u8>, so that's a single char.
     }
 
+    if path.last() == Some(&b'\0') {
+        path.pop();
+    }
+
     PathBuf::from(OsString::from_vec(path))
 }
 


cgwalters:
This smells like a bug in the erofs reading code to me.

allisonkarlitskaya:
> This smells like a bug in the erofs reading code to me.

The only thing we use erofs reading for is during GC operations, and we don't even look at the filenames — only the xattrs to collect a list of referred objects.

This is in the tar code: I'm kinda curious about why we have tar files with filenames with `/../` in them.  Did you mean that literally or was that an ellipsis?  Note: symlink target handling is in a separate function, so it's not that.

At the very least I'd like to know which path this was on.  Just up from where your patch is there's three different sources that the filename might have come from: the traditional header, the PAX extensions, or the GNU extensions.  Can you find out which one of those is actually getting the `\0` into the filename and trace it back?  It's probably a bug in the handling of where the data comes from...

Do you have an example image that's exhibiting this behaviour that I could look at?

Johan-Liebert1:
Debugging it further, it seems to be part of the tar header?

Adding some print statements to https://github.com/containers/composefs-rs/blob/main/crates/composefs-oci/src/tar.rs#L76 function, I get the following result

```
GnuHeader {
    entry_size: 107,
    size: 107,
    path: "././@LongLink",
    link_name: None,
    mode: 0o644,
    uid: 0,
    gid: 0,
    mtime: 0,
    username: Some(
        "",
    ),
    groupname: Some(
        "",
    ),
    cksum: 4090,
    cksum_valid: true,
    is_extended: false,
    sparse: [],
}

Buffer: [46, 47, 117, 115, 114, 47, 115, 104, 97, 114, 101, 47, 108, 105, 99, 101, 110, 115, 101, 115, 47, 109, 111, 98, 121, 45, 101, 110, 103, 105, 110, 101, 47, 118, 101, 110, 100, 111, 114, 47, 103, 111, 46, 111, 112, 101, 110, 116, 101, 108, 101, 109, 101, 116, 114, 121, 46, 105, 111, 47, 99, 111, 110, 116, 114, 105, 98, 47, 105, 110, 115, 116, 114, 117, 109, 101, 110, 116, 97, 116, 105, 111, 110, 47, 103, 111, 111, 103, 108, 101, 46, 103, 111, 108, 97, 110, 103, 46, 111, 114, 103, 47, 103, 114, 112, 99, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

The path is 

```
./usr/share/licenses/moby-engine/vendor/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc
```
which has a length of 106, but the header mentions a length of 107 hence the final null byte gets included.

I am not sure if this is reproducible, but here is the Containerfile I used

```dockerfile
FROM quay.io/fedora/fedora-coreos:stable
RUN dnf install -y setroubleshoot setroubleshoot-server
RUN passwd -d root
```

Johan-Liebert1:
Added a note, and created an issue https://github.com/containers/composefs-rs/issues/162

cgwalters (approved):

crates/composefs-oci/src/tar.rs
```diff
@@ -147,6 +147,13 @@ fn path_from_tar(pax: Option<Box<[u8]>>, gnu: Vec<u8>, short: &[u8]) -> PathBuf
         path.pop(); // this is Vec<u8>, so that's a single char.
     }
 
+    // NOTE: We use a custom tar parser since splitstreams are not actual tar archives
+    // The `tar` crate does have a higher level `path` function that would do this for us.
+    // See: https://github.com/alexcrichton/tar-rs/blob/a1c3036af48fa02437909112239f0632e4cfcfae/src/header.rs#L1532
+    if path.last() == Some(&b'\0') {
```
allisonkarlitskaya, line 153: I'd still really strongly prefer to have this on the path that it actually pops up (ie: the GNU one).  It should ideally happen before we even collect the name into the variable: that's where the bug is.

allisonkarlitskaya (changes requested):

crates/composefs-oci/src/tar.rs
Johan-Liebert1: Updated

cgwalters:
OK I did https://github.com/containers/composefs-rs/pull/166 - if you rebase this PR on that, then this PR can drop the `[ignore]`

Johan-Liebert1:
Tests in #166 pass pass with these changes. This also happens with PAX extensions though. Tested with the following snippet

```rust
    #[test]
    fn test_pax_long_filename_reproduction() {
        // Create a very long path that will definitely trigger GNU long name extensions
        let very_long_path = format!(
            "very/long/path/that/exceeds/the/normal/tar/header/limit/{}",
            "x".repeat(120)
        );
        let content = b"Content for very long path";

        // Use append_data to create a tar with a very long filename that triggers GNU extensions
        let mut tar_data = Vec::new();
        {
            let mut builder = Builder::new(&mut tar_data);
            append_file(&mut builder, &very_long_path, content).unwrap();
            builder.append_pax_extensions(vec![("path", very_long_path.as_bytes())]).unwrap();
            builder.finish().unwrap();
        };

        let entries = read_all_via_splitstream(tar_data).unwrap();
        assert_eq!(entries.len(), 1);
        let abspath = format!("/{very_long_path}");
        assert_eq!(entries[0].path, Path::new(&abspath));
    }
```

Seeing how this is the case, feels like stripping off the NULL byte in https://github.com/containers/composefs-rs/blob/main/crates/composefs-oci/src/tar.rs#L132 makes more sense.

cgwalters:
> This also happens with PAX extensions though.

Ah OK. BTW in this vein I created https://github.com/cgwalters/tarsmith recently (for a different problem, xref https://github.com/containers/buildah/issues/6297#issuecomment-3129822907 ) - I've been poking at polishing it up more and we could probably add it as a `dev-dependency` at some point to help fuzz splitstream-tar.

> Seeing how this is the case, feels like stripping off the NULL byte in https://github.com/containers/composefs-rs/blob/main/crates/composefs-oci/src/tar.rs#L132 makes more sense.

Hmm, I honestly don't have a really strong opinion on this but since I guess Allison feels more strongly we could just apply the same trimming in the PAX case for now? How about rebasing on git main, adding your test, dropping the `ignore` on the other one and doing that?

Johan-Liebert1:
While retesting, this case does not happen with Pax Extensions. I think the last time I tested it, I had the GNU fix commented out and that's why it failed.

cgwalters:
Looks like we need to either bump MSRV or back out using `pop_if()`. As a baseline we need to build in what's in CentOS Stream 9...which right now is 1.88 so let's bump that MSRV.

Johan-Liebert1:
Bumped MSRV to 1.88.0

cgwalters:
LGTM but we still need @allisonkarlitskaya to clear the "requested changes" flag

cgwalters:
> If we're bumping MSRV do we also want to move to 2024 edition?

No objections from me but surely it should be a separate PR right?

(My previous experience here is edition 2024 is a relatively minor improvement, but the biggest thing is folks should *not* just blindly run `cargo fix --edition` ref https://github.com/bootc-dev/bootc/pull/1130#discussion_r1964182260 )

PR #163, opened 2025-07-24
tests/verity-forking: Limit execution time to 10s

cgwalters:
This test was running over 60s on my machine (32 cores); digging in it's because the raw variant was simply not getting anywhere close to hitting 100 successes.

First, rework to use a JoinSet instead of a tokio::join! so we see the completions as they come in instead of waiting for all of them.

Add a hard limit of execution time of 10s.

Add an assertion that the *copy* variant succeeds at least once. For the other two, we can't really make any assertions at all.


Johan-Liebert1:
LGTM

Johan-Liebert1 (approved):

PR #165, opened 2025-07-31
oci/pull: Allow passing in custom ImageProxyConfig

Johan-Liebert1:
Take `img_proxy_config` as an argument to `composefs_oci::pull` enabling callers to override default proxy configuration.

If a custom `skopeo` command is not passed in the config, we add our own.

### DIFF
--- a/crates/composefs-oci/src/skopeo.rs
+++ b/crates/composefs-oci/src/skopeo.rs
@@ -24,7 +24,11 @@ struct ImageOp<ObjectID: FsVerityHashValue> {
 }
 
 impl<ObjectID: FsVerityHashValue> ImageOp<ObjectID> {
-    async fn new(repo: &Arc<Repository<ObjectID>>, imgref: &str) -> Result<Self> {
+    async fn new(
+        repo: &Arc<Repository<ObjectID>>,
+        imgref: &str,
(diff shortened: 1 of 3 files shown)


crates/composefs-oci/src/skopeo.rs
```diff
@@ -24,7 +24,11 @@ struct ImageOp<ObjectID: FsVerityHashValue> {
...
 
 impl<ObjectID: FsVerityHashValue> ImageOp<ObjectID> {
-    async fn new(repo: &Arc<Repository<ObjectID>>, imgref: &str) -> Result<Self> {
+    async fn new(
+        repo: &Arc<Repository<ObjectID>>,
+        imgref: &str,
+        img_proxy_config: Option<ImageProxyConfig>,
+    ) -> Result<Self> {
         // See https://github.com/containers/skopeo/issues/2563
         let skopeo_cmd = if imgref.starts_with("containers-storage:") && !geteuid().is_root() {
```
cgwalters, line 33: I think this is OK as is but it'd be better to split this bit out as a separate opt-in for the caller. That's how it's done in https://github.com/bootc-dev/bootc/blob/6c3bba52308a26a83e934f0f3943ed655954989d/crates/ostree-ext/src/container/mod.rs#L421

PR #167, opened 2025-08-04
Split off `mod generic_tree`

cgwalters:
I copied this code into https://github.com/cgwalters/tarsmith/blob/main/src/tree.rs because I wanted a virtual FS but that wasn't specific to fsverity.

This adds a `mod generic_tree` which abstracts over that - then the existing `mod tree` just plugs in the core composefs strategy of storing regular files as an enum that can be either inline or external.


crates/composefs/src/tree.rs
```diff
@@ -1,471 +1,32 @@
...
-        if !self.have_root_stat {
-            self.root.stat.st_mtim_sec = self.root.newest_file();
-            self.have_root_stat = true;
-        }
-    }
-}
+// Re-export generic types. Note that we don't need to re-write
+// the generic constraint T: FsVerityHashValue here because it will
+// be transitively enforced.
+pub type LeafContent<T> = generic_tree::LeafContent<RegularFile<T>>;
```
cgwalters, line 18: I think it's pretty cool how this avoided the churn in the rest of the codebase.

crates/composefs/src/tree.rs
```diff
@@ -1,471 +1,32 @@
...
+pub type Leaf<T> = generic_tree::Leaf<RegularFile<T>>;
+pub type Directory<T> = generic_tree::Directory<RegularFile<T>>;
+pub type Inode<T> = generic_tree::Inode<RegularFile<T>>;
+pub type FileSystem<T> = generic_tree::FileSystem<RegularFile<T>>;
 
 #[cfg(test)]
 mod tests {
```
cgwalters, line 25: In theory we actually don't *really* need any tests at all here I think; we get coverage from the generic tests and higher level tests. But I kept a few just to prove things out.

crates/composefs/src/tree.rs
jeckersb: :crab: :tada:

jeckersb (approved):
LGTM

Issue #164, opened 2025-07-29
erofs: Incorrect selinux labels


Johan-Liebert1:
The generated EROFS image does not have correct SELinux labels for certain directories and files. These directories/files end up getting the `default_t` label. 

One example, among many, is `/home` which is a symlink to `/var/home`

Here are the labels on the mounted EROFS image

```bash
total 88K
drwxr-xr-x.  12 root root system_u:object_r:root_t:s0               358 Jan  1  1970 .
drwx------.   4 core core unconfined_u:object_r:user_home_dir_t:s0  192 Jul 29 10:52 ..
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              7 Jan  1  1970 bin -> usr/bin
drwxr-xr-x.   2 root root system_u:object_r:boot_t:s0                27 Jan  1  1970 boot
drwxr-xr-x.   2 root root system_u:object_r:device_t:s0              27 Jan  1  1970 dev
drwxr-xr-x. 101 root root system_u:object_r:etc_t:s0               4.0K Jul 24 04:46 etc
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              8 Jan  1  1970 home -> var/home
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              7 Jan  1  1970 lib -> usr/lib
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              9 Jan  1  1970 lib64 -> usr/lib64
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              9 Jan  1  1970 media -> run/media
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              7 Jan  1  1970 mnt -> var/mnt
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              7 Jan  1  1970 opt -> var/opt
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0             14 Jan  1  1970 ostree -> sysroot/ostree
drwxr-xr-x.   2 root root system_u:object_r:default_t:s0             27 Jan  1  1970 proc
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0             12 Jan  1  1970 root -> var/roothome
drwxr-xr-x.   5 root root system_u:object_r:var_run_t:s0             87 Jul 24 04:46 run
lrwxrwxrwx.   2 root root system_u:object_r:default_t:s0              8 Jan  1  1970 sbin -> usr/sbin
[...]

PR #169, opened 2025-08-04
boot: Empty /sysroot too

cgwalters:
This aids our compatibility with existing ostree-containers.

Closes: https://github.com/containers/composefs-rs/issues/164

### DIFF
--- a/crates/composefs-boot/src/lib.rs
+++ b/crates/composefs-boot/src/lib.rs
@@ -13,6 +13,24 @@ use composefs::{fsverity::FsVerityHashValue, repository::Repository, tree::FileS
 
 use crate::bootloader::{get_boot_resources, BootEntry};
 
+/// These directories are required to exist in images.
+/// They may have content in the container, but we don't
+/// want to expose them in the final merged root.
+///
+/// # /boot
+///
+/// This is how sealed UKIs are handled; the UKI in /boot has the composefs
+/// digest, so we can't include it in the rendered image.
+///
+/// # /sysroot
+///
+/// See https://github.com/containers/composefs-rs/issues/164
+/// Basically there is only content here in ostree-container cases,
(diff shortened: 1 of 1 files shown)


jeckersb (approved):
Needs DCO but otherwise :+1:

allisonkarlitskaya (commented):
Thanks for the change, and the comments.  I more or less agree.  Two (related notes):
 - how intentional is the added `continue`?  Should we not create `/boot` if it doesn't exist?  Or is that an error?  How about `/sysroot`?
 - if we do decide to always create/sysroot then we can simplify the examples to skip that step

cgwalters:
And then it makes sense in the code to have `/boot` be mandatory too.

That said, there *is* I think a rationale for having that be configurable; the spec argues for mounting the ESP at `/efi`, not `/boot/efi`.

allisonkarlitskaya:
Last I heard you're supposed to put the ESP at `/boot` unless there's also an xloader or whatever partition.  So `/boot` is always mandatory, imho.

I agree that `/sysroot` could be optional and we do indeed go out of our way to support not mounting it if the mountpoint is missing.  On the other hand, `Repository` (and `cfsctl`) assume that `/sysroot/composefs/` is the location of the "system" repository.  Moving that around is just causing ourselves grief for no good reason.

If we did want another (fixed) spot for this under `/run` I'd consider that to be interesting, but it would complicate things: the distro expects to mount its own tmpfs for `/run` (or move the one from the initramfs environment?) after (or before?) `/sbin/init` is started... the fact that I don't know the answer to either of those questions is an indication that we might not want to rely on this...

Also: everyone knows what `/sysroot` is.  Maybe the ship has sailed on moving this one.

But: if someone wants it to be optional to avoid having that mount by default, I think that's legit...

PR #174, opened 2025-08-13
Fix warnings for mismatched_lifetime_syntaxes lint

jeckersb:
Rust 1.89 introduced this new lint[1], fix the handful of places where
it comes up across the repo.

[1] https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#mismatched-lifetime-syntaxes-lint


allisonkarlitskaya (approved):
I read a bit about this new rule and thought it seemed vaguely useful but in practice it's kinda annoying... this is what elided lifetimes is supposed to be about.... I guess I kinda see the point though that it's not obvious that the lifetime is captured here.

Thanks for taking care of this.

PR #182, opened 2025-09-22
uki: Make function `get_text_section` public

Johan-Liebert1:
Useful in bootc for getting version/version id and maybe more in the future from the UKI


jeckersb (approved):
LGTM, with the caveat that the comment above should be changed to a proper docstring and rewritten since it's going to public.  But not a blocker, since large chunks of the codebase need updated for docs anyways.

cgwalters:
Yep you could say that section needs more text

Issue #181, opened 2025-09-18
repo metadata


cgwalters:
Right now one can't even know (AFAIK) if a repo is sha256 or sha512 immediately - one would need to look at filenames or fsverity digests. One idea I had here is to have a "stamp/metadata file" which canonically signals the format.

jeckersb:
Yeah it's basically impossible to get that right in our tools because there's no way to know ahead of time whether any given karg is a valid kernel arg or not, short of generating it from the big list at https://docs.kernel.org/admin-guide/kernel-parameters.html#kernel-build-options or similar.

Issue #14, opened 2024-10-24
a way to verify body sha256 of small files on SplitStreamReader


allisonkarlitskaya:
We currently read the container config JSON without any sort of validation of the content.  We can know the fsverity hash of the stream from when we create it, but assuming we don't make note of that, then when we go to open the config via its sha256 ID, we perform no checking at all.

We could fix this by adding an optional sha256 verification layer to the reader similar to the one we have in the writer.  We could also read the entire (small) json document into a vec and checksum it there.  We could also try to think about a reasonable way to make use of fsverity, but this seems to cut a little bit against how containers work today...

Solving this is important to the "mount verified container rootfs" goal.

allisonkarlitskaya:
This kinda-sorta got slightly addressed in 9c5fde531676bdb6f72d753b5a86dad012c17d78 with `Repository::check_stream()` and `oci::open_config()`.  It's not part of the reader API, but it's a start.

cgwalters:
I'm reusing this issue that's about splitstream to ask something tangentially related but much bigger picture.

I don't quite understand the "centrality" of splitstream for this project. I think what we should have as the "root" is composefs metadata blobs that we both read and write. In particular, the `compose_filesystem` function I think should be parsing `composefs-info dump`, not splitstreams.

In an ideal world, we have "canonical tar" and for "verified OCI images" then *there is no splitstream* - we can reliably reproduce a canonical tar from a composefs.

cgwalters:
(The reason I bring this up is I was just replying something similar in https://github.com/containers/storage/pull/2155#discussion_r1831332889 )

allisonkarlitskaya:
We talked about this before, but got caught up on two issues:
 - we don't live in a world where all of the existing container engine implementations have produced canonical tar files, and we need to continue to support those images, including being able to re-push them.  We need something like tar-split, but with the data chunks stored in the composefs object storage.  That's essentially what splitstream is — it was designed for precisely this purpose.
 - when creating the composefs-per-layer there was the question of how we should handle whiteouts:
   - do we leave them as is and write a literal `.wh.` file to the composefs; or
   - do we do the thing that overlayfs would expect and convert `.wh.` files to `0:0` chardevs and `.wh..wh.opq` into the opaque directory xattr?

cgwalters:
> we don't live in a world where all of the existing container engine implementations have produced canonical tar files, and we need to continue to support those images, including being able to re-push them. 

[...]

PR #185, opened 2025-09-29
Preparatory splitstream format changes for ostree support

alexlarsson:
This changes the splitstream format a bit, with the goal of allowing splitstreams to support ostree files as well (see #144), but it it imho a generally nice change.

The primary differences are:

 * The header is not compressed
 * All referenced fs-verity objects are stored in the header, including external chunks, mapped splitstreams and (a new feature) references that are not used in chunks.
 * The mapping table is separate from the reference table (and generally smaller), and indexes into it.
 * There is a magic value to detect the file format.
 * There is a magic content type to detect the type wrapped in the stream.
 * We store a tag for what ObjectID format is used
 * The total size of the stream is stored in the header.

The ability to reference file objects in the repo even if they are not part of the splitstream "content" will be useful for the ostree support to reference file content objects.

This change also allows more efficient GC enumeration, because we don't have to parse the entire splitstream to find the referenced objects.

### DIFF
--- a/crates/composefs/src/repository.rs
+++ b/crates/composefs/src/repository.rs
@@ -18,8 +18,8 @@ use anyhow::{bail, ensure, Context, Result};
 use once_cell::sync::OnceCell;
 use rustix::{
     fs::{
-        fdatasync, flock, linkat, mkdirat, open, openat, readlinkat, statat, AtFlags, Dir,
-        FileType, FlockOperation, Mode, OFlags, CWD,
+        flock, linkat, mkdirat, open, openat, readlinkat, statat, syncfs, AtFlags, Dir, FileType,
+        FlockOperation, Mode, OFlags, CWD,
     },
     io::{Errno, Result as ErrnoResult},
 };
@@ -128,12 +128,18 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
     ///
     /// Same as `ensure_object` but runs the operation on a blocking thread pool
     /// to avoid blocking async tasks. Returns the fsverity digest of the object.
+    ///
+    /// For performance reasons, this function does *not* call fsync() or similar.  After you're
+    /// done with everything, call `Repository::sync_async()`.
     pub async fn ensure_object_async(self: &Arc<Self>, data: Vec<u8>) -> Result<ObjectID> {
         let self_ = Arc::clone(self);
         tokio::task::spawn_blocking(move || self_.ensure_object(&data)).await?
     }
 
     /// Given a blob of data, store it in the repository.
+    ///
+    /// For performance reasons, this function does *not* call fsync() or similar.  After you're
+    /// done with everything, call `Repository::sync()`.
     pub fn ensure_object(&self, data: &[u8]) -> Result<ObjectID> {
         let dirfd = self.objects_dir()?;
         let id: ObjectID = compute_verity(data);
@@ -179,14 +185,15 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         let fd = ensure_dir_and_openat(dirfd, &id.to_object_dir(), OFlags::RDWR | OFlags::TMPFILE)?;
         let mut file = File::from(fd);
         file.write_all(data)?;
-        fdatasync(&file)?;
-
         // We can't enable verity with an open writable fd, so re-open and close the old one.
         let ro_fd = open(
             proc_self_fd(&file),
             OFlags::RDONLY | OFlags::CLOEXEC,
             Mode::empty(),
         )?;
+        // NB: We should do fdatasync() or fsync() here, but doing this for each file forces the
+        // creation of a massive number of journal commits and is a performance disaster.  We need
+        // to coordinate this at a higher level.  See .write_stream().
         drop(file);
 
         let ro_fd = match enable_verity_maybe_copy::<ObjectID>(dirfd, ro_fd.as_fd()) {
@@ -279,16 +286,36 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         }
     }
 
-    /// Write the given splitstream to the repository with the
-    /// provided content identifier and optional reference name.
+    /// Write the given splitstream to the repository with the provided content identifier and
+    /// optional reference name.
+    ///
+    /// This call contains an internal barrier that guarantees that, in event of a crash, either:
+    ///  - the named stream (by `content_identifier`) will not be available; or
+    ///  - the stream and all of its linked data will be available
+    ///
+    /// In other words: it will not be possible to boot a system which contained a stream named
+    /// `content_identifier` but is missing linked streams or objects from that stream.
     pub fn write_stream(
         &self,
         writer: SplitStreamWriter<ObjectID>,
         content_identifier: &str,
         reference: Option<&str>,
     ) -> Result<ObjectID> {
-        let stream_path = Self::format_stream_path(content_identifier);
         let object_id = writer.done()?;
+
+        // Right now we have:
+        //   - all of the linked external objects and streams; and
+        //   - the binary data of this splitstream itself
+        //
+        // in the filesystem but but not yet guaranteed to be synced to disk.  This is OK because
+        // nobody knows that the binary data of the splitstream is a splitstream yet: it could just
+        // as well be a random data file contained in an OS image or something.
+        //
+        // We need to make sure that all of that makes it to the disk before the splitstream is
+        // visible as a splitstream.
+        self.sync()?;
+
+        let stream_path = Self::format_stream_path(content_identifier);
         let object_path = Self::format_object_path(&object_id);
         self.symlink(&stream_path, &object_path)?;
 
@@ -623,6 +650,24 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         let mut data = vec![];
         std::fs::File::from(image).read_to_end(&mut data)?;
         Ok(crate::erofs::reader::collect_objects(&data)?)
+    }
+
+    /// Makes sure all content is written to the repository.
+    ///
+    /// This is currently just syncfs() on the repository's root directory because we don't have
+    /// any better options at present.  This blocks until the data is written out.
+    pub fn sync(&self) -> Result<()> {
+        syncfs(&self.repository)?;
+        Ok(())
+    }
+
+    /// Makes sure all content is written to the repository.
+    ///
+    /// This is currently just syncfs() on the repository's root directory because we don't have
+    /// any better options at present.  This won't return until the data is written out.
+    pub async fn sync_async(self: &Arc<Self>) -> Result<()> {
+        let self_ = Arc::clone(self);
+        tokio::task::spawn_blocking(move || self_.sync()).await?
     }
 
     /// Perform a garbage collection operation.


crates/composefs-oci/src/skopeo.rs
```diff
@@ -10,12 +10,13 @@ use oci_spec::image::{Descriptor, ImageConfiguration, ImageManifest, MediaType};
...
 use tokio::{io::AsyncReadExt, sync::Semaphore};
 
-use composefs::{
-    fsverity::FsVerityHashValue, repository::Repository, splitstream::DigestMap, util::Sha256Digest,
-};
+use composefs::{fsverity::FsVerityHashValue, repository::Repository, util::Sha256Digest};
 
 use crate::{sha256_from_descriptor, sha256_from_digest, tar::split_async, ContentAndVerity};
 
+pub const TAR_LAYER_CONTENT_TYPE: u64 = 0x2a037edfcae1ffea;
```
cgwalters, line 17: Can you add a comment for where these came from? I'm guessing random? If so just add a comment `// Random unique ID` ?

That said I wonder if it wouldn't be nicer to store (variable length) strings for this in the format? Maybe it could go all the way to literally suggested to be the mediaType from OCI (if applicable)

crates/composefs/src/repository.rs
```diff
@@ -335,6 +335,12 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         Ok(object_id)
     }
 
+    pub fn has_named_stream(&self, name: &str) -> bool {
+        let stream_path = format!("streams/refs/{}", name);
+
+        readlinkat(&self.repository, &stream_path, []).is_ok()
```
cgwalters, line 341: I don't like "swallowing" errors like this, I'd say call `stat` instead and require it's `S_IFLNK`

crates/composefs/src/splitstream.rs
```diff
@@ -19,8 +19,31 @@ use crate::{
...
+    pub n_mappings: u64,
+    // Followed by n_refs ObjectIDs, sorted
+    // Followed by n_mappings MappingEntry, sorted by body
+    // Followed by zstd compressed chunks
+}
+
+#[derive(Clone, Debug, FromBytes, Immutable, IntoBytes, KnownLayout)]
+#[repr(C)]
+pub struct MappingEntry {
+    pub body: Sha256Digest,
```
cgwalters, line 41: If we're changing the format here...I think it'd be nice to make this one extensible.

However...bigger picture there's another consideration: There's obviously a metric *ton* of binary serialization formats out there. A custom one isn't wrong necessarily but...how about say [CBOR](https://www.rfc-editor.org/rfc/rfc8949.html) ? It has some usage and a proper RFC etc.

I guess a dividing line is "do we care about mmap()"? Probably not?

crates/composefs-oci/src/skopeo.rs
alexlarsson: Yes. These are random.

But, I'd rather avoid having variable length things in the header. That makes parsing it much more tricky.

crates/composefs/src/splitstream.rs
alexlarsson: splitstreams are essentially thin wrappers of existing binary formats (tars, ostree objects, etc), adding just references to other composefs repo objects. I'm not sure its overly helpful to use a complicated binary format for the wrapping, especially one which is completely different from the inner format.

That said, I agree that we should make it at least a bit extensible. This MR adds a magic header, but also adding a version field and a few  bytes of unused/unparsed space does seem quite useful.

crates/composefs-oci/src/skopeo.rs
alexlarsson: We could make it a real uuid tho

cgwalters: I'm fine keeping as u64 too.

crates/composefs/src/splitstream.rs
cgwalters: Although what about the algorithm agility? There's been some thoughts that for post-quantum crypto we may need to get away from sha256 in theory as far as I understand things.

crates/composefs-oci/src/skopeo.rs
alexlarsson: I added some comments here

doc/splitstream.md
```diff
@@ -22,39 +22,50 @@ extremely well.
...
+pub const SPLITSTREAM_MAGIC : [u8; 7] = [b'S', b'p', b'l', b't', b'S', b't', b'r'];
+
+struct MappingEntry {
+    pub body: Sha256Digest,
+    pub reference_idx: u64, // index into references table
+}
+
+struct SplitstreamHeader {
+    magic: [u8; 7], // Contains SPLITSTREAM_MAGIC
+    algorithm: u8,  // The fs-verity algorithm used, 1 == sha256, 2 == sha512
```
allisonkarlitskaya, line 41: We should also do fs-verity block size here.  That's usually expressed as a bit-shift count, so 12 or 16...

We could also write it like "fsverity-sha256-12" or so as a string... some relevant discussion in #181.

doc/splitstream.md
```diff
@@ -22,39 +22,50 @@ extremely well.
...
+    total_size: u64, // total size of inline chunks and external chunks
+    n_refs: u64,
+    n_mappings: u64,
+    refs: [ObjectID; n_refs]    // sorted
+    mappings: [MappingEntry; n_mappings] // sorted by body
```
allisonkarlitskaya, line 46: It's so sketch that we hardcode sha256 here... I think that's probably OK, but maybe we'd add an extension mechanism so we could add new types of mappings tables...

allisonkarlitskaya: Something like this from the start:
 - magic
 - n_sections
 - n_sections * (
   - section_start
   - section_size_in_bytes
 - )

We could name the sections but I think it's quite OK to just know what the numbers mean and require that they're all present, in order.  An empty section would be denoted by a zero size.

We could also get into compat vs. uncompat extensions... not sure how far it's worth going here...

allisonkarlitskaya (changes requested):
Let's talk about this.  I might have a bit of bandwidth to work on this if you like.

cgwalters:
The thing I'll raise here again is I'm still uncertain about the "centrality" of splitstream for composefs-rs; this previously was discussed in https://github.com/containers/composefs-rs/issues/14#issuecomment-2787378133

To continue and flesh that out, I still feel it'd be a lot easier to understand if what we always parsed was EROFS images, and splitstream was just a lookaside that could help map from the EROFS back bit-for-bit only for heavyweight serializations like tar - and would only be used when we actually needed to return back a tar. Storing small data objects like manifests and config JSON would just be saved bit for bit as is.

cgwalters:
Basically everyone dealing with composefs+OCI needs to deal with *two* binary serialization formats: tar and erofs. tar-split is funnily enough not binary but perhaps arguably should be...man I hadn't looked at this in a while but the choice of JSON with base64 encoded payloads is a bit wasteful...but eh, I guess compression still does well with it anyways. Anyways splitstream is a *third* binary format competing for mental bandwidth with both tar and erofs.

doc/splitstream.md
```diff
@@ -1,86 +1,158 @@
...
-For example, if we had a file that mapped `1234..` to `abcd..` and `5678..` to
-`efab..`, the header would look like:
+The file format consists of a fixed-sized header at the start of the file plus
+a number of sections located at arbitrary locations inside of the file.  All of
+these sections are referred to by a 64-bit `[start..end)` range expressed in
+terms of overall byte offsets within the complete file.
 
+```rust
+struct FileRange {
+    start: U64,
```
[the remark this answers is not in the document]
allisonkarlitskaya: That's from zerocopy: it's a little endian u64.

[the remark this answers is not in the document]
allisonkarlitskaya: Are you asking for a clarifying note in the docs?  That's a good suggestion.  I actually thought I had added one...

allisonkarlitskaya:
> Sorry, remind me why one would need to do that outside of a push operation?

Because the layers are referred to by their diffid from the config, which is the sha256 over their content, including the tar headers.  When we construct the image we want to ensure that the file we're constructing from hasn't been tampered, which means we need to measure it...

> Or a different approach here is to keep splitstreams mostly as is, but just don't make them the Source Of Truth for garbage collection.

So splitstreams contain references to the objects, which means that we need some way to keep those objects around if the splitstream exists.  But it's not the only story here: if you create an erofs image (eg. `prepare-boot`), you could then throw the splitstreams away and still do GC based on the references found in the image.

GC on the erofs is unfortunately substantially more expensive since we need to walk the tree to find and parse all the xattrs....

crates/composefs/src/splitstream.rs
```diff
@@ -6,172 +6,258 @@
...
-    /// Vector of digest map entries, kept sorted by body hash
-    pub map: Vec<DigestMapEntry<ObjectID>>,
+// The info block can be located anywhere, indicated by the "info" FileRange in the header.
+#[derive(Debug, FromBytes, Immutable, IntoBytes, KnownLayout)]
+struct SplitstreamInfo {
+    pub stream_refs: FileRange, // location of the stream references array
+    pub object_refs: FileRange, // location of the object references array
+    pub stream: FileRange,      // location of the zstd-compressed stream within the file
+    pub named_refs: FileRange,  // location of the compressed named references
+    pub content_type: U64,      // user can put whatever magic identifier they want there
```
alexlarsson: I thought you were looking at removing the content type? I think we should do that given the new stream type prefixes.

allisonkarlitskaya: There's one thing that makes me want to keep content type: some future "deep fsck" mode where we not only ensure basic connectivity at the repository level, but make sure that the splitstreams have the expected content.  We need to know what kind of file we're dealing with in order to do that.

alexlarsson (commented):
I looked mainly at the core splitstream and repo changes, not in detail at the oci changes. But overall I think this looks good. I thought you wanted to remove the content type, and i think we should. Also, i thought you wanted to add the ability to store the content identifier in the splitstream itself. That I'm not sure how useful it is in practice, but wouldn't hurt.

One question that comes up though, is: do we ever want to support streaming writes of a splitstream? I.e. do we ever expect individual inline chunks to be large (in comparison with the total ram size). Because, the file format as defined is not really amendable to streaming, because we build up the inline_buffer in ram before flushing it, and we can't flush early because we need to first write out the chunk size, and we can't back-patch the chunk size at flush time because the data is compressed.

We could make it streamable by doing a "virtual flush" when the inline buffer is at some max size. However, that will not be read back the same way, because we have APIs and code relying on the chunking being represented perfectly (for example read_inline_exact() and expected_bytes in ensure_chunk()).  If we want to be streamable, yet exactly reproduce the chunk boundaries we need to be able to mark chunks as being virtual or real.

allisonkarlitskaya:
> Isn't the whole point again of https://github.com/containers/composefs-rs/blob/main/doc/plans/oci-sealing-spec.md that we can construct the erofs and verify its integrity without also checksumming the entire tar?

But this assumes that all images we consume would have this label set on it... including for base layers that we might want to build on, or images that we import to seal for ourselves...

There is an interesting argument here, I suppose: we don't care about the integrity of the OCI layers at all because the only thing that ultimately matters is that the assembled image has the correct bits in it.  By that logic, someone could (wild example) sneak a rootkit into one layer and delete it in another, and the checksum of the image in the end would still be the the same... and indeed, the system image wouldn't contain the rootkit...

We also need to worry a bit about parsing untrusted data... you could corrupt a layer by playing with header bits (or whatever) in an attempt to exploit a bug in composefs-rs without changing the resulting image... Rust helps a whole lot here, but it's not perfect...

So ya, there's some very specific scenario in which what you're proposing might be OK.  And we might even be able to make that "the normal case" in the future... but all told I think, I'm not quite ready to walk away from the idea that diffids are meaningful and should be validated...

allisonkarlitskaya:
New year, same old PR...

I want to land this, preferably today.  It's 99%.  There's too much outstanding work and a lot of it is going to conflict.  We shouldn't have such a large change out for so long...

doc/splitstream.md
```diff
@@ -1,86 +1,164 @@
...
+```rust
+const SPLITSTREAM_MAGIC: [u8; 11] = *b"SplitStream";
 
+struct SplitstreamHeader {
+    pub magic: [u8; 11],  // Contains SPLITSTREAM_MAGIC
+    pub version: u8,      // must always be 0
+    pub _flags: U16,      // is currently always 0 (but ignored)
+    pub algorithm: u8,    // kernel fs-verity algorithm identifier (1 = sha256, 2 = sha512)
+    pub lg_blocksize: u8, // log2 of the fs-verity block size (12 = 4k, 16 = 64k)
+    pub info: FileRange,  // can be used to expand/move the info section in the future
```
cgwalters: Given that `FileRange` is effectively a pointer, I think it'd be helpful to include at least pseudo-types for the values here. Like `pub info: FileRange // (points to SplitStreamInfo)`

doc/splitstream.md
```diff
@@ -1,86 +1,164 @@
...
+referred to from the info section (via `stream_refs` and `object_refs`).
 
-### Data blocks
+The number of items in the array is determined by the size of the section
+divided by the size of the fs-verity hash value (determined by the algorithm
+identifier in the header).
 
-After the header comes a number of data blocks.  Each block starts with a u64
-le "size" field followed by some amount of data.
+The values are not in any particular order, but implementations should produce
```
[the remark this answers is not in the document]
allisonkarlitskaya: The reason for having it unsorted is that we want to write stable indexes into the stream as we go and if do sorted insert (or sort at end) then the indexes can change.  Alex's first patch did sort this (and had a lookup function for a second pass at the end)  for ostree but he never tried to apply it to the content of the splitstream itself.  I made the change when we did that.

doc/splitstream.md
```diff
@@ -1,86 +1,164 @@
...
-    first may be unaligned.  This decision was taken to keep the format simple,
-    and because the data is compressed before being stored, which removes the
-    main advantages of aligned data.
+It's possible to have named references to other streams.  These are stored in
+the `named_refs` section referred to from the info section.
 
-  - "External" blocks (`size == 0`): in this case the length of the data is 32
-    bytes.  This is the binary form of a sha256 hash value and is a reference
-    to an object in the composefs repository (by its fs-verity digest).
+This section is also zstd-compressed, and is a number of nul-terminated text
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's going to be hex strings, which would compress well.  Alex mentioned this as well but I don't think it makes a whole lot of difference either way... the only time those references matter to us is when we're accessing the content and it's also compressed...

The reason I did it is that everything in the splistream right now is either
 - header stuff (magic numbers, offsets, etc)
 - hashes in binary form (uncompressable)
 - other data (which is compressed)

For the fast "do GC" thing we have separate flat binary arrays..

doc/splitstream.md
```diff
@@ -1,86 +1,164 @@
...
-`efab..`, the header would look like:
+```rust
+const SPLITSTREAM_MAGIC: [u8; 11] = *b"SplitStream";
 
+struct SplitstreamHeader {
+    pub magic: [u8; 11],  // Contains SPLITSTREAM_MAGIC
+    pub version: u8,      // must always be 0
+    pub _flags: U16,      // is currently always 0 (but ignored)
+    pub algorithm: u8,    // kernel fs-verity algorithm identifier (1 = sha256, 2 = sha512)
+    pub lg_blocksize: u8, // log2 of the fs-verity block size (12 = 4k, 16 = 64k)
```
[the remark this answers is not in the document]
cgwalters: >  That being said, I have a vague long term plan that we might have some sort of an ObjectStore trait that has a more flexible identifier...

Yep https://github.com/containers/composefs-rs/pull/202 adds such a trait, but needs more work to actually abstract the digest stuff - and would be more of a conflict fest here, so I didn't try hard on it. We can look at that once this lands.

doc/splitstream.md
allisonkarlitskaya: I tried to do this with proper generics but it turned into the "everything needs an objectid generic" thing because some of those ranges refer to the array of object IDs.  I think I'm going to leave this one alone.

cgwalters:
Regression in this fixed in https://github.com/containers/composefs-rs/pull/207/changes/d04f9e2ea2ab3b35c10b3d444160004462428714

allisonkarlitskaya:
> I think it would have made sense to split out the sync() thing to a separate PR for the record.

You're probably right about that.  It was a bit of a yolo at the end there...

PR #187, opened 2025-10-08
Drop support for embedding type1 entries directly in an image

cgwalters:
We're not going to use this in bootc and right now we just error if we encounter it, and I am not aware of a good use case.


Johan-Liebert1:
Most cases would be either bootc or compoesfs-boot create generating these config flies using the kernel + initrd right? Would we have a case where we'd want to use the configs provided by a user?

PR #188, opened 2025-10-13
erofs: Fix reading directory entries

Johan-Liebert1:
Read inline inode entries only if not empty

Before this patch, gathering objects from EROFS, the program panics with
the following error

```
thread 'main' panicked at crates/composefs/src/erofs/reader.rs:404:13:
range end index 12 out of range for slice of length 0
```

### DIFF
--- a/crates/composefs/src/erofs/reader.rs
+++ b/crates/composefs/src/erofs/reader.rs
@@ -167,7 +167,7 @@ pub trait InodeOps {
     /// Returns the extended attributes section if present
     fn xattrs(&self) -> Option<&InodeXAttrs>;
     /// Returns the inline data portion
-    fn inline(&self) -> &[u8];
+    fn inline(&self) -> Option<&[u8]>;
     /// Returns the range of block IDs used by this inode
     fn blocks(&self, blkszbits: u8) -> Range<u64>;
 }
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (dismissed):
+1 on wondering why the order had to change... it seems a bit odd.

The commit message could be a bit better.  Why did you hit this?  Is there some error elsewhere with handling empty blocks?  Or is it a general cleanup kind of thing?

Thanks!

hsiangkao:
> Can we land #189 and then this can be rebased on top ?

An empty directory inode is actually an invalid directory.  Before the `..` dirent is useful for exportfs at least...
So, it seems like a regression out of C-version mkcomposefs...

Johan-Liebert1:
So, apparently the issue was not with the read ordering but parsing inline entry even when they're empty

hsiangkao:
> > An empty directory inode is actually an invalid directory. Before the .. dirent is useful for exportfs at least...
> > So, it seems like a regression out of C-version mkcomposefs...
> 
> Thanks for commenting. Is this not something that `fsck.erofs` catches though?

is there a quick golden binary for this? I could try it out and fix if it doesn't catch this.

Johan-Liebert1:
> One minor tweak we could make is to have .inline() return Optional<>... if we have zero bytes of inline data then arguably there's not an inline block at all...

makes sense. Updated

crates/composefs/src/erofs/reader.rs
```diff
@@ -175,8 +175,14 @@ impl<Header: InodeHeader> InodeOps for &Inode<Header> {
         }
     }
 
-    fn inline(&self) -> &[u8] {
-        &self.data[self.header.xattr_size()..]
+    fn inline(&self) -> Option<&[u8]> {
+        let data = &self.data[self.header.xattr_size()..];
```
cgwalters: Unrelated to this but it made me think that we likely panic if the xattr size is out of bounds.

Of course handling panic freedom here in this whole module probably needs nontrivial work.

cgwalters (approved):

PR #189, opened 2025-10-15
Add unit tests for erofs reading+write cycles

cgwalters:
A lot of `Assisted-by: Claude Code` here, but the goal is to add a test case for https://github.com/containers/composefs-rs/pull/188 that can just have its `[ignore]` removed.


cgwalters:
Hmm looks like there's CI gardening to do here unrelated to extant PRs

PR #192, opened 2025-10-20
tree-wide: Add docs for all `pub` items

cgwalters:
Add `#[deny(missing_docs)]` to enforce this.

Assisted-by: Claude Code

### DIFF
--- a/crates/composefs-boot/src/bootloader.rs
+++ b/crates/composefs-boot/src/bootloader.rs
@@ -49,24 +49,56 @@ fn substr_range(parent: &str, substr: &str) -> Option<Range<usize>> {
     }
 }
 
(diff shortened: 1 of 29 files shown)


crates/composefs-boot/src/bootloader.rs
```diff
@@ -134,6 +172,16 @@ impl<ObjectID: FsVerityHashValue> Type1Entry<ObjectID> {
     // This is a bit of a strange operation: for each file mentioned in the bootloader entry, move
     // the file into the given 'entry_id' pathname and rename the entry file itself to
     // "{entry_id}.conf".
+    /// Relocates boot resources to a new entry ID directory.
```
Johan-Liebert1: We could remove the above comment?

PR #194, opened 2025-10-20
doc/plans: Add oci sealing spec and impl documents

cgwalters:
I used some AI assistance to consolidate all the Github discussions in the previous composefs repository and issues into a spec and implmenetation plan.


jeckersb (approved):
LGTM, I noted one place where SHA256 is implied, but there's a few other places where this explicitly says "SHA256" and it would make sense to be generic.  But then there's some other places talking about the OCI config specifically, where "SHA256" *does* make sense.  Not a huge deal, we can follow up and clarify that later.

PR #199, opened 2025-12-28
Support for handling sha512-addressed repos in cfsctl and composefs-setup-root

### LINKED ISSUE #197: `cfsctl` digests are incompatible with recent `bootc`
Since bootc-dev/bootc@02c57b3e30f28de37d7f83cdf0a3a775d5549b9f, `bootc` exclusively uses SHA-512 digests. This means that building a composefs-based image and installing it with `bootc install to-disk` won't work:

- If you use a Type 1 BLS entry (`/boot/loader/entries/`), installation will work properly, but `bootc` will write a SHA-512 digest to `/boot/loader/entries/[…].conf`, so booting the system fails in the initramfs with an `Error: Invalid string length` error.

  https://github.com/containers/composefs-rs/blob/0f636031a1ec81cdd9e7f674909ef6b75c2642cb/crates/composefs-setup-root/src/main.rs#L239

- If you use a Type 2 BLS entry (UKI), then `cfsctl` will only output SHA-256 digests

  https://github.com/containers/composefs-rs/blob/0f636031a1ec81cdd9e7f674909ef6b75c2642cb/crates/cfsctl/src/main.rs#L255-L261

  so the digests embedded in the UKI will also use SHA-256. This means that when installing with `bootc install to-disk`, you'll get an `error: Installing to disk: Setting up composefs boot: Setting up UKI boot: Writing <kver>.efi to ESP: Parsing composefs=: Invalid string length` error.

  https://github.com/bootc-dev/bootc/blob/71dc8e52d3697155c1f0b668313a00b555ce78f1/crates/lib/src/bootc_composefs/boot.rs#L609

There are examples of both methods in [`examples/`](https://github.com/containers/composefs-rs/tree/main/examples), but because of this, none of them are working for me right now.

(I may also be completely misunderstanding something and/or building my containers incorrectly, so sorry in advance if everything that I mentioned above turns out to be wrong.)

chaserhkj:
This PR changes `cfsctl` to use sha512-addressed composefs repo by default, this confronts to the behavior of bootc, making the two tools compatible with each other. An additional flag `--use-sha256-object-id` is also added to enable `cfsctl` to work with legacy repos.

kernel cmdline parse logic in `composefs-setup-root` is also changed, both sha256 and sha512 length addresses would be attempted before returning an error, improving robustness within initramfs.

Fixes #197

### DIFF
--- a/crates/cfsctl/src/main.rs
+++ b/crates/cfsctl/src/main.rs
@@ -11,14 +11,14 @@ use std::{
 };
 
 use anyhow::Result;
-use clap::{Parser, Subcommand};
+use clap::{Parser, Subcommand, ValueEnum};
 
 use rustix::fs::CWD;
 
 use composefs_boot::{write_boot, BootOps};
 
 use composefs::{
-    fsverity::{FsVerityHashValue, Sha256HashValue},
+    fsverity::{FsVerityHashValue, Sha256HashValue, Sha512HashValue},
     repository::Repository,
 };
 
@@ -33,6 +33,10 @@ pub struct App {
     #[clap(long, group = "repopath")]
     system: bool,
 
+    /// What hash digest type to use for composefs repo
+    #[clap(long, value_enum, default_value_t = HashType::Sha512)]
+    hash: HashType,
+
     /// Sets the repository to insecure before running any operation and
     /// prepend '?' to the composefs kernel command line when writing
     /// boot entry.
@@ -41,6 +45,13 @@ pub struct App {
 
     #[clap(subcommand)]
     cmd: Command,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ValueEnum, Default)]
+enum HashType {
+    Sha256,
+    #[default]
+    Sha512,
 }
 
 #[cfg(feature = "oci")]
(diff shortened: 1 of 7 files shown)


crates/composefs-setup-root/src/main.rs
```diff
@@ -222,6 +222,16 @@ fn gpt_workaround() -> Result<()> {
     Ok(())
 }
 
+// Try parse cmdline with sha512 digest address first, if failed, parse again with legacy sha256 digest address
+fn parse_image_address(cmdline: &str) -> Result<(String, bool)> {
+    if let Ok((id, insecure)) = get_cmdline_composefs::<Sha512HashValue>(cmdline) {
+        Ok((id.to_hex(), insecure))
+    } else {
```
Johan-Liebert1: We'd only want to do this if we encounter `FromHexError::InvalidStringLength`
Any other error should be propagated, imo

Johan-Liebert1 (changes requested):
I had reviewed earlier, but forgot to press submit :|

chaserhkj:
Sorry missed another `cargo fmt` that pulled in another CI failure. Just fixed. I'll just add cargo fmt as my git hook

crates/cfsctl/src/main.rs
```diff
@@ -43,6 +47,13 @@ pub struct App {
     cmd: Command,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ValueEnum, Default)]
+enum HashType {
+    Sha256,
+    #[default]
```
cgwalters: So...this will increase compatibility with bootc when using the raw CLI, but it will also have the effect of changing the default here too, including AFAIK all CI jobs here. I think that's probably fine...but worth noting.

cgwalters (approved):
So I'm OK with this, but I do think we should add the repo metadata at some point.

In practice...because we're not shipping this project even on crates.io (much less as a standalone package/component) making "breaking" changes like this doesn't matter much.

But...we do want to eventually do that I think in order to replace the composefs-c project. But when we do *that* there will probably be deeper changes anyways.

crates/composefs-setup-root/src/main.rs
```diff
@@ -236,11 +252,12 @@ fn setup_root(args: Args) -> Result<()> {
         Some(cmdline) => cmdline,
         None => &std::fs::read_to_string("/proc/cmdline")?,
     };
-    let (image, insecure) = get_cmdline_composefs::<Sha256HashValue>(cmdline)?;
+
+    let (image_addr, insecure) = parse_image_address(cmdline)?;
```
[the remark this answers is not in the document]
chaserhkj: Agreed. But for the scope of this PR without introducing metadata changes yet, this should be good enough to keep existing installations happy by maintaining backward-compatibility.

chaserhkj:
Indeed I forgot about `default_value_t`. Just fixed, PTAL.

If CI is still failing due to the newly added flag, I could try adding the flag to the CI as well. Let's see how it goes with the default fix first.

Johan-Liebert1:
Thanks. Reran the workflows

chaserhkj:
Rebased to main

Johan-Liebert1 (approved):
Looks good to me

PR #200, opened 2026-01-03
cfsctl: Fix garbage collection

chaserhkj:
This PR fixes the broken garbage collection implementation provided by `cfsctl`. To my knowledge this GC implementation is *not* used anywhere else, but this fixed implementation could be helpful for downstream project like bootc as well, see bootc-dev/bootc#1808

Previously, the GC implementation was completely broken as it didn't add any object IDs to the live set and would thus always mark every object for deletion. The code in `gc_category` seems to be assuming an old composefs structure where non-first-level entries in `streams/` and `images/` directory (e.g. `streams/refs/some_distro/some_version`) could directly link to object stores. But currently these links always link to first-level entries first and first-level entries would then link to objects in store.

This PR fixes this by doing a proper walk to add all objects referred from named references to the live set, such that all unlinked objects would be marked for deletion.

Furthermore, the old implementation was doing a naive shallow walk for the streams, this is problematic for pulled OCI images in composefs repo, since they have two layers of links, a config split stream linking all layers, and each layers linking to their layer contents. This PR adds a full walk algorithm to walk down and prune the entire stream tree to mark unlinked objects for deletion.

Currently this is still dry-run only, but I have changed the output format to add a "#" before all non-delete lines and the output could now just be piped to a shell to perform the deletion.

Note that bootc has its own GC implementations [here](https://github.com/bootc-dev/bootc/blob/main/crates/lib/src/bootc_composefs/gc.rs#L166). But bootc uses bootloader entries as part of the GC root, which I believe should be considered out of composefs scope. Bootc GC also does not prune streams. I think ideally we should have a complete GC implementation in composefs and bootc should just forward the call.

### DIFF
--- a/crates/composefs/src/repository.rs
+++ b/crates/composefs/src/repository.rs
@@ -18,7 +18,7 @@ use std::{
     thread::available_parallelism,
 };
 
-use log::{debug, info, trace};
+use log::{debug, trace};
 use tokio::sync::Semaphore;
 
 use anyhow::{bail, ensure, Context, Result};
@@ -97,10 +97,25 @@ impl<ObjectID: FsVerityHashValue> Drop for Repository<ObjectID> {
     }
 }
 
-// For Repository::gc_category
+/// For Repository::gc_category
 enum GCCategoryWalkMode {
     RefsOnly,
     AllEntries,
+}
+
+/// Statistics from a garbage collection operation.
+///
+/// Returned by [`Repository::gc`] to report what was (or would be) removed.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct GcResult {
+    /// Number of unreferenced objects removed (or that would be removed)
+    pub objects_removed: u64,
+    /// Total bytes of object data removed (or that would be removed)
+    pub objects_bytes: u64,
+    /// Number of broken symlinks removed in images/
+    pub images_pruned: u64,
+    /// Number of broken symlinks removed in streams/
+    pub streams_pruned: u64,
 }
 
 impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
@@ -912,7 +927,10 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
     }
 
     // Remove all broken links from a directory, may operate recursively
-    fn cleanup_broken_links(fd: &OwnedFd, recursive: bool) -> Result<()> {
+    /// Remove broken symlinks from a directory.
+    /// If `dry_run` is true, counts but does not remove. Returns the count.
+    fn cleanup_broken_links(fd: &OwnedFd, recursive: bool, dry_run: bool) -> Result<u64> {
+        let mut count = 0;
         for item in Dir::read_from(fd)? {
             let entry = item?;
             match entry.file_type() {
@@ -928,7 +946,7 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
                             OFlags::RDONLY | OFlags::CLOEXEC,
                             Mode::empty(),
                         )?;
-                        Self::cleanup_broken_links(&dirfd, recursive)?;
+                        count += Self::cleanup_broken_links(&dirfd, recursive, dry_run)?;
                     }
                 }
 
@@ -938,8 +956,11 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
                         .filter_errno(Errno::NOENT)
                         .context("Testing for broken links")?;
                     if result.is_none() {
-                        unlinkat(fd, filename, AtFlags::empty())
-                            .context("Unlinking broken links")?;
+                        count += 1;
+                        if !dry_run {
+                            unlinkat(fd, filename, AtFlags::empty())
+                                .context("Unlinking broken links")?;
+                        }
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


Johan-Liebert1:
I think a lot of this will change once https://github.com/containers/composefs-rs/pull/185 lands. We'd probably want to wait until then, I think

crates/composefs/src/repository.rs
```diff
@@ -608,53 +616,96 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
+            return Ok(());
         }
+        walked_streams.insert(stream_digest.to_owned());
+
+        let mut split_stream = self.open_stream(stream_digest, None)?;
+        let mut streams_to_walk = HashSet::new();
+        split_stream.get_object_refs(|id| {
+            println!("#   with {id:?}");
+            objects.insert(id.clone());
+            if stream_map.contains_key(id) {
```
cgwalters: AI: (low) Minor nit: `contains_key` followed by `get` is slightly redundant. Could use `if let Some(digest) = stream_map.get(id)` instead to avoid the double lookup.

crates/composefs/src/repository.rs
```diff
@@ -608,53 +616,96 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
+        // linking to layer contents. This function walks the streams down and collects all linked objects
+        if walked_streams.contains(stream_digest) {
+            return Ok(());
         }
+        walked_streams.insert(stream_digest.to_owned());
+
+        let mut split_stream = self.open_stream(stream_digest, None)?;
+        let mut streams_to_walk = HashSet::new();
+        split_stream.get_object_refs(|id| {
+            println!("#   with {id:?}");
```
cgwalters: AI: *Important*: Debug output via `println!` in library code is generally undesirable. Consider either:
1. Using a tracing/logging crate and `debug!` or `trace!` level
2. Making debug output conditional on a flag/parameter
3. Returning the information in a structured way for the caller to handle

This applies to all the `println!` calls in `walk_streams` and `gc`.

crates/composefs/src/repository.rs
```diff
@@ -608,53 +616,96 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
-            }
+    // Traverse split streams to resolve all linked objects
+    fn walk_streams(
+        &self,
+        stream_digest: &str,
+        walked_streams: &mut HashSet<String>,
+        objects: &mut HashSet<ObjectID>,
+        stream_map: &HashMap<ObjectID, String>,
+    ) -> Result<()> {
+        // A split stream links to stored objects, but the the linked objects themselves could be streams that links
```
cgwalters: AI: (low) Typo: "the the" should be "the".

crates/composefs/src/repository.rs
```diff
@@ -608,53 +616,96 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
+    // object IDs for each entry, as a reverse lookup map
+    // If refs_only == false, all entries will be returned, and only orphans in
+    // objects/ will be GC'd
+    // If refs_only == true, only entries explicitly referred to in <category>/refs
+    // will be returned, allow unlinked entries to be GC'ed
+    fn gc_category(&self, category: &str, refs_only: bool) -> Result<HashMap<ObjectID, String>> {
         let Some(category_fd) = self
             .openat(category, OFlags::RDONLY | OFlags::DIRECTORY)
             .filter_errno(Errno::NOENT)
             .context("Opening {category} dir in repository")?
```
cgwalters: AI: (low) Pre-existing issue, but worth noting: this string literal doesn't interpolate `{category}`. The `context()` method takes a string, not a format macro. Same on line 643.

crates/composefs/src/repository.rs
```diff
@@ -675,21 +726,21 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
-            println!("{object:?} lives as a stream");
-            objects.insert(object.clone());
-
-            let mut split_stream = self.open_stream(&object.to_hex(), None)?;
-            split_stream.get_object_refs(|id| {
-                println!("   with {id:?}");
-                objects.insert(id.clone());
-            })?;
+        let mut walked_streams = HashSet::new();
+        let stream_map = self.gc_category("streams", false)?;
```
cgwalters: AI: The semantics here are subtle: `refs_only=false` returns all entries (including unreferenced ones), but only for building the reverse lookup map. Then `refs_only=true` returns only referenced entries for actual GC roots. A comment explaining this two-pass strategy would help maintainability.

cgwalters (commented):
We're not adding any tests here, but we definitely want that. Basic ones would be ensuring adding content and removing all the streams also GC's all objects, and two streams with shared objects but removing just one keeps the right objects etc.

chaserhkj:
I rebased the branch onto main after the splitstream changes. It seems that in the new format manifests referring to other streams in a separate table `stream_refs`. This slightly changes how things should be handled in my gc implementation. I'll work that changes in first then address all the review comments and add unit tests. I also have a few more improvements on the gc on my other branch that I feel should fall in the same scope so I might also pull these in.

crates/composefs/src/repository.rs
```diff
@@ -780,7 +782,7 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
         Ok(ObjectID::from_object_pathname(link_content.to_bytes())?)
     }
 
-    fn walk_symlinkdir(fd: OwnedFd, objects: &mut HashSet<ObjectID>) -> Result<()> {
+    fn walk_symlinkdir(fd: OwnedFd, entry_digests: &mut HashSet<CString>) -> Result<()> {
```
cgwalters: I think we should prefer Rust-native representations in memory.

crates/composefs/src/repository.rs
```diff
@@ -816,53 +825,165 @@ impl<ObjectID: FsVerityHashValue> Repository<ObjectID> {
...
-        .filter_errno(Errno::NOENT)
-        .context("Opening {category}/refs dir in repository")?
-        {
-            Self::walk_symlinkdir(refs, &mut objects)?;
+        let mut entry_digests = HashSet::new();
+        if refs_only {
+            if let Some(refs) = openat(
+                &category_fd,
+                "refs",
+                OFlags::RDONLY | OFlags::DIRECTORY,
```
cgwalters: missing `CLOEXEC`?

cgwalters (commented):
Thanks for this!! Sorry about the delay. I only have a relatively superficial review so far

PR #205, opened 2026-01-07
composefs-oci: Fix wrong OCI opaque whiteout filename

### LINKED ISSUE #203: Podman image layers are processed incorrectly
This can be observed in my [bootc](https://github.com/bootc-dev/bootc) image [arch-bootc-testing](https://github.com/hecknt/arch-bootc-testing). It is derived FROM [archlinux-bootc](https://github.com/hecknt/archlinux-bootc).

In [archlinux-bootc](https://github.com/hecknt/archlinux-bootc), I install the `linux` package. Then, in [arch-bootc-testing](https://github.com/hecknt/arch-bootc-testing), i remove the `linux` package, rm `/usr/lib/modules`, and then install `linux-zen`. When I boot into this image (with the composefs backend of [bootc](https://github.com/bootc-dev/bootc)), the kernel that I removed in the image build is still there. This is different from what shows up in podman.

(Pictured: Left is the image loaded into podman. Right is the exact same image booted with the composefs backend.)
Image

Weirdly enough, the `linux` package is actually removed from the pacman database. It just isn't removed from the image itself.

Image

chaserhkj:
Per OCI spec [ref], opaque white out filename should be `.wh..wh..opq` instead of `.wh..wh.opq` (two dual dots instead of one)

ref: https://github.com/opencontainers/image-spec/blob/main/layer.md#opaque-whiteout

Fixes #203

### DIFF
--- a/crates/composefs-oci/src/image.rs
+++ b/crates/composefs-oci/src/image.rs
@@ -69,8 +69,8 @@ pub fn process_entry<ObjectID: FsVerityHashValue>(
 
     let bytes = filename.as_bytes();
     if let Some(whiteout) = bytes.strip_prefix(b".wh.") {
-        if whiteout == b".wh.opq" {
-            // complete name is '.wh..wh.opq'
+        if whiteout == b".wh..opq" {
+            // complete name is '.wh..wh..opq'
             dir.clear();
         } else {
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Thanks!

PR #208, opened 2026-01-09
oci: Parallelize object storage within tar layers

cgwalters:
Add parallel object storage when processing tar archives. Large files are streamed to O_TMPFILE via a channel, and fs-verity digests are computed in background blocking tasks. This avoids blocking the async runtime while allowing multiple files to be processed concurrently.

Key components:

- FsVerityHasher: Made public for incremental digest computation from files on disk without loading them entirely into memory.

- Repository:
  - create_object_tmpfile(): Create O_TMPFILE for streaming writes
  - finalize_object_tmpfile(): Compute verity, enable it, link to objects/
  - write_semaphore(): Shared semaphore limiting concurrent writes
  - register_stream(): Register stored object as named stream (with sync)

- SplitStreamBuilder: Collects inline data and pending external object handles, resolves ObjectIDs at finalization with proper deduplication.

- split_async(): Now takes AsyncBufRead, streams via channel to blocking tasks, uses repository's shared semaphore for backpressure.

For a 2GB tar (10,000 files × 200KB), achieves ~7x speedup: ~980ms → ~140ms (~14 GB/s throughput).

### DIFF
--- a/crates/composefs-oci/Cargo.toml
+++ b/crates/composefs-oci/Cargo.toml
@@ -13,6 +13,7 @@ version.workspace = true
 [dependencies]
 anyhow = { version = "1.0.87", default-features = false }
 async-compression = { version = "0.4.0", default-features = false, features = ["tokio", "zstd", "gzip"] }
+bytes = { version = "1", default-features = false }
 composefs = { workspace = true }
 containers-image-proxy = { version = "0.9.0", default-features = false }
 hex = { version = "0.4.0", default-features = false }
@@ -22,6 +23,7 @@ rustix = { version = "1.0.0", features = ["fs"] }
 sha2 = { version = "0.10.1", default-features = false }
 tar = { version = "0.4.38", default-features = false }
 tokio = { version = "1.24.2", features = ["rt-multi-thread"] }
+tokio-util = { version = "0.7", default-features = false, features = ["io"] }
 
 [dev-dependencies]
 similar-asserts = "1.7.0"
--- a/crates/composefs-oci/src/skopeo.rs
+++ b/crates/composefs-oci/src/skopeo.rs
@@ -18,7 +18,10 @@ use containers_image_proxy::{ImageProxy, ImageProxyConfig, OpenedImage};
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use oci_spec::image::{Descriptor, ImageConfiguration, ImageManifest, MediaType};
 use rustix::process::geteuid;
-use tokio::{io::AsyncReadExt, sync::Semaphore};
+use tokio::{
+    io::{AsyncReadExt, BufReader},
+    sync::Semaphore,
+};
 
 use composefs::{fsverity::FsVerityHashValue, repository::Repository};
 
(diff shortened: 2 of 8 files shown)


cgwalters:
OK...so...it's hard to overstate just HOW MUCH BETTER this makes `bootc install` with composefs.

I need to inject some precise timings but before a fresh install was measured in minutes, now it's measured in seconds. The "writing files" phase takes about 10s here which is SO SO SO much better than before.

jeckersb:
I need to spend some more hands-on time with this, but at the eyeball level this looks awesome!  Definitely want to A/B bootc installs with/without to compare, probably a job for next week at this point though.

cgwalters:
> Also: I'm kinda curious about why this is such a performance increase and I wonder if it's being conflated with https://github.com/containers/composefs-rs/commit/c3677f1a3599be0015db6798b60c5bdfa3e43d7e

I had an agent work on this overnight, I didn't do a deep verification but smells right to me (and matches your intuition):


# Performance Improvement Summary: composefs-rs PR #208

## Benchmark: `bootc install` Composefs Import Time

| Version | Commit | Import Time | Speedup vs Baseline |
|---------|--------|-------------|---------------------|
| **Baseline (fdatasync per object)** | `54c3c6d` | ~580s (~9.7 min) | 1x |
| **syncfs optimization** | `c3677f1` | ~29.4s | **~20x** |
| **syncfs + parallel** | `5f1cce7` + fix | ~16.7s | **~35x** |

## Root Cause of Baseline Slowness

The baseline called `fdatasync()` after writing each object file. While this ran inside `spawn_blocking` (so it didn't block the async runtime directly), the **thousands of per-object fdatasync calls** forced the filesystem to create a massive number of journal commits, causing severe I/O overhead.

## Optimization Breakdown

### 1. syncfs optimization (Allison Karlitskaya)

- **Commit**: `c3677f1` ("repository: change our sync story")
- **Change**: Replace per-object `fdatasync()` with a single `syncfs()` before linking the splitstream
- **Why it helps**: Instead of thousands of individual journal commits, there's now one bulk sync at the end
- **Time**: ~29.4s → **~20x faster**

### 2. Parallel object storage (Colin Walters)

- **Commit**: `5f1cce7` ("oci: Parallelize object storage within tar layers")
- **Change**: Process multiple large files concurrently within each tar layer using channels and `spawn_blocking` tasks with semaphore-based concurrency limiting
- **Why it helps**: Previously files were processed sequentially (one at a time). Now multiple files stream to tmpfiles in parallel, with fs-verity digests computed concurrently
- **Time**: ~16.7s → **~2x additional**, **~35x total**

## Bug Fix Required

The parallel implementation had a race condition where forked processes could inherit writable file descriptors, causing `enable_verity` to fail with "File is opened for writing".

**Fix**: Changed `finalize_object_tmpfile()` to use `enable_verity_maybe_copy()` instead of `enable_verity_with_retry()`. This creates a copy of the file if verity can't be enabled on the original, providing robustness against fd inheritance races.

## Visual Summary

```
┌─────────────────────────────────────────────────────────────────┐
│           Composefs Import Performance Improvement              │
├─────────────────────────────────────────────────────────────────┤
│                                                                 │
│  Baseline                        ████████████████████  580s     │
│  (fdatasync per object)                                         │
│                                                                 │
│  + syncfs optimization           █  29.4s  (20x faster)         │
│  (single bulk sync)                                             │
│                                                                 │
│  + parallel file processing      ▌  16.7s  (35x faster)         │
│  (concurrent I/O within layers)                                 │
│                                                                 │
└─────────────────────────────────────────────────────────────────┘
```

## Test Environment

- Benchmark command: `bcvk to-disk --composefs-backend --disk-size=10G localhost/bootc target/bench-disk.img`
- Image: CentOS Stream 10 bootc with composefs-sealeduki-sdboot variant
- Multiple runs averaged for each configuration

PR #209, opened 2026-01-09
fs: Add read_container_root() and copy metadata from /usr -> /

cgwalters:
OCI container layer tars often don't include a root directory entry, and when they do, container runtimes (Podman, Docker) ignore it and use hardcoded defaults. We had some attempt to handle this with our "stat root" stuff but that only handled mtime - i.e. not permissions or other metadata.

Since the contents of `/` are ill defined, let's just default to defining them based on `/usr` since I am not aware of anyone who cares for those to be different (*except* for the selinux label but we handle that already).

Add new APIs to handle this:

- copy_root_metadata_from_usr(): copies mode, uid, gid, mtime, xattrs from /usr to root directory for deterministic metadata

- read_container_root(): convenience wrapper that calls read_filesystem() then copy_root_metadata_from_usr()

Remove the incomplete have_root_stat/ensure_root_stat() mechanism which only handled mtime. Callers should now explicitly use the new helpers.

Update cfsctl commands (compute-id, create-image, create-dumpfile) to use read_container_root() by default. Add --no-propagate-usr-to-root flag to use raw read_filesystem() for testing.

### DIFF
--- a/crates/composefs-boot/src/lib.rs
+++ b/crates/composefs-boot/src/lib.rs
@@ -14,6 +14,8 @@ pub mod selabel;
 pub mod uki;
 pub mod write_boot;
 
+use std::ffi::OsStr;
+
 use anyhow::Result;
 
 use composefs::{fsverity::FsVerityHashValue, repository::Repository, tree::FileSystem};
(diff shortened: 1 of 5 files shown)


cgwalters:
Now on `/run` we could also error out instead of emptying it. We could also try to allow it, but it would make working around the *other* podman bug a lot more annoying; I think we'd always have to make mount points in `/` which is too ugly to live.

I don't have a really strong opinion on erroring versus emptying, the main reason I chose emptying is just because we already were doing that for `/sysroot` and `/boot`.

PR #211, opened 2026-01-21
examples: Pass `--hash sha256` to unified examples

Johan-Liebert1:
Remove `/var/lib/systemd/random-seed` to prevent `user.` xattr diff
between dumpfile created from mounted fs and oci container

Also update to use fedora 43 for tests


Johan-Liebert1:
I thought it was because of #209, but I guess not. Will need to debug it a bit more

PR #216, opened 2026-01-27
    oci: Add OCI-native manifest/config storage and image management

cgwalters:
Switch how we store OCI images to also include and identify
images by having explicit `oci/` tags which point to a manifest,
which points to a config.
    
We continue to generate splitstream for manifest and config, but
we now always store the original context as "external" objects.
A big rationale for this is it will align with a future proposal
for composefs OCI sealing where we include fsverity signatures
for the manifest/config as a detached object.
    
With this for example, bootc can stop storing the manifest on its own.
See-also: https://github.com/bootc-dev/bootc/pull/1830
   
Other big features:
    
 - Full general support for OCI artifacts as well
 - Support for OCI referrers as well (to store sigstore and future
   composefs signatures)
   
Finally we have an initial sketch for storing multi-arch images.


allisonkarlitskaya:
I looked briefly.  This seems like pretty cool/useful stuff...

There's a bit of tension here with #218.  What's the eventual plan?  Import via containers-image-proxy-rs, or skip skopeo and use the cstor-rs stuff (via jsonrpc) directly?

PR #218, opened 2026-01-29
Import cstor-rs, use for reflinked imports from containers-storage

cgwalters:
Obviously massively `Assisted-by: OpenCode (Opus 4.5)` here; I did a lot of design work of course and but this is looking pretty nice to me, and it seems to work well!

### DIFF
--- a/crates/composefs-oci/src/cstor.rs
+++ b/crates/composefs-oci/src/cstor.rs
@@ -1,0 +1,727 @@
+//! containers-storage integration for zero-copy layer import.
+//!
+//! This module provides functionality to import container images directly from
+//! containers-storage (as used by podman/buildah) into composefs repositories.
+//! It uses the cstorage crate to access the storage and leverages reflinks when
+//! available to avoid copying file data, enabling efficient zero-copy extraction.
+//!
+//! This module requires the `containers-storage` feature to be enabled.
+//!
+//! The main entry point is [`import_from_containers_storage`], which takes an
+//! image ID and imports all layers into the repository.
+//!
+//! # Overview
+//!
+//! When importing from containers-storage, we:
+//! 1. Open the storage and locate the image
+//! 2. For each layer, iterate through the tar-split metadata
+//! 3. For large files (> INLINE_CONTENT_MAX_V0), reflink directly to objects/
+//! 4. For small files, embed inline in the splitstream
+//! 5. Handle overlay whiteouts properly
+//!
+//! # Rootless Support
+//!
+//! When running as an unprivileged user, files in containers-storage may have
+//! restrictive permissions (e.g., `/etc/shadow` with mode 0600 owned by remapped
+//! UIDs). In this case, we spawn a helper process via `podman unshare` that can
+//! read all files, and it streams the content back to us via a Unix socket with
+//! file descriptor passing.
+//!
+//! # Example
+//!
+//! ```ignore
+//! use composefs_oci::cstor::import_from_containers_storage;
+//!
+//! let repo = Arc::new(Repository::open_user()?);
+//! let (result, stats) = import_from_containers_storage(&repo, "sha256:abc123...", None, false).await?;
+//! println!("Imported config: {}", result.0);
+//! println!("Stats: {:?}", stats);
+//! ```
+
+use std::os::unix::fs::FileExt;
+use std::os::unix::io::OwnedFd;
+use std::sync::Arc;
+
+use anyhow::{Context, Result};
+use base64::Engine;
+use indicatif::{ProgressBar, ProgressStyle};
+
+use composefs::{
+    INLINE_CONTENT_MAX_V0,
+    fsverity::FsVerityHashValue,
+    repository::{ImportContext, ObjectStoreMethod, Repository},
+};
+
+use cstorage::{
+    Image, Layer, ProxiedTarSplitItem, Storage, StorageProxy, TarSplitFdStream, TarSplitItem,
+    can_bypass_file_permissions,
+};
+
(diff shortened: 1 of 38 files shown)


crates/composefs-oci/src/cstor.rs
```diff
@@ -308,6 +402,113 @@ fn import_layer_with_writer<ObjectID: FsVerityHashValue>(
     Ok((verity, stats))
 }
 
+/// Import a single layer via the proxy (rootless mode).
+async fn import_layer_proxied<ObjectID: FsVerityHashValue>(
+    repo: &Arc<Repository<ObjectID>>,
+    proxy: &mut StorageProxy,
+    storage_path: &str,
+    layer_id: &str,
```
cgwalters: I think these should also become `oci_spec::Digest` types

crates/cstorage/src/userns.rs
```diff
@@ -0,0 +1,917 @@
...
+
+    // Fall back to getpwuid via /etc/passwd parsing
+    let uid = getuid().as_raw();
+    let passwd = etc_dir
+        .read_to_string("passwd")
+        .map_err(|e| UsernsError::CurrentUser(format!("failed to read /etc/passwd: {}", e)))?;
+
+    for line in passwd.lines() {
+        let parts: Vec<&str> = line.split(':').collect();
+        if parts.len() >= 3 {
```
cgwalters: Minor nit: let's use itertools collect_array()

cgwalters (commented):
Doing a review here so it's more visible

cgwalters:
> I've not looked into details yet, but are we comfortable with the risks of these intrusive handling of the storage library? The storage library doesn't have any stability guarantees and can change at any time

Is there anything fundamentally different from the storage access in https://github.com/bootc-dev/bootc/pull/1962 though?

As far as the "can change at any time" - in reality we have pretty strong constraints there on any truly backwards-incompatible changes right? Because even though there's one Go library for it, there's multiple tools using that library which may not always be upgraded at the same time and I definitely remember an incident with cri-o + podman on different schedules.

I think if we did any future breaking changes, we'd just need to remember to also update this library right? We can enforce that with CI.

allisonkarlitskaya:
So from what I can tell, the main thrust here is to get the reflinking thing between c-storage and composefs-rs working, right?

I'd have a really strong preference here to add a new API to the existing skopeo proxy and containers-image-store-rs.  There are already fd-passing APIs there: we'd just need a lot more fds...

@giuseppe's argument about having a parallel implementation has traction with me.  I'm not sure that I'd prefer a vibecoded containers-storage reimplementation, even if it is in Rust, compared to the real thing, specifically because we're gonna RPC to it anyway...

See also https://github.com/containers/composefs-rs/pull/137 for existing code that understands enough of tarsplit (because it's part of zstd:chunked) to get its job done...

I know you're trying to reduce CI pain here by decreasing the size of the stack of modules we have, but I'm not sure how I feel about mass-importing a vibecoding session here... the fact that most of that is in a separate crate helps a bit, of course.

Also: are you still 100% sure "pull into c-storage and then copy to composefs" is the way we definitely want to do this?  I know the arguments about exotic authentication mechanisms, but skopeo would cover us there.  I also know the arguments about zstd:chunked, but we have our own support in #137.  I also know the arguments about combining zstd:chunked with the exotic authentication mechanisms thing and that's where I have a bit of sympathy for this approach, but ... it seems like maybe we could possibly figure something out still?  Also: if this is the plan, then we'll need skopeo (at least) around anyway to do the pulling for us, so having a separate implementation of c-storage won't really gain us anything in terms of reducing on-disk size, etc...

allisonkarlitskaya:
> I suppose I always dreamed of containers-storage backending directly on to composefs.

Actually, there's a thought: we could teach skopeo a bit about how to write to composefs... the object storage part is obviously the most important bit, but it's also the easiest bit: instead of it handing us a tonne of fds that we do reflinks to we could just hand it our object directory (as an fd, or by name) and have it do that part...

cgwalters:
Argh, I may have added enough new tests here that it breaks due to capacity in GHA. Reworking the suite is a bit nontrivial, but looking

cgwalters:
This is now used by https://github.com/bootc-dev/bootc/pull/2044

cgwalters:
OK yep, the new https://nexte.st/ usage here is way nicer to read!

crates/cstorage/src/tar_split.rs
```diff
@@ -0,0 +1,711 @@
...
+                                    )
+                                })?;
+
+                                let mut file = self.open_file_in_chain(path)?;
+
+                                // Verify CRC64 if provided
+                                if let Some(ref crc64_b64) = crc64 {
+                                    self.verify_crc64(&mut file, crc64_b64, file_size)?;
+
+                                    // Reopen file since we consumed it for CRC check
```
Johan-Liebert1: we could Seek to 0?

Johan-Liebert1:
Skimmed over the code and tested locally. It seems okay to me. Currently the `containers-storage` feature is turned on by default. Maybe we should make it opt-in instead of opt-out?

crates/cstorage/src/tar_split.rs
cgwalters: Thanks, fixed!

cgwalters:
> Skimmed over the code and tested locally. It seems okay to me. Currently the containers-storage feature is turned on by default. Maybe we should make it opt-in instead of opt-out?

OK, this is now a compile-time feature (default on) but also a runtime choice (default off) for both the CLI and the library.

Johan-Liebert1 (approved):

PR #221, opened 2026-02-03
Support SHA-512 digests in composefs-setup-root

gucci-on-fleek:
After generating a composefs cmdline using `bootc container compute-composefs-digest-from-storage`, `composefs-setup-root` (compiled from the latest `main`) gives an `Expected digest size 32` error, making the system unbootable. This commit adds support for SHA-512 digests to `composefs-setup-root`.

I've never written any Rust before, so I tried a bunch of other ideas, but this was the only solution that `rustc` didn't refuse to compile. I'm sure that there's a better way of doing this than just copy-and-pasting the code block, but I couldn't figure it out. I've tested this and can confirm that it works as expected though.

See also #199.

### DIFF
--- a/crates/composefs-setup-root/src/main.rs
+++ b/crates/composefs-setup-root/src/main.rs
@@ -179,9 +179,19 @@ fn open_root_fs(path: &Path) -> Result<OwnedFd> {
 }
 
 fn mount_composefs_image(sysroot: &OwnedFd, name: &str, insecure: bool) -> Result<OwnedFd> {
-    let mut repo = Repository::<Sha256HashValue>::open_path(sysroot, "composefs")?;
-    repo.set_insecure(insecure);
-    repo.mount(name).context("Failed to mount composefs image")
+    match name.len() {
+        128 => {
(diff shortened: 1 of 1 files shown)


crates/composefs-setup-root/src/main.rs
```diff
@@ -179,9 +179,19 @@ fn open_root_fs(path: &Path) -> Result<OwnedFd> {
 }
 
 fn mount_composefs_image(sysroot: &OwnedFd, name: &str, insecure: bool) -> Result<OwnedFd> {
-    let mut repo = Repository::<Sha256HashValue>::open_path(sysroot, "composefs")?;
-    repo.set_insecure(insecure);
-    repo.mount(name).context("Failed to mount composefs image")
+    match name.len() {
+        128 => {
+            let mut repo = Repository::<Sha512HashValue>::open_path(sysroot, "composefs")?;
```
Johan-Liebert1: Looks fine to me, but the below two lines are duplicated, maybe we can move them out

PR #222, opened 2026-02-04
composefs-oci: use chunked reading in tar processing

giuseppe:
Process tar entries in 1MiB chunks to avoid large memory allocations.

### DIFF
--- a/crates/composefs-oci/src/tar.rs
+++ b/crates/composefs-oci/src/tar.rs
@@ -65,6 +65,8 @@ pub fn split(
     tar_stream: &mut impl Read,
     writer: &mut SplitStreamWriter<impl FsVerityHashValue>,
 ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


crates/composefs-oci/src/tar.rs
```diff
@@ -76,16 +78,39 @@ pub fn split(
...
+            }
+
+            let tmpfile = tmpfile.into_inner()?;
+            let object_id = writer
+                .repo()
+                .finalize_object_tmpfile(tmpfile, actual_size as u64)?;
+            writer.add_external_size(actual_size as u64);
+            writer.write_reference(object_id)?;
+
+            let padding_size = storage_size - actual_size;
```
cgwalters: might as well use e.g. `checked_sub().unwrap()` here for safety

cgwalters (approved):

PR #223, opened 2026-02-05
project: Add error contexts to functions

Johan-Liebert1:
Assisted-By: Claude Code (Opus)


jeckersb:
And since AI apparently can't execute on "turn that list into markdown" here's one more that it originally listed but didn't include in the markdown version:

```
15. In crates/composefs/src/splitstream.rs:
    - The file adds #[context()] to SplitStreamReader::new() but then in repository.rs line ~734: SplitStreamReader::new(file, expected_content_type).context("Creating split stream reader") — Double context
```

PR #224, opened 2026-02-05
doc/plans: Update OCI sealing spec (kernel sigs, flattened layers)

cgwalters:
Two big goals:

- Support for kernel-native fsverity signatures to be associated with a digest
- Generalize the "flattened vs per-layer digest"; any layer can now have either.


doc/plans/oci-sealing-spec.md
```diff
@@ -20,163 +24,200 @@ Incremental verification must be supported, enabling verification of individual
...
+```json
+{
+  "layers": [
+    {
+      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
+      "digest": "sha256:9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0",
+      "size": 32654,
+      "annotations": {
+        "composefs.layer.sha512-12": "3abb6677af34ac57c0ca5828fd94f9d886c26ce59a8ce60ecf6778079423dccff1d6f19cb655805d56098e6d38a1a710dee59523eed7511e5a9e4b8ccb3a4686",
+        "composefs.layer.sha512-12.signature": "MIIBkgYJKo...base64-encoded-pkcs7...",
```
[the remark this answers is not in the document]
cgwalters: Actually also... :thinking: if we support an external signature object (or piggy back this on a cosign object), I think we should also support a `composefs.manifest.sha512-12` and `composefs.config.sha512-12` style thing - and the awesome thing about that is we could enable signed fsverity for the manifest and config objects too.

The benefit of that is we have a really clear story where *everything* is validated, not just the tar layers.

To implement this sanely we'd need to wean composefs-rs off of using splitstream to store manifest/config...I did some prep work in this space (supporting OCI artifacts) in https://github.com/containers/composefs-rs/pull/216

doc/plans/oci-sealing-impl.md
```diff
@@ -1,210 +1,119 @@
 # OCI Sealing Implementation in composefs-rs
 
-This document describes the implementation of OCI sealing in composefs-rs. For the generic specification applicable to any composefs implementation, see [oci-sealing-spec.md](oci-sealing-spec.md).
+Implementation plan for [oci-sealing-spec.md](oci-sealing-spec.md) in composefs-rs.
 
-<!-- This document was generated primarily by Claude Sonnet 4.5. Human review level: low-medium -->
+## Current State
 
-## Current Implementation Status
+The `composefs-oci` crate has basic sealing: `seal()` in `lib.rs` computes the fsverity digest of the merged EROFS via `compute_image_id()`, stores it in the container runtime config labels under `containers.composefs.fsverity`, and `mount()` verifies it by looking up the image by digest and comparing against the kernel's fsverity measurement.
```
allisonkarlitskaya: I read that as "`mount(2)` verifies it" which doesn't do anything, of course.  Maybe make it explicit that we're talking about the one in `composefs::`.

allisonkarlitskaya (commented):
I started reviewing this before I noticed how large it was.  A few comments from the first part...

doc/plans/oci-sealing-spec.md
```diff
@@ -8,175 +8,337 @@ Container images need cryptographic verification that efficiently covers the ent
...
   ]
 }
 ```
 
-Per-layer digests enable incremental verification during pull, create caching opportunities where shared layers have known composefs digests, and enable runtime choice between flattened versus layered mounting strategies.
+Note: For the first layer, `composefs.merged.sha512-12` equals `composefs.layer.sha512-12` since a single layer's merge is itself.
+
+### Signatures
+
+#### Linux kernel fsverity signatures (recommended)
```
allisonkarlitskaya: This is a very significant departure from the way the trust model works now and how the object store works in general.  From what I understand, each file can only have a single fs-verity signature on it, but we store objects by their content hash, which means if we had two objects enter the object store from differently-signed containers, we'd be in trouble, no?

Is this for every object or just the erofs image?

I'm also not sure that kernel-level fsverity signatures provide very strong protection (at the level that would be provided by signatures on disk images, for example) because they are on a file-by-file basis, assuming that is the intent here.  If you ignore the userspace stuff, you could probably still use your ability to freely mix-and-match various individually-signed files into a system configuration that let's you do "bad things"...

cgwalters: I agree, "one accepted signature" is fine - we can't do anything different today because the Linux kernel's fsverity mechanism only allows the same.

doc/plans/standardized-erofs-meta.md
```diff
@@ -0,0 +1,74 @@
+# Standardized EROFS Metadata Serialization
```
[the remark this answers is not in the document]
hsiangkao: I don't have the answer: I could find some technical ways, yet when it comes to OCI world, it really su*ks...

doc/plans/oci-sealing-spec.md
```diff
@@ -8,175 +8,338 @@ Container images need cryptographic verification that efficiently covers the ent
...
+
+The merged EROFS represents a fully flattened filesystem and is designed to be mounted directly, not stacked with other EROFS layers via overlayfs. During the merge process, OCI whiteouts (`.wh.*` files and opaque directory markers) are fully processed: files and directories marked for deletion in upper layers are removed from the merged result. The final merged EROFS contains no whiteout entries — it is a clean, whiteout-free snapshot of the complete filesystem tree as it would appear after all layers are applied.
+
+### Signatures
+
+#### Linux kernel fsverity signatures (recommended)
+
+The primary signature mechanism is Linux kernel [fsverity built-in signature verification](https://docs.kernel.org/filesystems/fsverity.html#built-in-signature-verification). The kernel's `FS_IOC_ENABLE_VERITY` ioctl accepts a PKCS#7 signature that is verified against the `.fs-verity` keyring. This provides a clear chain of trust: the same component that controls data access (the kernel) also validates the signature. The kernel additionally integrates with the [IPE](https://docs.kernel.org/admin-guide/LSM/ipe.html) (Integrity Policy Enforcement) subsystem.
```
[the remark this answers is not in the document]
cgwalters: Yeah I think the Fedora 34 (and openSUSE Leap 15.6) kernel configuration is incoherent; they set `CONFIG_DM_VERITY_VERIFY_ROOTHASH_SIG=y` but `# CONFIG_FS_VERITY_BUILTIN_SIGNATURES is not set`. Whereas for e.g. Debian 13, it sets both. 

I understand that there's disagreement among some kernel maintainers as to whether in-kernel signatures should be used or not. But IPE and IMA also reuse the in-kernel signature infrastructure, and this proposal is explicitly targeting support for IPE.

cgwalters:
Discussed this in the meeting today; one issue with this sealing in general is "metadata amplification".

Especially if we generate per-layer EROFS. I'm looking at gathering some data on this, and also looking at whether we should e.g. zstd compress the EROFS.

cgwalters:
TL;DR - it probably is important to have a +zstd variant encoding for EROFS. And even then, the larger the image (in number of files) the more the metadata amplification bites for the single merged EROFS.

# EROFS Metadata Size Analysis

🤖 Generated-by: OpenCode (Claude Opus 4)
Reviewed-by: Colin Walters

Composefs EROFS images are metadata-only: they contain the filesystem tree
structure (inodes, directory entries, xattrs) but no file content. File content
lives in the content-addressed object store, referenced via
`trusted.overlay.metacopy` and `trusted.overlay.redirect` xattrs.

This analysis measures the EROFS metadata overhead for real container images
and evaluates whether zstd compression of the EROFS blob would be worthwhile.

All measurements were taken with `cfsctl oci erofs-info` using SHA-256 digests.
The recommended default algorithm is SHA-512, which increases per-file EROFS
overhead by ~37% (see per-inode breakdown below).

## Results

| Image | Unique files | Dirs | Symlinks | Hardlinks | Content | EROFS | zstd | Overhead |
|---|--:|--:|--:|--:|--:|--:|--:|--:|
| busybox:latest | 16 | 20 | 2 | 405 | 4.2 MiB | 16.0 KiB | 4.8 KiB | 0.37% |
| debian:bookworm | 5,241 | 778 | 642 | 2 | 111 MiB | 1.4 MiB | 521 KiB | 1.25% |
| fedora:43 | 3,861 | 2,296 | 1,644 | 849 | 173 MiB | 1.4 MiB | 418 KiB | 0.80% |
| centos-bootc:stream10 | 98,054 | 5,611 | 9,217 | 38,949 | 1.4 GiB | 18.8 MiB | 5.3 MiB | 1.28% |
| fedora-silverblue:43 | 289,339 | 11,895 | 28,570 | 114,914 | 4.9 GiB | 55.2 MiB | 15.4 MiB | 1.09% |

Overhead is uncompressed EROFS size as a percentage of total file content.

## Per-inode cost breakdown

Each external regular file in the EROFS costs roughly **256 bytes** with
SHA-256 or **352 bytes** with SHA-512. The digest size drives most of the
difference — it appears twice per file (once binary in the metacopy xattr,
once hex-encoded in the redirect xattr):

| Component | SHA-256 | SHA-512 |
|---|--:|--:|
| `ExtendedInodeHeader` | 64 | 64 |
| Xattr header + metacopy | 68 | 100 |
| Xattr header + redirect (hex path) | 88 | 152 |
| Null chunk index | 4 | 4 |
| Directory entry (~20-char name) | ~32 | ~32 |
| **Total per file** | **~256** | **~352** |

The redirect xattr is the single largest contributor — it hex-encodes the
digest into a path string, a 2x expansion of the same information already
present in binary form in the metacopy xattr. With SHA-512 this means a
130-byte hex path per file.

Global overhead (composefs header + superblock) is a fixed 1,152 bytes.

## Compression characteristics

zstd compression achieves a consistent **2.7–3.6:1** ratio across all tested
images. The EROFS metadata compresses well because xattr names
(`trusted.overlay.metacopy`, `trusted.overlay.redirect`) repeat on every
external file, inode headers have a fixed structure, and hex-encoded digests
use only 16 distinct characters. Larger images compress slightly better due to
more repetition.

With zstd, metadata overhead drops to **0.1–0.5%** of content size.

## Should we zstd-compress the EROFS?

EROFS supports internal compression (zstd, lz4) of file data blocks. This is
structurally inapplicable to composefs: composefs EROFS images contain no file
data blocks. All non-trivial file content is stored externally in the
content-addressed object store, referenced by fsverity digest. The only data
inside the EROFS is structural metadata — inodes, directory entries, and
xattrs. EROFS internal compression operates on data blocks, not on metadata
structures, so there is simply nothing for it to compress.

The composefs EROFS reader enforces this: it validates
`available_compr_algs == 0` and rejects any image that claims compression
support. This isn't a limitation — it reflects the fact that there's nothing to
compress inside the image.

The real question is whether the EROFS blob should be compressed **on the
wire**. In OCI, compression is not automatic — it's declared by the media type
on the layer descriptor, and the blob is pushed already compressed. For
example, tar layers use `application/vnd.oci.image.layer.v1.tar+gzip` or
`+zstd` to indicate the blob is gzip'd or zstd'd. The current sealing spec
defines `application/vnd.composefs.v1.erofs` with no compression suffix,
meaning the EROFS blob travels uncompressed.

To get wire compression, we would need to define compressed media type
variants: `application/vnd.composefs.v1.erofs+zstd` (and/or `+gzip`),
following the same convention as tar layers. Clients would push the
zstd-compressed EROFS blob and declare it with the `+zstd` media type;
consumers would decompress before storing and mounting locally. This is
straightforward but does require updating the sealing spec and having both
producers and consumers handle the compressed variant.

Whether it's worth doing depends on the use case — see the
chunkah/incremental-pull analysis below.

If the on-disk metadata size itself becomes a concern for very large images, a
more productive approach would be to reduce the per-file overhead (e.g.,
binary-encode the redirect path instead of hex, or merge the metacopy and
redirect xattrs into a single composefs-native xattr).

## The chunkah/incremental-pull case changes the calculus

The analysis above assumes the traditional OCI pull model: download N layers as
complete tarballs, where the EROFS metadata is a rounding error on multi-GB
layer transfers.

With [chunkah](https://github.com/coreos/chunkah) and the composefs
incremental pull design (see `doc/plans/incremental-pulls.md` and
`doc/plans/oci-sealing-spec.md`), the model shifts to content-addressed
file-level deduplication. In an update scenario (pulling version N+1 when
version N is already local), the client already has most of the content objects
and only needs to fetch the files that actually changed. The delta can be very
small — a minor package update on a 5 GiB desktop image might change only
10–50 MiB of file content.

In this scenario, the EROFS metadata becomes a significant fraction of the
transfer:

| Scenario | Content delta | EROFS (merged) | EROFS as % of delta |
|---|--:|--:|--:|
| silverblue minor update (~1% files changed) | ~50 MiB | 55.2 MiB | **110%** |
| silverblue minor update (zstd EROFS) | ~50 MiB | 15.4 MiB | **31%** |
| centos-bootc minor update (~2% files changed) | ~30 MiB | 18.8 MiB | **63%** |
| centos-bootc minor update (zstd EROFS) | ~30 MiB | 5.3 MiB | **18%** |

The EROFS metadata for the full merged image must be transferred on every
update because it describes the complete filesystem tree, not just the delta.
For incremental pulls, the metadata *is* the table of contents — the client
reads it to discover which content objects it's missing. So it's fetched early
and in full.

The "zstd" rows show what wire transfer would look like if we defined a
compressed media type variant (e.g. `application/vnd.composefs.v1.erofs+zstd`).
Without that, the full uncompressed EROFS is what travels on the wire today.

**The real lever for incremental pulls is per-layer EROFS, not just merged.**
The sealing spec defines both per-layer and merged EROFS entries in the
composefs artifact. With chunkah's content-based layers, each layer is smaller
and more stable across versions. The per-layer EROFS entries are individually
small, and a layer that didn't change has an identical EROFS that doesn't need
re-fetching at all — the client already has it. Only the EROFS entries for
changed layers need to be transferred.

The merged EROFS is the expensive case: it changes on every update regardless
of how small the delta is, because it represents the full flattened tree. For a
silverblue-sized image that's 55 MiB uncompressed / 15 MiB compressed, which
is non-trivial when the content delta is only 50 MiB. Whether to ship a merged
EROFS or have the client construct it locally from per-layer EROFS entries is a
design tradeoff: the merged EROFS saves the client from needing to do an
overlayfs-style merge, but it costs a full re-transfer on every update.

In summary: the EROFS metadata overhead is negligible for full image pulls but
becomes a meaningful cost for incremental updates. Defining a compressed media
type variant (`application/vnd.composefs.v1.erofs+zstd`) in the sealing spec
would bring the 55 MiB merged silverblue EROFS down to ~15 MiB on the wire.
Per-layer EROFS combined with chunkah's stable content-based layers minimizes
re-transfer further by avoiding re-fetch of unchanged layers' EROFS entirely.
The tooling added here (`cfsctl oci erofs-info`) allows measuring these costs
for specific images.

cgwalters:
OK, I completely rewrote this again. (Kind of tempting to make a new PR)

I think since we now have the v1 support merged - let's just go all in on that as our "standard erofs". I completely dropped support for the separate OCI artifact with EROFS.

We now support a metadata-only artifact as well as inlining the data into the manifest. I think the latter will be easy enough for people to adopt (e.g. once we ship the varlink support, having `podman build --composefs-metadata=inline` and `--composefs-sign` will be quite viable.

mike-sul:
I'm designing a flow that uses composefs + a content-addressed object store (CAS) as the on-disk representation of container image rootfs for IoT/edge devices.

One requirement is image "sealing": at publish time, compute the fs-verity digest of the composefs/EROFS metadata image generated from the image’s fully merged rootfs, and bind that digest to OCI image metadata.

IIUC, for inline mode this PR moves the merged-rootfs digest from [the earlier config-label model](https://github.com/composefs/composefs-rs/blob/main/doc/plans/oci-sealing-spec.md#composefs-digest-storage) to a top-level annotation on the image manifest, for example:

```text
composefs.merged.erofs.v1.fsverity-sha256-12=<fsverity-digest-of-the-erofs-metadata-image>
```

Is that correct?

Also, is this annotation key shape stable enough for new tooling to build against, or would you recommend waiting for the PR to settle/merge?

cgwalters:
> I'm designing a flow that uses composefs + a content-addressed object store (CAS) as the on-disk representation of container image rootfs for IoT/edge devices.

Is your project in Rust? WDYT about the goal of using cfsctl via varlink as a primary API? https://github.com/composefs/composefs-rs/issues/322

Also, for the bootable host, have you looked at https://github.com/bootc-dev/bootc/ ? Would https://github.com/bootc-dev/bootc/issues/2289 help?

> One requirement is image "sealing": at publish time, compute the fs-verity digest of the composefs/EROFS metadata image generated from the image’s fully merged rootfs, and bind that digest to OCI image metadata.

Right now for bootc (and is supported here) we handle this via a kernel commandline in a signed UKI, see https://github.com/bootc-dev/bootc/blob/main/docs/src/experimental-composefs.md#how-sealed-images-work - were you interested in doing this host root via fsverity signatures?

> IIUC, for inline mode this PR moves the merged-rootfs digest from [the earlier config-label model](https://github.com/composefs/composefs-rs/blob/main/doc/plans/oci-sealing-spec.md#composefs-digest-storage) to a top-level annotation on the image manifest, for example:

Right.

> Also, is this annotation key shape stable enough for new tooling to build against, or would you recommend waiting for the PR to settle/merge?

Well, https://github.com/composefs/composefs-rs/pull/226 has a lot of pending work that tries to implement the spec.

Were you planning a new/different implementation?

I will try to get more consensus on this spec soon across stakeholders, but I personally think what we have now is viable.

I've updated this PR.

crates/composefs-oci/src/incremental_pulls_spec.rs
```diff
@@ -0,0 +1,135 @@
...
+//!
+//! The zstd frame boundaries are an overlay on top of the canonical ordering. For files above a size threshold (e.g. 4 KiB), the compressor closes and restarts the zstd frame around the file's payload, making it independently decompressible. Files below the threshold are simply compressed together with their neighbors in whatever order they naturally appear. The threshold aligns with the filesystem block size.
+//!
+//! Files ≤ 64 bytes are already inline in the EROFS metadata (`INLINE_CONTENT_MAX`) and are never fetched from the tar layer during an incremental pull, regardless of framing.
+//!
+//! Unlike zstd:chunked, there are no trailing skippable frames (no embedded JSON TOC, no tar-split data). The offset map in the composefs metadata artifact serves as the TOC; the EROFS itself is generated locally by the client.
+//!
+//! Unlike zstd:chunked, there is no sub-file content-defined chunking. Composefs deduplicates at the whole-file level (by fsverity digest), so rolling-checksum chunk boundaries provide no dedup benefit. This simplifies the format and the offset map.
```
giuseppe: zstd:chunked uses sub-file content chunking to save on network traffic.

crates/composefs-oci/src/incremental_pulls_spec.rs
```diff
@@ -0,0 +1,135 @@
...
+//! | Content digests | SHA-256 | SHA-256 | fsverity (SHA-256 or SHA-512 Merkle tree) |
+//! | Dedup granularity | Sub-file chunks (~64 KiB, rolling checksum) | Per-file | Whole files (by fsverity digest) |
+//! | Kernel integration | None (userspace only) | None (userspace only) | EROFS + overlayfs + fsverity |
+//! | Push after incremental pull | Reconstruct via tar-split | Cannot reconstruct original tar | Canonical tar generation (see below) |
+//!
+//! ## Design
+//!
+//! ### Layer Format: composefs-chunked
+//!
+//! A composefs-chunked layer is a valid `tar+zstd` blob that any OCI client can pull and decompress normally. The difference is in how the zstd compression is structured internally: large files are compressed as independent zstd frames, making them individually addressable via byte offset.
```
giuseppe: do we really need to define a new format instead of building on top of what already exists?

What is missing in zstd:chunked to achieve this except the fs-verity digest?

crates/composefs-oci/src/incremental_pulls_spec.rs
```diff
@@ -0,0 +1,135 @@
...
+//!
+//! Tar entries are in **canonical order**, the same deterministic ordering defined by the [canonical tar format](crate::canonical_tar_spec). This is essential: a client that does an incremental pull must be able to regenerate byte-identical tar for push, so the entry ordering cannot be compression-driven.
+//!
+//! The zstd frame boundaries are an overlay on top of the canonical ordering. For files above a size threshold (e.g. 4 KiB), the compressor closes and restarts the zstd frame around the file's payload, making it independently decompressible. Files below the threshold are simply compressed together with their neighbors in whatever order they naturally appear. The threshold aligns with the filesystem block size.
+//!
+//! Files ≤ 64 bytes are already inline in the EROFS metadata (`INLINE_CONTENT_MAX`) and are never fetched from the tar layer during an incremental pull, regardless of framing.
+//!
+//! Unlike zstd:chunked, there are no trailing skippable frames (no embedded JSON TOC, no tar-split data). The offset map in the composefs metadata artifact serves as the TOC; the EROFS itself is generated locally by the client.
```
giuseppe: I prefer this to be self-contained and do not require an external artifact for each layer (doubling the number of objects on the registry compared to what we have today).

So what can we do to fix zstd:chunked to address this use case as well?

giuseppe (commented):
left some comments, I think it is better to fix/reuse zstd:chunked so it works with this use case as well, since it is already implemented and supported in containers tools.  Having yet another format seems unnecessary and more difficult to maintain.

Personally, I am not a big fan of requiring artifacts, I'd prefer layers to be self contained (even if the additional information is stored in the compression layer).

We can change zstd:chunked to generate always the same stream so we can drop the tarsplit part.  The missing part there was the canonical representation, that now it is possible to achieve.

cgwalters:
> left some comments, I think it is better to fix/reuse zstd:chunked so it works with this use case as well, since it is already implemented and supported in containers tools. Having yet another format seems unnecessary and more difficult to maintain.

Thanks for the review! In retrospect, I think I made a big mistake in trying to pivot this proposal to the "erofs as OCI artifact" - *that's* what led into the whole side realization that maybe we could do something zstd:chunked like using the fact that we now had an EROFS etc.

To recap, the rationale for EROFS-as-artifact was to avoid trying to standardize exactly how we generate the EROFS layout. But we now have that - we have two independent implementations of the composefs V1 EROFS (including one here) and we are now standardizing on that.

That means we can go back to focusing on something *much* simpler: how we ship the digest and fsverity signatures and *that's it*.

I think most of your comments here now apply to https://github.com/composefs/composefs-rs/pull/356 - but yes...from my PoV it is not important to have a new chunking implementation at all, and trying to scope that in here was again a big mistake!

---

> Personally, I am not a big fan of requiring artifacts, I'd prefer layers to be self contained (even if the additional information is stored in the compression layer).

My intuition here is that the dividing line here is: Including the *digest* inline in the image is conceptually straightforward - there's implementation details (like fsverity digest caching) to consider, but at some point I think `podman build` etc should just start including the digest.

The signature is harder because:

- There are good operational reasons for wanting to split up the build process from a signing service
- It's also useful to be able to add this data to images that were built before the spec existed (this applies to the digest too of course)

giuseppe (approved):
LGTM

mike-sul:
> I will try to get more consensus on this spec soon across stakeholders, but I personally think what we have now is viable.
> 
> I've updated this PR.

@cgwalters Thanks, that helps.

> 
> Is your project in Rust? WDYT about the goal of using cfsctl via varlink as a primary API? https://github.com/composefs/composefs-rs/issues/322
> 
> Also, for the bootable host, have you looked at https://github.com/bootc-dev/bootc/ ? Would https://github.com/bootc-dev/bootc/issues/2289 help?

The project is being written in Go and integrates with the "dockerd->containerd" path as a remote snapshotter. It targets regular application container images on IoT/edge devices, not bootable/system container images that bootc is already focused on.

On cfsctl/varlink: I’d use it today if it covered what I need, but it does not yet. The cfsctl/API is pre-stable, and sealing does not appear to be exposed end to end: `Pull` and `ComputeId` exist, but annotating the manifest and publishing the sealed manifest/index does not.

> Were you planning a new/different implementation?

So for the MVP, I’ve been building on existing composefs userspace tooling such as `mkcomposefs --from-file` and the dump format, while keeping the semantics aligned with composefs-rs. Once the varlink API stabilizes, I’d prefer to replace custom code with upstream functionality for the parts it cleanly covers. The less composefs-specific code I own, the better. If it helps, I’m happy to write up the concrete gaps against our use cases.

I guess, there is one commitment regardless: reproducibility across implementations is central to this spec, so I’ll use `Oci.ComputeId` as a conformance check: the same inputs and EROFS format version should produce the same merged image ID. I’ll report anything that diverges.

cgwalters:
> On cfsctl/varlink: I’d use it today if it covered what I need, but it does not yet. The cfsctl/API is pre-stable, and sealing does not appear to be exposed end to end: Pull and ComputeId exist, but annotating the manifest and publishing the sealed manifest/index does not.

Look for more work to appear in https://github.com/composefs/composefs-rs/pull/226 soon!

> So for the MVP, I’ve been building on existing composefs userspace tooling such as mkcomposefs --from-file and the dump format, while keeping the semantics aligned with composefs-rs.

OK. There are some nontrivial details here to consider though, such as the EROFS digest depending also on the underlying file paths. We *do* have integration test coverage that there's compatibility with the original C `mkcomposefs` (and there's a brand new Rust implementation in this repo now) but it is certainly within possibility that there's something missed here.

What would probably be good is basically if your tooling is FOSS - post it. Or if it's not, at least please do try to evaluate compatibility vs what should hopefully soon land in this repo.

>  Once the varlink API stabilizes, I’d prefer to replace custom code with upstream functionality for the parts it cleanly covers. The less composefs-specific code I own, the better. If it helps, I’m happy to write up the concrete gaps against our use cases.

Yes please do!

mike-sul:
@cgwalters Thanks. I’ll follow the work in #226 closely and evaluate compatibility with it as the implementation lands.

> What would probably be good is basically if your tooling is FOSS - post it. Or if it's not, at least please do try to evaluate compatibility vs what should hopefully soon land in this repo.

The intent is for the tooling to be FOSS, but I still need to initiate and complete our internal open-source approval process before I can publish the repository.

I checked the convention used by composefs-rs: the relative backing-object path is derived from the file’s fsverity digest. I’ll use the same convention and compare the result with `mkcomposefs`, the rust implementation, and Oci.ComputeId and report any mismatch.



   

>> `Once the varlink API stabilizes, I’d prefer to replace custom code with upstream functionality for the parts it cleanly covers. The less composefs-specific code I own, the better. If it helps, I’m happy to write up the concrete gaps against our use cases.`

> Yes please do!

Is issue #322 the best place to report these gaps?

mike-sul:
@cgwalters [This](https://github.com/composefs/composefs-rs/issues/358) is the key gap I've discovered so far.

mike-sul:
> We _do_ have integration test coverage that there's compatibility with the original C `mkcomposefs` (and there's a brand new Rust implementation in this repo now) but it is certainly within possibility that there's something missed here.

@cgwalters While evaluating cross-implementation compatibility I found two inconsistency between the C tools and this repo (I reported them on the C side):

1. composefs/composefs#442: `composefs-info measure-file` reports a digest the kernel can never produce for empty files; cfsctl's reimplementation matches the kernel.

- composefs/composefs#444: `mkcomposefs <dir> <out>` loses hardlinks, so the same tree yields a different image digest than `cfsctl` (or `mkcomposefs --from-file`) produces.

PR #225, opened 2026-02-06
Initial reimplementation of composefs-c

cgwalters:
Basically starting on https://github.com/composefs/composefs/discussions/423

3 key goals:

- Compatible CLI interfaces
- Compatible EROFS output format (this is a big deal!)
- Next: Compatible C shared library (ugly and messy)

Assisted-by: OpenCode (Claude Sonnet 4)

### DIFF
--- a/crates/composefs-ctl/src/composefs_info.rs
+++ b/crates/composefs-ctl/src/composefs_info.rs
@@ -1,0 +1,318 @@
+//! composefs-info - Query information from composefs images.
+//!
+//! This is a Rust reimplementation of the C composefs-info tool, providing
+//! commands to inspect EROFS images, list objects, and compute fs-verity digests.
+//!
+//! ## Compatibility status
+//!
+//! Implemented subcommands:
+//! - `ls` — lists files with type suffixes, skips whiteout entries
+//! - `dump` — outputs composefs-dump(5) text format (image → tree → dumpfile)
+//! - `objects` — lists all backing file object paths (XX/XXXX...)
+//! - `missing-objects` — lists objects not present in `--basedir`
+//! - `measure-file` — computes fs-verity digest of files
+//!
+//! Compatibility notes:
+//! - `measure-file` tries `FS_IOC_MEASURE_VERITY` first; falls back to
+//!   in-process computation when the kernel reports verity is absent or the
+//!   filesystem doesn't support it, matching the C `lcfs_fd_get_fsverity()`
+//!   behaviour.
+
+use std::collections::HashSet;
+use std::io::Write;
+use std::path::Path;
+use std::{fs::File, io::Read, path::PathBuf};
+
+use anyhow::{Context, Result};
+use clap::{Parser, Subcommand};
+
+use composefs::{
+    dumpfile::write_dumpfile,
+    erofs::reader::erofs_to_filesystem,
+    fsverity::{FsVerityHashValue, Sha256HashValue, measure_verity_with_fallback},
+    generic_tree::{Inode, LeafContent, LeafId},
+    tree::{FileSystem, RegularFile},
+};
+
+/// Query information from composefs images.
+#[derive(Parser, Debug)]
+#[command(
+    name = "composefs-info",
+    version,
+    about = "Query information from composefs images"
+)]
+struct Cli {
+    /// The subcommand to run.
+    #[command(subcommand)]
+    command: Command,
+}
+
+/// Available subcommands.
+#[derive(Subcommand, Debug)]
+enum Command {
+    /// Simple listing of files and directories in the image.
+    Ls {
+        /// Filter entries at the root level by name (can be specified multiple times).
+        #[arg(long = "filter", action = clap::ArgAction::Append)]
+        filter: Vec<String>,
+        /// Composefs image files to inspect.
+        images: Vec<PathBuf>,
+    },
+
+    /// Full dump in composefs-dump(5) format.
+    Dump {
+        /// Filter entries at the root level by name (can be specified multiple times).
+        #[arg(long = "filter", action = clap::ArgAction::Append)]
+        filter: Vec<String>,
+        /// Composefs image files to dump.
+        images: Vec<PathBuf>,
+    },
+
+    /// List all backing file object paths.
+    Objects {
+        /// Composefs image files to inspect.
+        images: Vec<PathBuf>,
+    },
+
+    /// List backing files not present in basedir.
+    MissingObjects {
+        /// Base directory for object lookups.
+        #[arg(long = "basedir", required = true)]
+        basedir: PathBuf,
+        /// Composefs image files to inspect.
+        images: Vec<PathBuf>,
+    },
+
+    /// Print the fs-verity digest of files.
+    MeasureFile {
+        /// Files to measure.
+        files: Vec<PathBuf>,
+    },
+}
+
+/// Entry point for the composefs-info multi-call mode.
+pub fn run() -> Result<()> {
+    let cli = Cli::parse();
+    run_with_cli(cli)
+}
+
+/// Entry point when invoked as a hidden `cfsctl composefs-info` subcommand.
+///
+/// `extra_args` contains everything after the `composefs-info` token as
+/// captured by clap's trailing-var-arg mechanism.  A synthetic `argv[0]` is
+/// prepended so that `Cli::parse_from` produces the same help text / error
+/// messages as the standalone binary.
+pub fn run_from_args(extra_args: Vec<std::ffi::OsString>) -> Result<()> {
+    let mut argv = vec![std::ffi::OsString::from("composefs-info")];
+    argv.extend(extra_args);
+    let cli = Cli::parse_from(argv);
+    run_with_cli(cli)
+}
+
+fn run_with_cli(cli: Cli) -> Result<()> {
+    match &cli.command {
+        Command::Ls { filter, images } => cmd_ls(filter, images),
+        Command::Dump { filter, images } => cmd_dump(filter, images),
+        Command::Objects { images } => cmd_objects(images),
+        Command::MissingObjects { basedir, images } => cmd_missing_objects(basedir, images),
+        Command::MeasureFile { files } => cmd_measure_file(files),
+    }
+}
+
+/// Print escaped path (matches C implementation behavior).
+fn print_escaped<W: Write>(out: &mut W, s: &[u8]) -> std::io::Result<()> {
+    for &c in s {
+        match c {
+            b'\\' => write!(out, "\\\\")?,
+            b'\n' => write!(out, "\\n")?,
+            b'\r' => write!(out, "\\r")?,
+            b'\t' => write!(out, "\\t")?,
+            // Non-printable or non-ASCII characters are hex-escaped
+            c if !c.is_ascii_graphic() && c != b' ' => write!(out, "\\x{c:02x}")?,
+            c => out.write_all(&[c])?,
+        }
+    }
+    Ok(())
+}
+
+/// Walk and print entries: directory line first, then recurse into children.
+fn ls_print<W: Write>(
+    out: &mut W,
+    fs: &FileSystem<Sha256HashValue>,
+    dir: &composefs::tree::Directory<Sha256HashValue>,
+    path: &[u8],
+    seen_leaf_ids: &mut HashSet<LeafId>,
+    filter: Option<&[String]>,
+) -> Result<()> {
+    for (name, child) in dir.sorted_entries() {
+        let name_bytes = name.as_encoded_bytes();
+
+        if let Some(filter) = filter
+            && !filter.is_empty()
+        {
+            let name_str = name.to_string_lossy();
+            if !filter.iter().any(|f| f == name_str.as_ref()) {
+                continue;
+            }
+        }
+
+        let mut child_path = path.to_vec();
+        child_path.push(b'/');
+        child_path.extend_from_slice(name_bytes);
+
+        match child {
+            Inode::Directory(child_dir) => {
+                // Print the directory entry with trailing slash.
+                print_escaped(out, &child_path)?;
+                write!(out, "/\t")?;
+                writeln!(out)?;
+                // Recurse into the directory.
+                ls_print(out, fs, child_dir, &child_path, seen_leaf_ids, None)?;
+            }
+            Inode::Leaf(leaf_id, _) => {
+                let leaf = fs.leaf(*leaf_id);
+
+                print_escaped(out, &child_path)?;
+
+                match &leaf.content {
+                    LeafContent::Regular(regular) => {
+                        let is_hardlink = !seen_leaf_ids.insert(*leaf_id);
+                        if !is_hardlink && let RegularFile::External(id, _) = regular {
+                            write!(out, "\t@ ")?;
+                            print_escaped(out, id.to_object_pathname().as_bytes())?;
+                        }
+                    }
+                    LeafContent::Symlink(target) => {
+                        write!(out, "\t-> ")?;
+                        print_escaped(out, target.as_encoded_bytes())?;
+                    }
+                    _ => {}
+                }
+
+                writeln!(out)?;
+            }
+        }
+    }
+    Ok(())
+}
+
+/// List files and directories in the image.
+fn cmd_ls(filter: &[String], images: &[PathBuf]) -> Result<()> {
+    let stdout = std::io::stdout();
+    let mut out = stdout.lock();
+
+    for image_path in images {
+        let image_data = read_image(image_path)?;
+        let fs = erofs_to_filesystem::<Sha256HashValue>(&image_data)
+            .with_context(|| format!("Failed to parse image: {image_path:?}"))?;
+
+        let mut seen_leaf_ids = HashSet::new();
+        ls_print(
+            &mut out,
+            &fs,
+            &fs.root,
+            b"",
+            &mut seen_leaf_ids,
+            Some(filter),
+        )?;
+    }
+
+    Ok(())
+}
+
+/// Dump the image in composefs-dump(5) text format.
+///
+/// This matches the C composefs-info dump output: the EROFS image is parsed
+/// back into a filesystem tree which is then serialized as a dumpfile.
+fn cmd_dump(_filter: &[String], images: &[PathBuf]) -> Result<()> {
+    let stdout = std::io::stdout();
+    let mut out = stdout.lock();
+
+    for image_path in images {
+        let image_data = read_image(image_path)?;
+        let fs = erofs_to_filesystem::<Sha256HashValue>(&image_data)
+            .with_context(|| format!("Failed to parse image: {image_path:?}"))?;
+        write_dumpfile(&mut out, &fs)
+            .with_context(|| format!("Failed to dump image: {image_path:?}"))?;
+    }
+
+    Ok(())
+}
+
+/// Collect all external object IDs from a parsed filesystem.
+///
+/// Iterates the leaves table directly — each `RegularFile::External` entry
+/// is a unique content-addressed object.  Because `erofs_to_filesystem`
+/// deduplicates hard-linked inodes into a single leaf, each object appears
+/// exactly once even if it is referenced by multiple paths.
+fn collect_objects_from_fs(fs: &FileSystem<Sha256HashValue>) -> HashSet<Sha256HashValue> {
+    fs.leaves
+        .iter()
+        .filter_map(|leaf| match &leaf.content {
+            LeafContent::Regular(RegularFile::External(id, _)) => Some(id.clone()),
+            _ => None,
+        })
+        .collect()
+}
+
+/// List all object paths from the images.
+fn cmd_objects(images: &[PathBuf]) -> Result<()> {
+    for image_path in images {
+        let image_data = read_image(image_path)?;
+        let fs = erofs_to_filesystem::<Sha256HashValue>(&image_data)
+            .with_context(|| format!("Failed to parse image: {image_path:?}"))?;
+
+        let mut objects: Vec<Sha256HashValue> = collect_objects_from_fs(&fs).into_iter().collect();
+        objects.sort_by_key(|id| id.to_hex());
+
+        for obj in objects {
+            println!("{}", obj.to_object_pathname());
+        }
+    }
+    Ok(())
+}
+
+/// List objects not present in basedir.
+fn cmd_missing_objects(basedir: &Path, images: &[PathBuf]) -> Result<()> {
+    let mut all_objects: HashSet<Sha256HashValue> = HashSet::new();
+
+    for image_path in images {
+        let image_data = read_image(image_path)?;
+        let fs = erofs_to_filesystem::<Sha256HashValue>(&image_data)
+            .with_context(|| format!("Failed to parse image: {image_path:?}"))?;
+        all_objects.extend(collect_objects_from_fs(&fs));
+    }
+
+    let mut missing: Vec<Sha256HashValue> = all_objects
+        .into_iter()
+        .filter(|obj| !basedir.join(obj.to_object_pathname()).exists())
+        .collect();
+
+    missing.sort_by_key(|a| a.to_hex());
+
+    for obj in missing {
+        println!("{}", obj.to_object_pathname());
+    }
+
+    Ok(())
+}
+
+/// Compute and print the fs-verity digest of each file.
+fn cmd_measure_file(files: &[PathBuf]) -> Result<()> {
+    for path in files {
+        let file = File::open(path).with_context(|| format!("Failed to open file: {path:?}"))?;
+        let digest = measure_verity_with_fallback::<Sha256HashValue>(file)
+            .with_context(|| format!("Failed to measure verity for {path:?}"))?;
+        println!("{}", digest.to_hex());
+    }
+    Ok(())
+}
+
+/// Read an entire image file into memory.
+fn read_image(path: &PathBuf) -> Result<Vec<u8>> {
+    let mut file = File::open(path).with_context(|| format!("Failed to open image: {path:?}"))?;
+    let mut data = Vec::new();
+    file.read_to_end(&mut data)
+        .with_context(|| format!("Failed to read image: {path:?}"))?;
+    Ok(data)
+}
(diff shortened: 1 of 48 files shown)


cgwalters:
https://github.com/bootc-dev/bootc/issues/1812 is related to this - we need to fix how we generate the EROFS. The problem in a nutshell is that https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9ed50b8231e37b1ae863f5dec8153b98d9f389b4 fixed a legitimate bug, but it also made the kernel parser less strict than it needed to be to fix the bug; we now accept symlinks that cross a block boundary, whereas before we didn't.

cgwalters:
OK, this is obviously a giant PR, but I think it's getting closer into shape - the much extended proptest found lots of corner cases and we now have pretty robust parity with composefs-c as far as the V1 (original) EROFS layout.

That said, when I was thinking about applying this to bootc I realized we have basically an upgrade compatibility hazard - `bootc container ukify` has to match what the client generates, so we need to make it an opt-in to generate v1. And that has a cascading set of design implications.

I'll attach some LLM-assisted info on this.

<details>
# Design: Dual-format EROFS generation (V1 + V2) with bootc UKI integration

🤖 Assisted-by: OpenCode (Claude Sonnet 4.6)

## Background

composefs-rs added V1 EROFS format support (compatible with C `mkcomposefs`/`composefs-info`
1.0.8) alongside the existing V2 (composefs-rs native) format. The `meta.json` `erofs_version`
field and `cfs_erofs_version` ro_compat flag allow repositories to signal their preferred format
to tools.

The problem this design addresses: bootc uses composefs for sealed UKIs, where the
`composefs=<digest>` kernel argument is baked into the UKI at signing time. Because V1 and V2
EROFS images of the same content have different fs-verity digests, a sealed UKI tied to a V2
digest will fail to boot if only a V1 image is present (and vice versa). We need a principled
way to:

- Default new repositories to V1-only (for C tooling compatibility).
- Let bootc opt into generating both V1 and V2 for every image, so existing sealed V2 UKIs
  continue to boot while new UKIs can use V1.
- Signal which EROFS format a UKI expects via the kernel command line, so the initramfs can
  find the right image.

## `meta.json` changes

A new ro_compat feature flag `"v1_erofs"` is introduced alongside the existing
`"cfs_erofs_version"`.

```json
{
  "version": 1,
  "algorithm": "fsverity-sha256-12",
  "erofs_version": "V1",
  "features": {
    "compatible": [],
    "read-only-compatible": ["cfs_erofs_version", "v1_erofs"],
    "incompatible": []
  }
}
```

Semantics of `"v1_erofs"` in ro_compat:

- **Present** (default for all new repos): only V1 EROFS images are generated. Old tools that
  do not recognize this flag open the repository read-only, which is correct — a repo in
  dual-format mode should not be written by tools that don't understand both formats.
- **Absent**: both V1 and V2 EROFS images are generated for every image commit. This is the
  mode bootc uses.

Relationship to `"cfs_erofs_version"`: that flag tells old C tools "the repo's default image
format is V1". `"v1_erofs"` tells new composefs-rs tools "generate only V1, not both". They
are complementary and can coexist.

`known_features::RO_COMPAT` gains `"v1_erofs"` so the unknown-flag check does not reject repos
that have it.

## `FormatSet` type

A new type expressing which EROFS format versions to generate for an image:

```rust
// In crates/composefs/src/erofs/format.rs (or repository.rs)
pub struct FormatSet(/* bitset over FormatVersion */);

impl FormatSet {
    pub const V1_ONLY: FormatSet = /* {V1} */;
    pub const BOTH: FormatSet = /* {V1, V2} */;

    pub fn contains(&self, v: FormatVersion) -> bool { ... }
}
```

The repository's **default `FormatSet`** is derived at open time from `meta.json`:

- `"v1_erofs"` present in ro_compat → `FormatSet::V1_ONLY`
- `"v1_erofs"` absent → `FormatSet::BOTH`

Exposed as `Repository::default_format_set() -> FormatSet`.

## `RepositoryConfig` changes

`RepositoryConfig` gains a field controlling whether `"v1_erofs"` is written at init time:

```rust
pub struct RepositoryConfig {
    pub algorithm: Algorithm,
    pub erofs_version: FormatVersion,
    pub erofs_formats: FormatSet,  // new; default FormatSet::V1_ONLY
    insecure: bool,
}
```

`init_path` writes `"v1_erofs"` into ro_compat when `erofs_formats == FormatSet::V1_ONLY`,
and omits it when `erofs_formats == FormatSet::BOTH`.

## Image generation API

`FileSystem::commit_image` currently generates a single EROFS image using the repo's
`erofs_version`. It is supplemented by a new method:

```rust
impl<ObjectID: FsVerityHashValue> FileSystem {
    /// Generate EROFS images for each version in `formats`, store them in `repo`,
    /// and return a map from FormatVersion to the stored object ID.
    pub fn commit_images(
        &self,
        repo: &Repository<ObjectID>,
        ref_name: Option<&str>,
        formats: FormatSet,
    ) -> Result<HashMap<FormatVersion, ObjectID>>;
}
```

The existing `commit_image` becomes a thin wrapper calling `commit_images` with
`FormatSet::from(repo.erofs_version())` (a single-version set), for callers that only need one
format and do not need to be format-set-aware.

`generate_boot_image` in `composefs-oci` gains a `FormatSet` parameter. When not specified by
the caller, it defaults to `repo.default_format_set()`.

## Splitstream named refs

Four constants in `composefs-oci/src/lib.rs`; the existing two are unchanged:

```rust
pub const IMAGE_REF_KEY: &str = "composefs.image";          // V2, existing
pub const IMAGE_REF_KEY_V1: &str = "composefs.image.v1";    // V1, new
pub const BOOT_IMAGE_REF_KEY: &str = "composefs.image.boot";      // V2, existing
pub const BOOT_IMAGE_REF_KEY_V1: &str = "composefs.image.boot.v1"; // V1, new
```

When `FormatSet` includes V1, the V1 digest is stored under the `.v1` keys. When it includes
V2, the existing keys are populated. When both, all four are written. GC follows all populated
named refs, so both EROFS objects remain alive for as long as the config splitstream is
reachable from a tag.

## Kernel argument scheme

Two karg names; the initramfs handles both:

| Karg | Format | Notes |
|------|--------|-------|
| `composefs=<digest>` | V2 | Existing; no change to meaning. All currently sealed UKIs use this. |
| `composefs.digest=<digest>` | V1 | New; used by `bootc container ukify --format=1`. |

`ComposefsCmdline` in `composefs-boot/src/cmdline.rs` becomes an enum:

```rust
pub enum ComposefsCmdline {
    /// Legacy V2 karg: `composefs=<digest>`
    V2 { digest: Hex },
    /// V1 karg: `composefs.digest=<digest>`
    V1 { digest: Hex },
}
```

Both variants are generated and parsed by `composefs-boot`. Since bootc maintains its own copy
of the initramfs karg parsing in `bootc/crates/initramfs/src/lib.rs`, the same logic must be
duplicated there. Both files carry a comment noting they must be kept in sync; the long-term
remedy is for bootc to consume `composefs-boot` directly rather than duplicating.

The initramfs logic at boot time:

1. Check for `composefs.digest=` → look up V1 image in `composefs/images/<digest>`, mount it.
2. Fall back to `composefs=` → look up V2 image, mount it (existing behaviour).

## `cfsctl init` CLI

`cfsctl init` gains an `--erofs` option:

```
--erofs=v1    (default) Generate only V1 EROFS. Sets "v1_erofs" in ro_compat.
--erofs=dual            Generate both V1 and V2 EROFS. Omits "v1_erofs" from ro_compat.
```

Idempotency: if the repo already exists and the `erofs_formats` setting differs, an error is
returned (same behaviour as a mismatched `algorithm`).

## bootc integration

bootc initializes its composefs repository with:

```rust
RepositoryConfig {
    algorithm: Algorithm::SHA512,
    erofs_formats: FormatSet::BOTH,  // no "v1_erofs" in ro_compat
    ..Default::default()
}
```

This causes every image pull to generate both V1 and V2 EROFS objects automatically.

`bootc container ukify` gains a `--format` option:

```
--format=1    Embed composefs.digest=<v1-digest> in the UKI cmdline.
--format=2    (default for now) Embed composefs=<v2-digest> in the UKI cmdline.
```

The plan is to flip the default to `--format=1` once the new initramfs logic ships broadly.

## Migration

**Existing bootc repos** (initialized before this change, with V2 default and no `v1_erofs`
flag): because bootc will now pass `erofs_formats: FormatSet::BOTH` at init time, and `init_path`
is idempotent, the repo's `meta.json` is updated to reflect BOTH on next open. `default_format_set()`
then returns `BOTH`. On the next image pull, both formats are generated. Existing sealed V2 UKIs
continue to boot via the existing `composefs=<v2-digest>` karg and V2 image. No manual migration
step needed.

**Existing plain `cfsctl` repos** (V1 default, `v1_erofs` present): remain V1-only unless the
user re-runs `cfsctl init --erofs=dual`. Upgrading from V1-only to dual is allowed by `init_path`;
downgrading from dual back to V1-only returns an error (would silently stop generating V2 images
that existing sealed UKIs may depend on).

## Files affected

### composefs-rs

| File | Change |
|------|--------|
| `crates/composefs/src/erofs/format.rs` | Add `FormatSet` type |
| `crates/composefs/src/repository.rs` | `known_features::V1_EROFS`, `RepositoryConfig::erofs_formats`, `Repository::default_format_set()`, `commit_images()` |
| `crates/composefs-oci/src/lib.rs` | `IMAGE_REF_KEY_V1`, `BOOT_IMAGE_REF_KEY_V1`; update `create_filesystem` and `generate_boot_image` |
| `crates/composefs-oci/src/boot.rs` | `generate_boot_image` gains `FormatSet` param |
| `crates/composefs-boot/src/cmdline.rs` | `ComposefsCmdline` enum; parse/generate both karg names |
| `crates/composefs-ctl/src/lib.rs` | `cfsctl init --erofs=v1|dual` |
| `doc/repository.md` | Document `v1_erofs` feature flag and `FormatSet` semantics |

### bootc (separate repo)

| File | Change |
|------|--------|
| `crates/lib/src/store/mod.rs` | Init repo with `FormatSet::BOTH` |
| `crates/lib/src/ukify.rs` | `--format=1|2`; embed correct karg |
| `crates/lib/src/bootc_composefs/status.rs` | Parse `composefs.digest=` alongside `composefs=` |
| `crates/initramfs/src/lib.rs` | Handle both karg names (duplicated from composefs-boot) |
</details>

cgwalters:
OK, passing CI now, though re-reviewing I see a few more things to fix.

crates/composefs/src/fs.rs
```diff
@@ -542,7 +734,21 @@ pub async fn read_filesystem<ObjectID: FsVerityHashValue>(
...
+    let store: Option<Arc<dyn ObjectStore<ObjectID>>> =
+        repo.map(|r| r as Arc<dyn ObjectStore<ObjectID>>);
+    read_filesystem_impl(dirfd, path, store, None).await
+}
+
+/// Like [`read_filesystem`] but accepts any [`ObjectStore`] implementation.
+///
+/// This is the preferred entry point when using a custom store (e.g.
+/// [`FlatDigestStore`] for C-compatible `--digest-store` behaviour).
+pub async fn read_filesystem_with_store<ObjectID: FsVerityHashValue>(
```
alexlarsson: It feels a bit iffy to add read_filesystem_with_store() when we just added the new read_filesystem_with_semaphore(). Especially when this one doesn't allow the passing in of a custom semaphore. Can we not just add one new function for both of these?

examples/common/fix-verity/fix-verity
```diff
@@ -18,17 +18,29 @@ if [ ! -f ${fix_verity_efi} ]; then
     mv "${fix_verity_efi}.tmp" "${fix_verity_efi}"
 fi
 
-ovmf=""
-for p in /usr/share/OVMF/OVMF_CODE.fd /usr/share/ovmf/OVMF.fd /usr/share/edk2/ovmf/OVMF_CODE.fd /usr/share/edk2/x64/OVMF.4m.fd; do
-    if [ -f "$p" ]; then ovmf="$p"; break; fi
+ovmf_code=""
+ovmf_vars=""
+for d in /usr/share/edk2/ovmf /usr/share/OVMF /usr/share/ovmf /usr/share/edk2/x64; do
```
alexlarsson, line 23: I would probably squash this commit into the previous one.

doc/booting.md
```diff
@@ -0,0 +1,77 @@
+# Booting from a composefs image
+
+This document describes how composefs-rs sets up the root filesystem during
```
alexlarsson, line 3: Why are these new files in `docs/` when previous files are in `doc/`?

crates/composefs-ctl/src/composefs_info.rs
```diff
@@ -0,0 +1,347 @@
...
+//!
+//! Implemented subcommands:
+//! - `ls` — lists files with type suffixes, skips whiteout entries
+//! - `dump` — outputs composefs-dump(5) text format (image → tree → dumpfile)
+//! - `objects` — lists all backing file object paths (XX/XXXX...)
+//! - `missing-objects` — lists objects not present in `--basedir`
+//! - `measure-file` — computes fs-verity digest of files
+//!
+//! Known gaps vs C composefs-info:
+//! - TODO(compat): `measure-file` uses userspace fs-verity computation instead
```
alexlarsson: Is there any particular reason for this? Honestly, i think measure-file was mostly added because lots of places (like cs9) didn't package fsverity-utils.  I don't think this is such an issue anymore, but often the usecase for this would be to check that some file actually had fs-verity enabled, so doing a software hash kinda breaks that.

crates/composefs/src/fs.rs
cgwalters: Good point, let's have `read_filesystem_with_opts` to take an options struct or so.

crates/composefs/src/generic_tree.rs
```diff
@@ -595,6 +611,71 @@ pub struct FileSystem<T> {
...
+                .entries
+                .insert(name.into_boxed_os_str(), Inode::leaf(leaf_id));
+        }
+    }
+
+    /// Add trusted.overlay.opaque="y" xattr to root directory.
+    ///
+    /// This is required for Format 1.0 when whiteout entries are present,
+    /// marking the directory as opaque for the overlay filesystem.
+    pub fn set_overlay_opaque(&mut self) {
```
alexlarsson: I don't think this function is ever called (other than a test).

crates/composefs-ctl/src/composefs_info.rs
```diff
@@ -0,0 +1,347 @@
...
+
+/// Walk and print entries: directory line first, then recurse into children.
+fn ls_print<W: Write>(
+    out: &mut W,
+    fs: &FileSystem<Sha256HashValue>,
+    dir: &composefs::tree::Directory<Sha256HashValue>,
+    path: &[u8],
+    seen_leaf_ids: &mut HashSet<LeafId>,
+    filter: &[String],
+    is_root: bool,
```
alexlarsson: In the C code we just pass NULL for filter when we recurse instead of passing an extra is_root. We could maybe do the same here.

crates/composefs-ctl/src/mkcomposefs.rs
```diff
@@ -0,0 +1,445 @@
...
+        return Ok(());
+    }
+
+    // Write image
+    let image_path = args.image.as_ref().unwrap();
+    write_image(image_path, &image)?;
+
+    // Optionally print digest
+    if args.print_digest {
+        let digest = compute_fsverity_digest(&image);
```
alexlarsson, line 229: I feel we could just print in one place (and remove the above print_digest_only copy of this, and then just make the write_image call above optional.

alexlarsson:
Ok, some random comments above, lets talk a bit about the high level issues.

First of all, how locked down are we with the v2 format? Because I see some issues with it. 

First of all, it doesn't seem to support the escaped whiteouts that v1 does, and I think we really do want to support that. For example, suppose your original image contained an additional image store that contained a layer with a whiteout. Then we can't just stuff a chrdev(0,0) in the erofs image, because that would be parsed not by containers/storage when you used the mounted image, but by the overlayfs mount that composefs uses. This is why v1 generates an escaped overlayfs xattr which will get unescaped by the first overlayfs mount, and then applied by the second overlayfs mount. We need this for V2 too.

Secondly, after composefs 1.0 was released (and the format fixed), erofs added support for adding custom xattr prefixes in the image header. Since composefs files all are full of "trusted.overlay.XXX" prefixes this could be used to limit the size of the images a bit. Its sad that we changed from the v1 format and did not take advantage of this. (NOTE: There was also the xattr bloom filter i mentioned to you before, but I notice that we actually managed to get that into v1, so we're good there.). I'm not sure if this is supported in CS9 though?

Also, the handling of versions in the docs/comments/etc is a bit unclear, sometimes it say v2_only is default, sometimes v1 only. Sometimes it says "both" means both, but sometimes it seems to mean v2 only. Honestly, looking at v1 vs v2 atm I'm not sure what advantages v2 gives? It makes some sense i guess to default to v1, although I would really want a newer version with the xattr prefixes.

There is no version of the mount helper (mount.composefs) in the v1 support. This can be added later though.

Should we maybe add some deprecation warnings in the composefs and mkcomposefs output?

cgwalters:
> First of all, how locked down are we with the v2 format? Because I see some issues with it.

If we changed it it would break everyone using bootc sealed UKIs. It's...a possibility to make that change, but...I would lean towards defining a v3 format instead that combines the advantages of v1 and v2. I think that's a *lot* more tractable now with the infrastructure that lands here.

> First of all, it doesn't seem to support the escaped whiteouts that v1 does, and I think we really do want to support that. For example, suppose your original image contained an additional image store that contained a layer with a whiteout.

Yeah though of course this is all circular because if podman used composefs natively, we wouldn't have those whiteouts on disk...

> Secondly, after composefs 1.0 was released (and the format fixed), erofs added support for adding custom xattr prefixes in the image header. Since composefs files all are full of "trusted.overlay.XXX" prefixes this could be used to limit the size of the images a bit. Its sad that we changed from the v1 format and did not take advantage of this. 

This does seem doable to change in v3.

However, when I dug into it one thing it'd be really nice to improve in general is the fact that we need both `trusted.overlay.metacopy` *and* `trusted.overlay.redirect` in two different formats. Ideally...while it'd be super special cased we could add something like a mount option for overlayfs `metacopy-redirect=xx/` or something to mean to convert the binary metacopy to hex. But it'd be a format string so if someone wanted they could do `xxx/` or `xx/xx/` or whatever.

## Reducing EROFS sizes in a future v3

Note with SHA-512 the overhead here is 129 bytes per external file for `trusted.overlay.redirect`...that's actually pretty huge.

Based one some quick research, it's pretty common for the `trusted.overlay.redirect` hex digests to be *half of the image size*.

That seems well worth fixing...

---

> (NOTE: There was also the xattr bloom filter i mentioned to you before, but I notice that we actually managed to get that into v1, so we're good there.). I'm not sure if this is supported in CS9 though?

It's just extra data right that would be safely ignored but unused?

> Also, the handling of versions in the docs/comments/etc is a bit unclear, sometimes it say v2_only is default, sometimes v1 only. Sometimes it says "both" means both, but sometimes it seems to mean v2 only. 

OK yeah this is an artifact of me going back and forth on this. I'm cleaning this up now to very clearly continue to default to v1.

> There is no version of the mount helper (mount.composefs) in the v1 support. This can be added later though.

That one should be easy, but yes we should track it. The shared library is the hard one.

alexlarsson:
If you want to compress it further, then encode the digest not as hex but base64 or something

alexlarsson:
You'd have to hardlink the hex version for backwards compatibility.

cgwalters:
> If you want to compress it further, then encode the digest not as hex but base64 or something

We already have the digest in a *binary* format in the EROFS, the idea with the overlayfs patch would be to avoid encoding it at all a second time in the EROFS - instead it'd be generated in-kernel on demand.

As far as the filename on disk...yeah, it could be smaller. That would be an even more incompatible change, not clear it's worth it. The thing is: we pay the metadata cost per image, whereas we only store an object once on disk.

alexlarsson:
So, unfortunately the custom prefix support is not in the erofs in cs9, so a v3 based on it would not be generally useful. (In contrast to the bloom filter, this is not an optional feature that can be ignored.).

As for auto-generating the redirect from the metacopy xattr,. Yes, that could be done either by a mount option, or via extending the (versioned) overlay.metacopy structure to add a flag for doing this. I guess a third approach is to add some magic in erofs itself to synthesize the redirect xattr from the metacopy one. However, any of these would require upstream kernel work, backporting to released kernels (or waiting) and then again, we end up with a new format that is not usable for old kernels. 

I think the right approach right now is to keep v1 the default, support v2 for old deployed bootc images only, and then collect reasons somewhere for eventually in the future adding a new v3.

crates/composefs/src/erofs/writer.rs
```diff
@@ -571,176 +1501,520 @@ fn share_xattrs(inodes: &mut [Inode<impl FsVerityHashValue>]) -> Vec<XAttr> {
...
+///
+/// Returns the complete EROFS image as a byte array.
+pub fn mkfs_erofs_versioned<ObjectID: FsVerityHashValue>(
+    fs: &mut ValidatedFileSystem<ObjectID>,
+    version: format::FormatVersion,
+) -> Box<[u8]> {
+    mkfs_erofs_inner(
+        &mut fs.0,
+        version,
+        0,
```
alexlarsson: Can we please add min-ver == 1 here so we don't create the unnecesessary 00-ff whiteouts. Those are not needed since support was added for data-only layers, which are everywhere by now. I understand we need to support this for mkcomposefs, but for the cfscfl we should default to v1 with min ver 1.

crates/composefs-ctl/src/lib.rs
```diff
@@ -221,6 +232,26 @@ pub enum HashType {
     Sha512,
 }
 
+/// The EROFS format version used when generating images.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ValueEnum)]
+pub enum ErofsVersion {
+    /// Format V1: compact inodes, BFS, C-compatible.
+    #[clap(name = "1")]
+    V1,
```
alexlarsson, line 227: Honestly, i think calling it V1 and then having the separate min-version thing be 0 or 1 is kinda weird. I would much prefer if they were separate versions V0 and V1, and we drop the min version thing. Then we can have cfsctl default to V1, and mkcomposefs (unless you pass --min-version) default to V0. I think that makes things much clearer.

crates/composefs/src/erofs/format.rs
```diff
@@ -271,11 +271,103 @@ impl std::ops::BitOr<u32> for FileType {
...
+/// A `FormatSet` is a small bitset (bit 0 = V1, bit 1 = V2) so it can be
+/// cheaply copied and tested without heap allocation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FormatSet(u8);
+
+impl FormatSet {
+    /// Generate only V1 EROFS (default for new repos; C-tool compatible).
+    pub const V1_ONLY: FormatSet = FormatSet(0b01);
+    /// Generate both V1 and V2 EROFS (used by bootc and other multi-format consumers).
+    pub const BOTH: FormatSet = FormatSet(0b11);
```
alexlarsson: Maybe BOTH is not a great name if we ever consider adding a V3.

alexlarsson:
Also, it seems like the default is still V2, both for "cfsctl init", and the #default for the FormatVersion enum.

alexlarsson:
Other than the above though, I think this is a good start and we should try to get it merged.

crates/composefs/src/erofs/format.rs
alexlarsson: (And if we add V0)

crates/composefs-ctl/src/lib.rs
cgwalters: Hmm, then in theory we'd need to test *all three* versions across things like our fuzzer etc. Which I guess is more formally correct anyways.

cgwalters:
> Also, it seems like the default is still V2, both for "cfsctl init", and the #default for the FormatVersion enum.

Yeah, I want to make flipping the default a clearly separate PR.

crates/composefs/src/erofs/format.rs
cgwalters: Yeah, I guess `const V1_V2`?

crates/composefs/src/erofs/writer.rs
cgwalters: AFAICS composefs-c always creates the 00-ff whiteouts.

alexlarsson: Gah, you're right. It seems my memory has holes.

crates/composefs-ctl/src/lib.rs
```diff
@@ -221,6 +232,30 @@ pub enum HashType {
     Sha512,
 }
 
+/// The EROFS format version used when generating images.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ValueEnum)]
+pub enum ErofsVersion {
+    /// Format V0: compact inodes, BFS, C-compatible (composefs_version auto-detects 0 or 1).
+    #[clap(name = "0")]
+    V0,
+    /// Format V1: same layout as V0, composefs_version always 1. Default for new repos.
```
alexlarsson: "Default for new repos" is not yet true

alexlarsson:
This now looks good to me. If I had one complaint it would be that the last commit is a bit too large. It would make sense to at least split out the mtim_nsec change, and the RepoConfig change to separate commits, and then split the main commit into v1 API support and CLI support.

But I'm fine with this not split up too.

alexlarsson (dismissed):

crates/composefs-ctl/src/lib.rs
cgwalters: Thanks, fixed!

cgwalters:
> If I had one complaint it would be that the last commit is a bit too large. It would make sense to at least split out the mtim_nsec change, and the RepoConfig change to separate commits, and then split the main commit into v1 API support and CLI support.

Yeah, I tried that, it got messy. This PR is itself the result of various splits too (a whole large sidequest here was e.g. https://github.com/composefs/composefs-rs/pull/256 )

cgwalters:
BTW this repo is setup so that pushing to a PR drops review approvals, so you'll need to readd

alexlarsson (approved):

alexlarsson:
This is now in https://github.com/composefs/composefs-run

PR #226, opened 2026-02-08
WIP: Rough draft for updated generic OCI sealing

cgwalters:
This is just some rough draft raw material that builds on:

- https://github.com/containers/composefs-rs/pull/216 (🆕 now requires manifest/config as external not inline splitstream objeects, also the OCI artifact and referrer support was key)
- https://github.com/containers/composefs-rs/pull/224 (Lots of updates there, I redid to go all-in on an external signature+digests by default)

### DIFF
--- a/crates/composefs-oci/src/oci_image.rs
+++ b/crates/composefs-oci/src/oci_image.rs
@@ -38,12 +38,15 @@
 //! This module handles both transparently. Use `is_container_image()` to check.
 
 use std::collections::{HashMap, HashSet};
+use std::str::FromStr;
 use std::sync::Arc;
 
 use anyhow::{Context, Result, ensure};
 use containers_image_proxy::oci_spec::image::{
-    Descriptor, Digest as OciDigest, ImageConfiguration, ImageManifest, MediaType,
+    Descriptor, DescriptorBuilder, Digest as OciDigest, ImageConfiguration, ImageManifest,
+    ImageManifestBuilder, MediaType,
 };
+use ocidir::OciRead;
 use rustix::fs::{AtFlags, Dir, Mode, OFlags, openat, readlinkat, unlinkat};
 use rustix::io::Errno;
 use serde::Serialize;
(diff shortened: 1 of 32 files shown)


crates/composefs-oci/src/oci_image.rs
```diff
@@ -857,6 +848,331 @@ pub fn cleanup_dangling_referrers<ObjectID: FsVerityHashValue>(
...
+/// # Arguments
+///
+/// * `repo` - The composefs repository
+/// * `subject_digest` - The manifest digest of the subject image (sha256:...)
+/// * `oci_layout_path` - Path to the OCI layout directory
+///
+/// # Returns
+///
+/// The number of referrer artifacts exported.
+pub fn export_referrers_to_oci_layout<ObjectID: FsVerityHashValue>(
```
cgwalters: Something like this could land as a prep commit

crates/composefs/src/fsverity/ioctl.rs
```diff
@@ -170,4 +190,23 @@ mod tests {
         assert!(matches!(err, EnableVerityError::Io(..)));
         assert_eq!(err.to_string(), "Bad file descriptor (os error 9)",);
     }
+
+    #[test]
+    fn test_fs_ioc_enable_verity_with_sig_bad_fd() {
+        let fd = ManuallyDrop::new(unsafe { OwnedFd::from_raw_fd(123456) });
```
cgwalters: Gross, no. This unit test has no value, should just be integration testing this.

alexlarsson:
I've been doing some experimental work on something similar with the current composefs support in podman at https://github.com/podman-container-tools/podman/pull/28658. That code encodes the expected composefs digests in the manifests and assumes it will be reproducibly generated from the tar layers locally. To get trust in that I then have to validate the manifest signature at runtime. I think long term I prefer the approach proposed here of distributing the erofs images directly, particularly because it allows kernel side signatures that can integrate with IMA, etc.

However... I wonder if not just signing the erofs image is missing something. Don't we actually want to sign the entire container specification, not just the file content? What about other parts of the image, like exposed ports, volumes, cmdline, etc? With the signature on the erofs blob we leave all the stuff in the manifest and config "unprotected".

cgwalters:
>  Don't we actually want to sign the entire container specification, not just the file content? What about other parts of the image, like exposed ports, volumes, cmdline, etc? With the signature on the erofs blob we leave all the stuff in the manifest and config "unprotected".

In the current design, the manifest and config are always stored as external objects, and we can then apply fsverity signatures to those as well.

It's a bit buried in the spec, but see https://github.com/composefs/composefs-rs/pull/224/changes#diff-def2a4eef2075f93a81da71d729f633c5f748feff036fd3789252ee37cf37dfdR281

cgwalters:
A simple way to say it is, I think with this proposal we don't actually *need* to sign the manifest via cosign at all - runtime integrity &gt;= transport integrity.

But of course in practice most use cases would do so, because you still want the transport integrity for deployments that aren't using runtime integrity yet.

PR #228, opened 2026-02-10
[WIP] composefs-splitfdstream: new crate

giuseppe:
Add a client library for communicating with the containers-storage splitfdstream server.  This enables importing OCI container layers via UNIX socket with file descriptor passing.

Needs: https://github.com/containers/container-libs/pull/651/

### DIFF
--- a/crates/composefs-oci/src/splitfdstream.rs
+++ b/crates/composefs-oci/src/splitfdstream.rs
@@ -1,0 +1,297 @@
+//! Client for importing container images and layers from a splitfdstream server.
+//!
+//! Uses the `jsonrpc-fdpass` crate for the JSON-RPC 2.0 + SCM_RIGHTS protocol.
+
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{Cursor, Read};
+use std::os::fd::OwnedFd;
+use std::path::Path;
+use std::sync::Arc;
+
+use anyhow::{Context, Result, bail};
+use containers_image_proxy::oci_spec::image::{ImageConfiguration, ImageManifest};
+use jsonrpc_fdpass::{JsonRpcMessage, JsonRpcRequest, MessageWithFds, UnixSocketTransport};
+use tokio::net::UnixStream;
+
+use composefs::fsverity::FsVerityHashValue;
+use composefs::repository::Repository;
+use splitfdstream::{Chunk, SplitfdstreamReader};
+
+use crate::skopeo::TAR_LAYER_CONTENT_TYPE;
+
+/// Send a JSON-RPC request and receive the response with file descriptors.
+async fn rpc_call(
+    socket_path: &Path,
+    method: &str,
+    params: serde_json::Value,
+) -> Result<(serde_json::Value, Vec<OwnedFd>)> {
+    let stream = UnixStream::connect(socket_path)
+        .await
+        .context("connecting to splitfdstream server")?;
+
+    let transport = UnixSocketTransport::new(stream);
+    let (mut sender, mut receiver) = transport.split();
+
+    let request = JsonRpcRequest::new(method.to_string(), Some(params), serde_json::json!(1));
+    let message = MessageWithFds::new(JsonRpcMessage::Request(request), Vec::new());
+    sender.send(message).await.context("sending request")?;
+
+    let response = receiver.receive().await.context("receiving response")?;
+
+    let resp_value = response
+        .message
+        .to_json_value()
+        .context("converting response to value")?;
+
+    if let Some(err) = resp_value.get("error") {
+        let msg = err
+            .get("message")
+            .and_then(|m| m.as_str())
+            .unwrap_or("unknown error");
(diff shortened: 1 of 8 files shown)


crates/composefs-splitfdstream/src/stream.rs
```diff
@@ -0,0 +1,185 @@
+//! Binary stream format parser for splitfdstream.
```
cgwalters: I did a whole implementation of this over in https://github.com/containers/composefs-rs/pull/218/changes#diff-69fe9e0f1388b28cd8d358d8f76589aaf790016b57fca196c2d22e0f93012b9b - any reason not to unify?

cgwalters (changes requested):

crates/composefs-oci/src/splitfdstream.rs
```diff
@@ -0,0 +1,376 @@
...
+            fd_queue: Vec::new(),
+        }
+    }
+
+    /// Read more data (and possibly FDs) from the socket.
+    /// Mirrors Go `Receiver.readMoreData` / `recvWithFDs`.
+    ///
+    /// Uses libc recvmsg directly because rustix 1.x has a bug where
+    /// msg_controllen is set to 0, preventing SCM_RIGHTS reception.
+    #[allow(unsafe_code)]
```
cgwalters: Hmm really? Are you sure? We were successfully using rustix in https://github.com/cgwalters/jsonrpc-fdpass ...

(Also why are we not using that crate? I can publish it formally...ok just transferred it to the bootc-dev org along with the Go version too)

If there's fixes needed in rustix the maintainer is responsive.

We're actually very close to `#[forbid(unsafe_code)]` in this project, it's just https://github.com/containers/composefs-rs/issues/123

cgwalters (changes requested):

crates/composefs-oci/src/splitfdstream.rs
```diff
@@ -0,0 +1,376 @@
...
+        UnixStream::connect(socket_path.as_ref()).context("connecting to splitfdstream server")?;
+
+    // Build request JSON.
+    let mut params = serde_json::json!({ "layerId": effective_layer_id });
+    if let Some(pid) = parent_id {
+        params["parentId"] = serde_json::json!(pid);
+    }
+    let request = serde_json::json!({
+        "jsonrpc": "2.0",
+        "method": "GetSplitFDStream",
```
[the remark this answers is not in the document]
cgwalters: Or in other words... look at what I did in https://github.com/containers/composefs-rs/pull/218 the key thing I think here is fast-pathing the `containers-storage:` transport using this.

I think in order to do that we really do need to have not just an "import a single layer" as a CLI verb but do the whole "fetch manifest+config + layers" right?

crates/composefs-oci/src/splitfdstream.rs
```diff
@@ -0,0 +1,393 @@
...
+            // Try to parse a complete JSON value from buffered data.
+            if !self.buffer.is_empty() {
+                let mut stream = serde_json::Deserializer::from_slice(&self.buffer)
+                    .into_iter::<serde_json::Value>();
+                match stream.next() {
+                    Some(Ok(value)) => {
+                        let consumed = stream.byte_offset();
+                        self.buffer.drain(..consumed);
+
+                        // Dequeue FDs indicated by the "fds" field.
```
cgwalters: This code can just use jsonrpc-fdpass rust crate now right?

crates/composefs-oci/src/splitfdstream.rs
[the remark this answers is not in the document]
cgwalters: @giuseppe did you see this one? I think to really make this work we need to wire it into the production pull code paths that are used by bootc, not just a lookaside CLI verb right?

Of course doing this is going to lead to conflicts with https://github.com/composefs/composefs-rs/pull/218 because it basically wants the same "detect containers-storage, dispatch to different pull path" logic.

My preference honestly is to land that PR first to get us unblocked, then we continue iteration on this one and once we can rely on always having a new enough skopeo/podman we can drop the cstor crate or so?

giuseppe:
> @giuseppe did you see this one? I think to really make this work we need to wire it into the production pull code paths that are used by bootc, not just a lookaside CLI verb right?

probably I can even close this PR, I am using it only to test the integration with the container-libs changes.

> Of course doing this is going to lead to conflicts with https://github.com/composefs/composefs-rs/pull/218 because it basically wants the same "detect containers-storage, dispatch to different pull path" logic.
> My preference honestly is to land that PR first to get us unblocked, then we continue iteration on this one and once we can rely on always having a new enough skopeo/podman we can drop the cstor crate or so?

#218 still scares me because it is not clear for how long we'd use that and if something goes wrong with the new API we will be stuck with that implementation, but if you’re confident in the approach, please go ahead.

cgwalters:
Closing in favor of https://github.com/composefs/composefs-rs/pull/309 - I think we're agreed on this "splitdirfdstream" approach instead.

PR #229, opened 2026-02-10
Add splitfdstream crate for binary streaming format

cgwalters:
Add a new crate implementing the splitfdstream binary format, which enables streaming tar archives with external file references.

The format uses an 8-byte LE signed prefix followed by optional data:
- Negative prefix: inline data of size |prefix|
- Non-negative prefix: external file descriptor reference by index

Key types:
- SplitfdstreamWriter: Build splitfdstreams with inline/external chunks
- SplitfdstreamReader: Parse splitfdstreams back into chunks
- SplitfdstreamTarReader: Read adapter that reconstructs byte streams

This enables zero-copy tar reconstruction from containers-storage's tar-split metadata by streaming headers inline while referencing file content via fd.

### DIFF
--- a/crates/splitfdstream/Cargo.toml
+++ b/crates/splitfdstream/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "splitfdstream"
+description = "Binary format for serializing data with external file descriptor references"
+keywords = ["splitfdstream", "fd", "serialization"]
+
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
I am fine with the code, but do we want to merge even without consensus on the c/storage side?

LGTM

PR #233, opened 2026-02-13
oci: Add fast path for oci: transport using ocidir crate

cgwalters:
When importing from local OCI layout directories (oci: transport), read
the layout directly instead of going through skopeo; this path
is faster.

But another motivation here is that right now the skopeo proxy
always peels manifest lists, but I want to add proper support
for storing those. Fetching from local `oci` helps us write
unit/integration tests for that without blocking on patching
skopeo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ composefs-ioctls = { version = "0.3.0", path = "crates/composefs-ioctls", defaul
 composefs-oci = { version = "0.3.0", path = "crates/composefs-oci", default-features = false }
 composefs-boot = { version = "0.3.0", path = "crates/composefs-boot", default-features = false }
 composefs-http = { version = "0.3.0", path = "crates/composefs-http", default-features = false }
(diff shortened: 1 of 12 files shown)


cgwalters:
> Looks like this needs rebased after https://github.com/composefs/composefs-rs/pull/248 landed

Done

Johan-Liebert1:
This needs another rebase

cgwalters:
This also wants https://github.com/bootc-dev/ocidir-rs/pull/73

Johan-Liebert1 (changes requested):
I get this error when trying `cfsctl oci pull <image>` locally

```
Caused by:
    0: Opening stream ''
    1: Creating new splitstream reader
    2: Invalid splitstream content type
```

Would be great to have contexts on functions as it's pretty tedious to trace this error

Johan-Liebert1:
The above error seems to be originating from `create_filesystem` function.

Johan-Liebert1 (approved):
LGTM

PR #283, opened 2026-04-07
tests: Use realistic file sizes and xattrs in test image layers

cgwalters:
Use external files with realistic sizes (4 KiB busybox, 8 KiB libc, etc.) instead of tiny inline content strings.  Files above the 64-byte INLINE_CONTENT_MAX_V0 threshold now exercise the external object storage path, which is important for testing containers-storage import later.

Add PAX extended header support to dumpfile_to_tar so that xattrs (e.g. security.capability) round-trip through the tar pipeline.  The ping symlink now carries a CAP_NET_RAW capability xattr, matching what real container images look like.

External files get deterministic pseudo-random content via (0..size).map(|i| (i % 251) as u8), ensuring distinct fsverity digests per file while remaining reproducible.

Assisted-by: OpenCode (Claude Opus 4)

### DIFF
--- a/crates/composefs-oci/Cargo.toml
+++ b/crates/composefs-oci/Cargo.toml
@@ -11,7 +11,7 @@ rust-version.workspace = true
 version.workspace = true
 
 [features]
-test = ["tar", "composefs/test"]
+test = ["tar", "rand", "composefs/test"]
 boot = ["composefs-boot"]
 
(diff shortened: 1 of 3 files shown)


cgwalters:
I just rolled this into https://github.com/composefs/composefs-rs/pull/218

PR #285, opened 2026-04-24
repo: Support reading and upgrading from older composefs-rs repos

cgwalters:
Production bootc (pinned at composefs-rs 2203e8f) predates three format changes:

 - ce6628525 repository: Add meta.json for repo metadata and cfsctl init
 - b7dc27065 Add repr(c) for SplitStream header structs
 - d5ec81d74 oci: Generate composefs EROFS at pull time, track via config refs

This commit adds transparent backward compatibility and a non-destructive upgrade path.

The splitstream reader detects old-layout headers (where the Rust compiler reordered SplitstreamHeader fields) and converts them on the fly. Repository::open_upgrade() infers the algorithm and verity mode from existing objects when meta.json is missing, writes it, and opens normally — replacing the destructive --reset-metadata flow. upgrade_repo() walks all tagged images and generates EROFS for any that lack it, rewriting config and manifest splitstreams in the current format. Layer splitstreams stay old-format on disk since the reader handles them transparently.

The CLI uses open_upgrade by default (opt out with --no-upgrade). An integration test pulls a real image with an old cfsctl binary (requires CFSCTL_PATH_OLD) to verify the full upgrade path.

Assisted-by: OpenCode (Claude Opus 4)

### DIFF
--- a/crates/cfsctl/src/lib.rs
+++ b/crates/cfsctl/src/lib.rs
@@ -99,6 +99,12 @@ pub struct App {
     #[clap(long)]
     require_verity: bool,
 
+    /// Don't automatically upgrade old-format repositories.
+    /// When set, commands will fail on repos without meta.json instead
+    /// of inferring metadata from existing objects.
+    #[clap(long)]
+    no_upgrade: bool,
+
     /// Don't open a repository. Only valid for commands that don't need one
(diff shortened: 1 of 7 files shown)


giuseppe (approved):
LGTM

PR #296, opened 2026-05-20
selabel: Break hardlinks when different SELinux labels are assigned

cgwalters:
When two filesystem paths are hardlinked (share the same LeafId in the in-memory tree) but the SELinux policy assigns them different labels, we had "last looked at wins" behavior.

This conflict between hardlinks and SELinux has come up many times (it bit me badly in bootc when trying to have the binary also be
 a systemd generator, so we ended up with a helper shell script).

Fix this by just breaking the hardlink, which is also what OSTree does.

Assisted-by: OpenCode (claude-sonnet-4-6@default)


giuseppe (approved):
LGTM

PR #297, opened 2026-05-22
Extend v1 EROFS to OCI and booting

cgwalters:
Builds on https://github.com/composefs/composefs-rs/pull/225

In a nutshell: I want to do a cutover to generating V1 EROFS (i.e. bit-for-bit compatible with the original https://github.com/composefs/composefs ). The rationale here is basically we want to be able to replace that project entirely in a compatible way *and* we need to support RHEL9 and our current EROFS ("v2") isn't compatible because we were implicitly relying on https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9ed50b8231e37b1ae863f5dec8153b98d9f389b4

## Handling EROFS versions

In an OCI image splitstream, we can now generate and reference either or both versions. For new repositories, we default to v1. Apart from UKIs, we don't care about the binary checksum of the EROFS, so it's totally fine to just change what we do there. One can use composefs to fetch OCI images and mount them as before, and the fact the EROFS is different is largely invisible.

### Sealed UKIs

However, we *do* care for sealed UKIs that have a checksum of the EROFS. A `composefs=` karg signals that V2 EROFS is used. A new `composefs.digest=` karg  signals that V1 EROFS is used.

Conveniently, I'm also using this transition to get rid of the annoying kernel warning because presence of a `.` signals it's for userspace, as is the case here.

### DIFF
--- a/crates/composefs-boot/src/cmdline.rs
+++ b/crates/composefs-boot/src/cmdline.rs
@@ -6,14 +6,245 @@
 //! insecure mode indicators.
 
 use anyhow::{Context, Result};
-use composefs::fsverity::FsVerityHashValue;
+use composefs::fsverity::{Algorithm, FsVerityHashValue};
+
+/// Legacy kernel argument for V2 EROFS: `composefs=<hex_digest>`.
+///
+/// Shorthand for `composefs.digest=v2-<hash>-12:<hex>`.  Used in existing
+/// sealed UKIs.  The initramfs checks for [`KARG_COMPOSEFS_DIGEST`] first,
+/// then falls back to this.
+pub const KARG_V2: &str = "composefs";
+
+/// Self-describing kernel argument: `composefs.digest=<version>-<hash>-<lg>:<hex>`.
+///
+/// The value encodes the EROFS format version, hash algorithm, and block size,
+/// e.g. `composefs.digest=v1-sha256-12:<hex>` or `composefs.digest=v2-sha512-12:<hex>`.
+/// Both `v1` and `v2` are accepted; `composefs=<hex>` is a legacy alias for
+/// the `v2` form.
+///
+/// Multiple entries may appear on the cmdline with different format/algorithm
+/// combinations; the initramfs tries each in order, mounting the first image
+/// that exists in the repository.
+pub const KARG_COMPOSEFS_DIGEST: &str = "composefs.digest";
+
+/// A composefs kernel argument identifying which EROFS image to mount at boot.
+///
+/// Two variants exist to distinguish EROFS format versions:
+/// - [`ComposefsCmdline::V2`]: V2 EROFS — either `composefs=<digest>` (legacy shorthand)
+///   or `composefs.digest=v2-<hash>-<lg>:<digest>` (explicit form)
+/// - [`ComposefsCmdline::V1`]: V1 EROFS — `composefs.digest=v1-<hash>-<lg>:<digest>`
+///
+/// The initramfs checks for `composefs.digest=` first (accepting both `v1` and `v2`
+/// descriptors), then falls back to the legacy `composefs=` shorthand.
+/// Multiple `composefs.digest=` entries may appear on the cmdline (different
+/// format/algorithm combinations); the initramfs tries each in order, mounting
+/// the first image that exists.
+///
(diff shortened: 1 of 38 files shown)


alexlarsson:
I think it is important to get this in and I'll start reviewing more carefully it on monday. Which PR is the latest though? Should i just ignore #225?

Also, I'm not 100% sure about defaulting to v1, because the xattr think in v2 is likely to improve perfomance in significant ways.

cgwalters:
Thanks Alex! This one is an extension of #225 - I think that one should go first.

crates/composefs-boot/src/cmdline.rs
```diff
@@ -6,14 +6,216 @@
...
+    /// such tokens may appear on the cmdline (different algorithms); the first one whose
+    /// format descriptor matches the `ObjectID` algorithm is returned.  Then falls back to
+    /// `composefs=` (→ [`ComposefsCmdline::V2`]).  Returns `None` if no matching token is
+    /// present.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if a matching karg is found but the hex digest cannot be parsed
+    /// for the given `ObjectID` type.
+    pub fn from_cmdline(cmdline: &str) -> Result<Option<Self>> {
```
cgwalters, line 145: TODO: unify with bootc karg parsing

jeckersb (dismissed):
Didn't review thoroughly but an AI-assisted review was mostly pointing out issues in the kargs parsing which I had already raised in a live call yesterday and we have a plan to remedy over in #180 so I'm not too worried about that for now.

cgwalters:
Investigating the bootc revdep CI failure here.

cgwalters:
OK! This one is finally passing the bootc revdep, should be good to go

giuseppe (dismissed):
LGTM

cgwalters:
OK one more tweak...I noticed the composers-boot logic wasn't handling multiple `composefs=/composefs.digest` arguments the same way bootc was. Now there's a shared helper.

cgwalters:
@giuseppe this repo is configured so that pushes to a PR invalidate previous approvals (we're conservative)

giuseppe (approved):
LGTM

alexlarsson:
I worry about the kernel command line length with this. The docs say:

> The number of kernel parameters is not limited, but the length of the complete command line (parameters including spaces etc.) is limited to a fixed number of characters. This limit depends on the architecture and is between 256 and 4096 characters. It is defined in the file ./include/uapi/asm-generic/setup.h as COMMAND_LINE_SIZE.

And, in the automotive space we've definitely had issues with such limits. For example, android boot is limited to 511 bytes of commandline. We do currently when producing such images use "bootconfig" to move options into the kernel binary itself, which would fix this. However, it may make sense, to ensure (while we have the chance) that the commandline is not unnecessary long, by e.g. using (or at least optionally supporting) a base64 encoded digest intead of a hex one.

cgwalters:
> I worry about the kernel command line length with this. 

Yeah, good point. I think what we could do is add support for looking up the digest from e.g. `/usr/lib/composefs/rootfs.[json|toml|bikeshed]` or so.

It'd be *relatively* straightforward to change `bootc container ukify` to instead inject that file into the initramfs.

It'd be a good idea to land support for this soon, then we can more reasonably cut over late.

> We do currently when producing such images use "bootconfig" to move options into the kernel binary itself, which would fix this.

Hmm...that's a thing we could also do when generating a UKI too right? Do options injected that way show up in `/proc/cmdline`?

alexlarsson:
@cgwalters Yes, these options work like normal in e.g. /proc/cmdline. The only weakness is that some commandline options are read very early by the kernel before the boot config data has been read. What we do is extract a subset of options (https://gitlab.com/CentOS/automotive/src/aboot-deploy/-/blob/main/aboot-update.c?ref_type=heads#L126) and extract only these as the main commandline, and then we dump the rest in a file, and use `bootconfig -a $the_file $the_initrd` to append the remaining options to the initrd.

In the end, it uses this: https://docs.kernel.org/admin-guide/bootconfig.html

alexlarsson:
Hmm, actually I'm not sure they will be in /proc/cmdline, there seems to be a /proc/bootconfig thing instead...

PR #303, opened 2026-05-29
Add support for mounting upper layers with images

alexlarsson:
This is useful from the cli if you want to add a writable layer to a composefs mount. This is better than doing it in a separate overlayfs layer as that would increase the overlayfs stacking depth, etc.

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -63,6 +63,7 @@ use composefs::{
     erofs::reader::erofs_to_filesystem,
     fsverity::{Algorithm, FsVerityHashValue, Sha256HashValue, Sha512HashValue},
     generic_tree::{FileSystem, Inode},
+    mount::MountOptions,
     repository::{
         REPO_METADATA_FILENAME, Repository, RepositoryConfig, read_repo_algorithm, system_path,
         user_path,
@@ -437,6 +438,15 @@ enum OciCommand {
         /// Mount the bootable variant instead of the regular EROFS image
         #[arg(long)]
         bootable: bool,
+        /// Writable upper layer directory for overlayfs
+        #[arg(long, requires = "workdir")]
+        upperdir: Option<PathBuf>,
+        /// Work directory for overlayfs (required with --upperdir)
+        #[arg(long, requires = "upperdir")]
+        workdir: Option<PathBuf>,
+        /// Mount read-write (requires --upperdir)
+        #[arg(long, requires = "upperdir")]
+        read_write: bool,
     },
     /// Compute the composefs image ID of a stored OCI image's rootfs
     ///
@@ -564,6 +574,15 @@ enum Command {
         name: String,
         /// the mountpoint
         mountpoint: String,
+        /// Writable upper layer directory for overlayfs
+        #[arg(long, requires = "workdir")]
+        upperdir: Option<PathBuf>,
+        /// Work directory for overlayfs (required with --upperdir)
+        #[arg(long, requires = "upperdir")]
+        workdir: Option<PathBuf>,
+        /// Mount read-write (requires --upperdir)
+        #[arg(long, requires = "upperdir")]
+        read_write: bool,
     },
     /// Read rootfs located at a path, add all files to the repo, then create the composefs image of the rootfs,
     /// commit it to the repo, and print its image object ID
@@ -664,6 +683,31 @@ where
     );
 
     run_app(args).await
+}
+
+fn get_mount_options(
+    upperdir: Option<&Path>,
+    workdir: Option<&Path>,
+    read_write: bool,
+) -> Result<MountOptions> {
+    let mut options = MountOptions::default();
+    if let (Some(u), Some(w)) = (upperdir, workdir) {
+        let upper_fd = rustix::fs::open(
+            u,
+            OFlags::PATH | OFlags::DIRECTORY | OFlags::CLOEXEC,
+            Mode::empty(),
+        )
+        .with_context(|| format!("Opening upperdir '{}'", u.display()))?;
+        let work_fd = rustix::fs::open(
+            w,
+            OFlags::PATH | OFlags::DIRECTORY | OFlags::CLOEXEC,
+            Mode::empty(),
+        )
+        .with_context(|| format!("Opening workdir '{}'", w.display()))?;
+        options.set_overlay(upper_fd, work_fd);
+    }
+    options.set_read_write(read_write);
+    Ok(options)
 }
 
 #[cfg(feature = "oci")]
@@ -1217,7 +1261,12 @@ where
                 ref image,
                 ref mountpoint,
                 bootable,
+                ref upperdir,
+                ref workdir,
+                read_write,
             } => {
+                let mount_options =
+                    get_mount_options(upperdir.as_deref(), workdir.as_deref(), read_write)?;
                 let img = if image.starts_with("sha256:") {
                     let digest: composefs_oci::OciDigest =
                         image.parse().context("Parsing manifest digest")?;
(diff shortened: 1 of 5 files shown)


alexlarsson:
I was chatting a bit with some people at the embedded recipes conf about using composefs without bootc (for a bit custom stuff) and they would need this to be able to couple composefs images with one or more writable layers.

crates/composefs-ctl/src/lib.rs
```diff
@@ -423,6 +424,15 @@ enum OciCommand {
         /// Mount the bootable variant instead of the regular EROFS image
         #[arg(long)]
         bootable: bool,
+        /// Writable upper layer directory for overlayfs
+        #[arg(long)]
+        upperdir: Option<PathBuf>,
+        /// Work directory for overlayfs (required with --upperdir)
+        #[arg(long)]
+        workdir: Option<PathBuf>,
+        /// Mount read-write (requires --upperdir)
```
Johan-Liebert1, line 447: We could make this cleaner by adding requires `#[arg(long, requires = "upperdir")`. Same with `workdir`

Johan-Liebert1: Or we could have an `overlay-auto` param which will create tmp directories? What do you think?

crates/composefs/src/mount.rs
```diff
@@ -142,6 +146,9 @@ pub fn composefs_fsmount(
     name: &str,
     basedir: impl AsFd,
     enable_verity: bool,
+    upperdir: Option<impl AsFd>,
```
Johan-Liebert1: since both `upperdir` and `workdir` have to exist simultaneously, we could just accept `Option<OverlayOptions>` struct?

Johan-Liebert1 (commented):
Some minor suggestions :)

crates/composefs-ctl/src/lib.rs
alexlarsson: The usecase I discussed was for persistent uppers. We could optionally support transient ones automatically though.

alexlarsson: Automatically handling lifetimes and cleanup for temporary directories in general use might be tricky though.

alexlarsson: I changed this to use the args attribute, and indeed it is much nicer in the code. However, the CLI experience is much worse imho. it basically says "you did something wrong" and then you have to figure out what yourself. 

That is a more generic issue though (the entire CLI sucks) that we can solve later.

crates/composefs/src/mount.rs
alexlarsson: I added this

alexlarsson:
Also, I didn't look at it in detail, but I think composefs-setup-root could use this instead of manually doing the overlayfs mount.

crates/composefs-ctl/src/lib.rs
Johan-Liebert1: Yeah the CLI errors are kinda not good, usually (when done right) the error message is much nicer like so

```
error: the following required arguments were not provided:
    <arg>
```

Johan-Liebert1:
I don't remember well, but I think the "non initramfs" part of `composefs-setup-root` doesn't work as intended. I could be wrong though

alexlarsson:
We have a reverse dependency issue in bootc, where it needs to pass composefs::mount::NO_OVERLAY to mount. What is our current API stability guarantees here? Do i add a new mount_with_overlay() for this, or do we just fix bootc later?

alexlarsson:
Seems easiest to go with mount_with_overlay(), so i added that.

alexlarsson:
I have no idea what this failure is about:
```
    ---- privileged_test_digest_equivalence_debian_bootc ----
    command exited with non-zero code `bcvk ephemeral run-ssh --memory 10G localhost/composefs-rs-test-c9s:latest -- cfsctl-integration-tests --exact privileged_test_digest_equivalence_debian_bootc`: 255


    failures:
        privileged_test_digest_equivalence_debian_bootc

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 60 filtered out; finished in 265.11s

```

cgwalters:
> I was chatting a bit with some people at the embedded recipes conf about using composefs without bootc (for a bit custom stuff)

Would it help for us to support building bootc without the ostree dep? I would like to be able to do that.

cgwalters:
> We have a reverse dependency issue in bootc, where it needs to pass composefs::mount::NO_OVERLAY to mount. What is our current API stability guarantees here? Do i add a new mount_with_overlay() for this, or do we just fix bootc later?

I think We *can* break API at any point, but it's also useful to have a bit of backpressure to avoid doing so unnecessarily. On the counterpoint, https://github.com/composefs/composefs-rs/pull/225 *does* break API and in that PR I pointed the bootc revdep CI at a patch to adapt bootc to it.

alexlarsson:
> > I was chatting a bit with some people at the embedded recipes conf about using composefs without bootc (for a bit custom stuff)
> 
> Would it help for us to support building bootc without the ostree dep? I would like to be able to do that.

Nah, this was for "Flipper one" and the requirement they had was quite custom, focusing on allowing users to tweak and break their systems. I don't think it will ever be a good fit for bootc.

cgwalters:
> I have no idea what this failure is about:

It's likely https://github.com/bootc-dev/bcvk/issues/153 - a rerun made it pass.

crates/composefs/src/mount.rs
```diff
@@ -120,6 +122,20 @@ pub fn erofs_mount(image: OwnedFd) -> Result<OwnedFd> {
...
+    /// Upper layer directory for overlayfs.
+    pub upperdir: U,
+    /// Work directory for overlayfs.
+    pub workdir: W,
+    /// Whether the resulting mount should be read-write.
+    pub read_write: bool,
+}
+
+/// Convenience constant for passing no overlay options to [`composefs_fsmount()`].
+pub const NO_OVERLAY: Option<OverlayOptions<OwnedFd, OwnedFd>> = None;
```
cgwalters: This can be an associated const ref https://docs.rs/gio/latest/gio/struct.Cancellable.html#associatedconstant.NONE

crates/composefs/src/mount.rs
```diff
@@ -120,6 +122,20 @@ pub fn erofs_mount(image: OwnedFd) -> Result<OwnedFd> {
     )?)
 }
 
+/// Options for adding a writable overlay layer to a composefs mount.
```
cgwalters: There's more possible options here - `volatile` is a big one!

I'm not saying we need to do that now, but we could make this `#[non_exhaustive]` to allow us to more easily add more options later.

Or we could just support `pub options: Option<&'str>`

cgwalters (dismissed):
Looks sane to me, but I would really like to build up the pattern of having integration tests again because I think they're easy to add.

cgwalters:
One thing to bear in mind on this one too is the intersection with https://github.com/composefs/composefs-rs/pull/304 - in the general case there what I think would work well is if we supported fd passing - for the mount operation we'd just hand back a detached fd by default.

crates/composefs/src/mount.rs
alexlarsson: I made this a more generic non_exhaustive MountOptions.

alexlarsson:
I updated this to use a more generic MountOption that we can later extend.

alexlarsson:
I rebased on main and added support for varlink to this

cgwalters:
OK but you weren't replying to my "needs tests" thread. I think the token cost of doing these things via LLMs is so cheap relative to the value they provide, so I spent some tokens and pushed a commit with a test case.

alexlarsson (commented):
lgtm

cgwalters (approved):

PR #304, opened 2026-05-29
cfsctl: Add varlink RPC API and unify structured output

### LINKED ISSUE #302: cfsctl: Add a RPC varlink API
I think for 1.0 we want it so that `cfsctl` can be used via a subprocess from non-Rust languages. I think it'd even be mainly viable to have bootc *mostly* fork cfsctl as a subprocess (e.g. for the meaty stuff like listing/pulling/removing images, fsck). We'd probably still want to link the crate just for various convenience functions (e.g. uki/karg parsing) but it would probably avoid a lot of code being duplicated between bootc and cfsctl.

We could talk a lot about possible choices here, but I think varlink is OK. I have gone back and forth on this myself a ton (see e.g. https://github.com/containers/buildah/pull/6675 ). 

## FD passing

The biggest thing I had issues with. The status quo right now is systemd does this, but it's not part of the varlink standard, and when I looked their implementation has some issues. One thing at least that is nice about varlink is it does support `upgrade` to cut over to other protocols - it's probably not fatal to have anything we do that needs fd passing cut over to jsonrpc-fdpass.

cgwalters:
We previously had ad-hoc `--json` on some CLI invocations. Let's
offer a varlink API instead.

Closes: https://github.com/composefs/composefs-rs/issues/302

Assisted-by: Opencode (Various models)
Signed-off-by: Colin Walters <walters@verbum.org>

### DIFF
--- a/crates/composefs-ctl/src/varlink.rs
+++ b/crates/composefs-ctl/src/varlink.rs
@@ -1,0 +1,1756 @@
+//! Varlink RPC service for `cfsctl`.
+//!
+//! Exposes a subset of repository operations over a Unix-socket varlink
+//! interface (`org.composefs.Repository`) so that integration tests and
+//! external callers can consume structured replies instead of scraping the
+//! human-oriented CLI output.
+//!
+//! Repositories are accessed through opaque `u64` handles: a client calls
+//! `OpenRepository` to obtain a handle, passes it to every subsequent method,
+//! and frees it with `CloseRepository`. No repository is opened at startup, so
+//! every call must carry a handle. Each handle stores an already-opened
+//! `Repository<ObjectID>` monomorphized over the digest algorithm detected at
+//! open time, wrapped in an `Arc` so the streaming `Pull` method can move an
+//! owned clone into its `'static` reply stream.
+//!
+//! The zlink server serializes `Service::handle` calls (a single task holds
+//! one `&mut self` borrow at a time), so the handle table is a plain
+//! `HashMap` with no interior locking.
+
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use anyhow::{Context as _, Result};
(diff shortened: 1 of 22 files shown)


crates/composefs-ctl/src/varlink.rs
```diff
@@ -0,0 +1,1645 @@
...
+    use composefs::fsverity::{Sha256HashValue, Sha512HashValue};
+
+    #[zlink::service(
+        interface = "org.composefs.Repository",
+        vendor = "org.composefs",
+        product = "cfsctl",
+        version = "0.1.0",
+        url = "https://github.com/composefs/composefs-rs"
+    )]
+    impl<Sock> CfsctlService {
```
Johan-Liebert1, line 631: We should also implement `org.varlink.service.GetInfo`. The crate supports this https://docs.rs/zlink/latest/zlink/attr.service.html#automatic-introspection-support

Johan-Liebert1: Maybe we'd also want to support initializing a repo?

Johan-Liebert1 (dismissed):
Tested basic operations. LGTM

alexlarsson:
What is the expected use of this? Like, who will launch the varlink server, and what users will it run as? i.e. It it meant for a consumer to start an instance of this (as the same uid) and then communicate with it to avoid having to call rust APIs, or do we expect there to be a system global instance of this?

I ask, because having a server do network i/o (like the oci pull) is complex wrt things like security, proxy config, etc, and passing arbitrary paths like we do in open_repo seems very dangerous if it is a global service.

cgwalters:
> What is the expected use of this? Like, who will launch the varlink server, and what users will it run as? i.e. It it meant for a consumer to start an instance of this (as the same uid) and then communicate with it to avoid having to call rust APIs

Just this.

>  or do we expect there to be a system global instance of this?

Not this.

Johan-Liebert1 (approved):

PR #305, opened 2026-06-02
Add support for cfsctl oci apply-delta

alexlarsson:
This allows applying an oci-delta tarfile to create an oci image based on a delta and a previous oci image in the repo.

For details on oci-delta, see: https://github.com/containers/oci-delta

### DIFF
--- a/crates/composefs-oci/src/delta.rs
+++ b/crates/composefs-oci/src/delta.rs
@@ -1,0 +1,989 @@
+//! OCI delta support: detect and apply oci-delta artifacts during pull.
+//!
+//! An oci-delta artifact has `artifactType` set to the delta media type.
+//! Its layers contain the target image manifest, config, and changed layer
+//! blobs (as tar-diff patches or original gzip layers). Layers identical
+//! between source and target (by diff_id) are omitted from the delta.
+//! For more information, see https://github.com/containers/oci-delta
+
+use std::collections::HashMap;
+use std::fs::File;
+use std::future::Future;
+use std::io::{self, BufRead, BufReader, Cursor, Read, Seek, SeekFrom, Write};
+use std::path::Path;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::thread::available_parallelism;
+
+use anyhow::{Context, Result, bail, ensure};
+use composefs::erofs::reader::erofs_to_filesystem;
+use composefs::fsverity::FsVerityHashValue;
+use composefs::repository::Repository;
+use composefs::tree::RegularFile;
+use containers_image_proxy::oci_spec::image::{
+    Digest as OciDigest, DigestAlgorithm, ImageConfiguration, ImageManifest, MediaType,
+};
+
+use tokio::sync::Semaphore;
+use tokio::task::JoinSet;
+
+use crate::oci_image;
+use crate::progress::{ComponentId, ProgressEvent, ProgressUnit, SharedReporter};
+use crate::skopeo::PullResult;
+use crate::{ImportStats, layer_identifier};
+
+pub(crate) const MEDIA_TYPE_DELTA: &str = "application/vnd.io.github.containers.oci-delta.v1";
+fn media_type_tar_diff() -> MediaType {
+    MediaType::Other("application/vnd.tar-diff".to_string())
+}
+const ANNOTATION_DELTA_SOURCE_CONFIG: &str = "io.github.containers.delta.source-config";
+const ANNOTATION_DELTA_TO: &str = "io.github.containers.delta.to";
+const ANNOTATION_DELTA_CONTENT: &str = "io.github.containers.delta.content";
+
+const TAR_DIFF_HEADER: &[u8; 8] = b"tardf1\n\0";
+
+// tar-diff opcodes
+const OP_DATA: u8 = 0;
+const OP_OPEN: u8 = 1;
+const OP_COPY: u8 = 2;
+const OP_ADD_DATA: u8 = 3;
+const OP_SEEK: u8 = 4;
+
+// DoS protection limits from the Go tar-patch reference implementation
+const MAX_FILENAME_SIZE: u64 = 4 * 1024;
+const MAX_ADD_DATA_SIZE: u64 = 100 * 1024 * 1024;
+
+// ─── Blob reader trait ──────────────────────────────────────────────────────
+
+/// Read blobs from a delta artifact by digest.
+///
+/// Implemented for OCI layout directories and pre-fetched blob maps
+/// (used by the skopeo proxy path which fetches blobs asynchronously).
+pub(crate) trait DeltaBlobReader: Send + Sync {
+    /// Open a blob for reading by digest.
+    /// For local storage this opens the file directly. For remote transports
+    /// this fetches the blob to a local temp file first.
+    fn open_blob(
+        &self,
+        digest: &OciDigest,
+    ) -> Pin<Box<dyn Future<Output = Result<File>> + Send + '_>>;
+}
+
+/// Check whether an OCI manifest is a delta artifact.
+pub(crate) fn is_delta_artifact(manifest: &ImageManifest) -> bool {
+    manifest
+        .artifact_type()
+        .as_ref()
+        .is_some_and(|t| t.to_string() == MEDIA_TYPE_DELTA)
+}
+
+// ─── Composefs-backed data source for tar-patch ─────────────────────────────
+
+/// Shared source image state: the parsed EROFS filesystem and the repository.
+struct SourceImage<ObjectID: FsVerityHashValue> {
+    fs: composefs::tree::FileSystem<ObjectID>,
+    repo: Arc<Repository<ObjectID>>,
+}
+
+enum CurrentFile {
+    Inline(Cursor<Vec<u8>>),
+    External(File),
+}
+
+impl Read for CurrentFile {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        match self {
+            CurrentFile::Inline(c) => c.read(buf),
+            CurrentFile::External(f) => f.read(buf),
+        }
+    }
+}
+
+impl Seek for CurrentFile {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        match self {
+            CurrentFile::Inline(c) => c.seek(pos),
+            CurrentFile::External(f) => f.seek(pos),
+        }
+    }
+}
+
+/// Per-task mutable cursor into the shared source image filesystem.
+struct ComposeFsDataSource<ObjectID: FsVerityHashValue> {
+    source: Arc<SourceImage<ObjectID>>,
+    current: Option<CurrentFile>,
+}
+
+impl<ObjectID: FsVerityHashValue> ComposeFsDataSource<ObjectID> {
+    fn set_current_file(&mut self, path: &str) -> Result<()> {
+        let path = Path::new(path);
+        let (dir, filename) = self
+            .source
+            .fs
+            .root
+            .split(path.as_os_str())
+            .with_context(|| format!("Source file not found: {}", path.display()))?;
+        let file = dir
+            .get_file(filename, &self.source.fs.leaves)
+            .with_context(|| format!("Source file not found: {}", path.display()))?;
+
+        self.current = Some(match file {
+            RegularFile::Inline(data) => CurrentFile::Inline(Cursor::new(data.to_vec())),
+            RegularFile::External(id, _size) => {
+                let fd = self
+                    .source
+                    .repo
+                    .open_object(id)
+                    .with_context(|| format!("Opening source object for {}", path.display()))?;
+                CurrentFile::External(File::from(fd))
+            }
+        });
+        Ok(())
+    }
+
+    fn read_exact_current(&mut self, buf: &mut [u8]) -> Result<()> {
+        let current = self
+            .current
+            .as_mut()
+            .context("No current file set in data source")?;
+        current.read_exact(buf)?;
+        Ok(())
+    }
+
+    fn seek_current(&mut self, offset: u64) -> Result<u64> {
+        let current = self
+            .current
+            .as_mut()
+            .context("No current file set in data source")?;
+        Ok(current.seek(SeekFrom::Start(offset))?)
+    }
+
+    fn copy_to(&mut self, dst: &mut impl Write, n: u64) -> Result<()> {
+        let current = self
+            .current
+            .as_mut()
+            .context("No current file set in data source")?;
+        let copied = io::copy(&mut Read::by_ref(current).take(n), dst)?;
+        ensure!(
+            copied == n,
+            "Short read from data source: expected {n}, got {copied}"
+        );
+        Ok(())
+    }
+}
+
+// ─── Tar-patch apply ────────────────────────────────────────────────────────
+
+fn read_uvarint(r: &mut impl io::BufRead) -> Result<u64> {
+    let mut result: u64 = 0;
+    let mut shift: u8 = 0;
+    loop {
+        let mut byte = [0u8; 1];
+        r.read_exact(&mut byte)?;
+        let bits = (byte[0] & 0x7f) as u64;
+        ensure!(
+            shift < 64 && bits <= (u64::MAX >> shift),
+            "uvarint overflow"
+        );
+        result |= bits << shift;
+        if byte[0] & 0x80 == 0 {
+            return Ok(result);
+        }
+        shift = shift.checked_add(7).context("uvarint overflow")?;
+    }
+}
+
+enum OciHasher {
+    Sha256(sha2::Sha256),
+    Sha384(sha2::Sha384),
+    Sha512(sha2::Sha512),
+}
+
+impl OciHasher {
+    fn new(algorithm: &DigestAlgorithm) -> Result<Self> {
+        use sha2::Digest;
+        match algorithm {
+            &DigestAlgorithm::Sha256 => Ok(Self::Sha256(sha2::Sha256::new())),
+            &DigestAlgorithm::Sha384 => Ok(Self::Sha384(sha2::Sha384::new())),
+            &DigestAlgorithm::Sha512 => Ok(Self::Sha512(sha2::Sha512::new())),
+            other => bail!("Unsupported digest algorithm: {other}"),
+        }
+    }
+
+    fn update(&mut self, data: &[u8]) {
+        use sha2::Digest;
+        match self {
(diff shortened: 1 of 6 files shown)


crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+impl DeltaFile {
+    fn open(path: &Path) -> Result<Self> {
+        let file = File::open(path).with_context(|| format!("Opening delta file {}", path.display()))?;
+        let mut blobs = HashMap::new();
+
+        let mut archive = ::tar::Archive::new(BufReader::new(file));
+        for entry in archive.entries().context("Reading delta tar entries")? {
+            let entry = entry.context("Reading tar entry")?;
+            let path_bytes = entry.path_bytes();
+            let name = String::from_utf8_lossy(&path_bytes).into_owned();
```
cgwalters: utf8_lossy is almost always a bad idea. I think we want `PathBuf` here.

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+
+/// Per-task mutable cursor into the shared [`SourceFileIndex`].
+struct ComposeFsDataSource<ObjectID: FsVerityHashValue> {
+    index: Arc<SourceFileIndex<ObjectID>>,
+    current: Option<CurrentFile>,
+}
+
+fn clean_path(path: &str) -> String {
+    let mut parts = Vec::new();
+    for component in path.split('/') {
```
cgwalters: There's better ways to do this iterating over https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.components - some similar code in bootc/ostree for example.

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+// ─── Tar-patch apply ────────────────────────────────────────────────────────
+
+fn read_uvarint(r: &mut impl io::BufRead) -> Result<u64> {
+    let mut result: u64 = 0;
+    let mut shift = 0u32;
+    loop {
+        let mut byte = [0u8; 1];
+        r.read_exact(&mut byte)?;
+        let b = byte[0];
+        result |= ((b & 0x7f) as u64) << shift;
```
cgwalters: Let's use checked arithmetic by default

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+        if b & 0x80 == 0 {
+            return Ok(result);
+        }
+        shift += 7;
+        ensure!(shift < 64, "uvarint overflow");
```
cgwalters: Then shift could be u8 right? Might as well constrain it as much as the language allows...

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+}
+
+// ─── Tar-patch apply ────────────────────────────────────────────────────────
+
+fn read_uvarint(r: &mut impl io::BufRead) -> Result<u64> {
+    let mut result: u64 = 0;
+    let mut shift = 0u32;
+    loop {
+        let mut byte = [0u8; 1];
+        r.read_exact(&mut byte)?;
```
cgwalters, line 183: I think it'd be more efficient to use https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#tymethod.fill_buf

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+    Ok(())
+}
+
+// ─── Apply delta ────────────────────────────────────────────────────────────
+
+/// Reconstruct a single layer's uncompressed tar from a delta blob.
+fn reconstruct_layer<ObjectID: FsVerityHashValue>(
+    delta_file: &DeltaFile,
+    source_index: &Arc<SourceFileIndex<ObjectID>>,
+    blob_digest: &str,
```
cgwalters: Let's prefer the oci_spec digest types

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+                let tar_buf = tokio::task::spawn_blocking({
+                    let diff_id = diff_id.clone();
+                    move || -> Result<Vec<u8>> {
+                        reconstruct_layer(&delta.delta_file, &source_index, &blob_digest, is_tar_diff)
+                            .with_context(|| format!("Reconstructing layer {diff_id}"))
+                    }
+                }).await??;
+
+                // Verify diff_id
+                let mut hasher = Sha256::new();
```
cgwalters: Shouldn't hardcode sha256

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+
+// ─── Apply delta ────────────────────────────────────────────────────────────
+
+/// Reconstruct a single layer's uncompressed tar from a delta blob.
+fn reconstruct_layer<ObjectID: FsVerityHashValue>(
+    delta_file: &DeltaFile,
+    source_index: &Arc<SourceFileIndex<ObjectID>>,
+    blob_digest: &str,
+    is_tar_diff: bool,
+) -> Result<Vec<u8>> {
```
cgwalters: It'd be more efficient to use an `O_TMPFILE` for this right?

crates/composefs-oci/testdata/delta/generate.sh
```diff
@@ -0,0 +1,146 @@
...
+    size=$(stat -c%s "${tar_file}.gz")
+    echo "$name $diff_id $digest $size ${tar_file}.gz"
+}
+
+read -r _ L1_DIFFID L1_DIGEST L1_SIZE L1_GZ <<< "$(compress_layer "$WORK/layer1.tar" layer1)"
+read -r _ L2S_DIFFID L2S_DIGEST L2S_SIZE L2S_GZ <<< "$(compress_layer "$WORK/layer2s.tar" layer2s)"
+read -r _ L2T_DIFFID L2T_DIGEST L2T_SIZE L2T_GZ <<< "$(compress_layer "$WORK/layer2t.tar" layer2t)"
+read -r _ L3_DIFFID L3_DIGEST L3_SIZE L3_GZ <<< "$(compress_layer "$WORK/layer3.tar" layer3)"
+
+# ── Build OCI layout helper ─────────────────────────────────────────
```
cgwalters: We have Rust code to do all of this stuff via the ocidir crate etc

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,878 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use composefs::fsverity::Sha256HashValue;
+    use composefs::test::TestRepo;
+    use std::path::PathBuf;
+
+    fn testdata_path(name: &str) -> PathBuf {
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("testdata/delta")
```
cgwalters: I'd like to not commit binary files like this to git if possible, surely we can generate them?

cgwalters (changes requested):
Just an initial style pass

crates/composefs-oci/src/delta.rs
alexlarsson: In practice OCI files will not have any non-ascii filenames, but sure.

crates/composefs-oci/src/delta.rs
alexlarsson: Its not really helpful, we do bounds checks anyway and the shift argument is u32 so we would have to cast each time its used.

crates/composefs-oci/src/delta.rs
alexlarsson: I added this, and some tests

crates/composefs-oci/src/delta.rs
alexlarsson: Not necessarily more efficient, but yes, it looks cleaner.

crates/composefs-oci/src/delta.rs
alexlarsson: I intentionally commited it to avoid depending on the external oci-delta tools. But, I made it skip the test if that is not available.

crates/composefs-oci/testdata/delta/generate.sh
alexlarsson: ok

crates/composefs-oci/src/delta.rs
alexlarsson: done

crates/composefs-oci/src/delta.rs
alexlarsson: Actually, u8 is fine, it just doesn't make sense to use a checked shift, because that checks something else (fails if n_shift > 64). I made this handle all the failure cases and added tests for them.

crates/composefs-oci/src/delta.rs
alexlarsson: done

crates/composefs-oci/src/delta.rs
alexlarsson: done, although the new OciHasher type should maybe be in some more generic place?

crates/composefs-oci/src/delta.rs
alexlarsson: done

alexlarsson:
Ok, i think I got all the comments fixed.

alexlarsson:
Note: This doesn't actually run all the new tests in CI, as there is no oci-delta binary. However it passes locally here. We can build oci-delta ourselves, or we can just wait a bit until it gets packaged up and pick it up as an rpm. (This is in progress elsewhere).

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,1095 @@
...
+        PathBuf::from(format!("blobs/{}/{}", digest.algorithm(), digest.digest()))
+    }
+
+    fn read_raw(&self, tar_path: &Path) -> Result<Vec<u8>> {
+        let entry = self
+            .blobs
+            .get(tar_path)
+            .with_context(|| format!("{} not found in delta tar", tar_path.display()))?;
+        let mut buf = vec![0u8; entry.size as usize];
+        File::open(&self.path)?.read_exact_at(&mut buf, entry.offset)?;
```
cgwalters: Seems better to support DeltaFile holding an open fd.

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,1095 @@
...
+            }
+            _ => {}
+        }
+    }
+    result
+}
+
+fn build_file_index<ObjectID: FsVerityHashValue>(
+    fs: &composefs::tree::FileSystem<ObjectID>,
+) -> HashMap<PathBuf, FileRef<ObjectID>> {
```
cgwalters: We're not talking about that many files, and I imagine it'd work equally well for us to just use `Directory::split_ref` instead of building up a hashmap-based index.

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,1095 @@
...
+        if byte[0] & 0x80 == 0 {
+            return Ok(result);
+        }
+        shift = shift.checked_add(7).context("uvarint overflow")?;
+        ensure!(shift < 64, "uvarint overflow");
+    }
+}
+
+enum OciHasher {
+    Sha256(sha2::Sha256),
```
cgwalters, line 198: I think we have code elsewhere using openssl for this. We're probably not consistent on this.

crates/composefs-ctl/src/lib.rs
```diff
@@ -431,6 +431,20 @@ enum OciCommand {
         #[clap(long)]
         cmdline: Vec<String>,
     },
+    /// Apply an OCI delta to reconstruct a target image
+    ///
+    /// The delta file must be an oci-delta tar archive. The source image
+    /// must already be present in the repository (with an EROFS image).
+    /// The result is stored as if pulled with `cfsctl oci pull`.
+    ApplyDelta {
```
cgwalters: First we should have a varlink version of this now - per other pattern the CLI can just be a wrapper for varlink.

I also wonder though...we want to support fetching deltas from a registry too right? So...this I think argues for actually wiring it into the `oci pull` path as a special case that looks at `oci-archive:` and dispatches based on mime type or so?

crates/composefs-oci/src/delta.rs
alexlarsson: I did that initially, but it caused problems in the parallel case, where fds were shared between threads, and then those shared the current offset, which broke things. Is there some pread based wrapper over File we could use for that?

crates/composefs-oci/src/delta.rs
alexlarsson: And keep the filesystem around you mean? I don't know if either is better than the other really. Does it matter?

crates/composefs-ctl/src/lib.rs
alexlarsson: oci-delta itself can actually suport oci dirs as well. We could conceptually support this as well. Not sure how hard it will be to dispatch based on the mimetype in pull though, will have a look,

crates/composefs-oci/src/delta.rs
cgwalters: https://github.com/composefs/composefs-rs/blob/552808a8c8b2af314f2d9fa6904cacc791425142/crates/composefs-splitfdstream/src/lib.rs#L266

crates/composefs-oci/src/delta.rs
cgwalters: I think it'd be good to have "centralized" filesystem traversal APIs. An advantage of walking the `Filesystem` vs the hashmap is we don't need e.g. `clean_path` to handle `//` and such.

alexlarsson: I switched this to use the Directory::* apis

alexlarsson:
Ok, I redid this to be part of "oci pull", and it seems to work well. I can even pull a delta as an artifact:
```
$ target/debug/cfsctl --repo repo oci pull docker://quay.io/alexl42/test-delta:41
$ target/debug/cfsctl --repo repo oci pull docker://quay.io/alexl42/test-delta:41to42
```

crates/composefs-oci/src/delta.rs
```diff
@@ -0,0 +1,988 @@
...
+        id: progress_id.clone(),
+        total: Some(n_layers),
+        unit: ProgressUnit::Items,
+    });
+
+    let mut layer_tasks = JoinSet::new();
+    let mut immediate_results: HashMap<usize, (OciDigest, ObjectID)> = HashMap::new();
+    let mut stats = ImportStats::default();
+
+    let threads = available_parallelism()?;
```
cgwalters: When we fetch OCI images today, we process the tar in a streaming fashion, writing objects into the composefs store as we go. Disk/memory usage is constant relative to the update size.

But one downside of deltas as designed/implemented today is that we have to download the entire delta layer to a file *and* regenerate the layer locally, so temporary disk space usage is O(delta size) plus O(layer).

I think for this reason I'd suggest we have at most 1 delta layer being downloaded, and 1 being processed at a time by default. WDYT?

cgwalters (commented):
Cool, this is looking nicer!

On tests, one thing we do in this project is copy over "fixtures" to `ghcr.io/composefs`. I think we could do the same with deltas. There's generic image mirroring logic.

That said it'd also help of course to have a static binary of oci-delta shipped from the upstream repo that we can use if installed (and install it in CI actions).

crates/composefs-oci/src/delta.rs
alexlarsson: I'm not sure how common download-and-apply will be, but in the local-only (i.e. oci dir or oci archive) case we definitely want to parallelize the apply, as it greatly speeds up applying the delta (i saw times going from 40 to 10 secs). So, lets be careful of exactly how we limit this.

alexlarsson:
I added binaries (build on cs10) to: https://github.com/containers/oci-delta/releases/tag/v0.1.0

crates/composefs-oci/src/delta.rs
cgwalters: Yes, in the `oci:` path we don't need to fetch anything at all, we should just parse from the existing files.

No harm to parallelizing the processing in that case. My concern is more about disk space usage spikes for nontrivial updates.

In the `ociarchive:` path...like ideally we can say that that file is *not* compressed (because the blobs will be, no point in double compressing them, and compressing the metadata brings almost no value).

Then we can do what this code is already doing I think, just walk the tar to find byte offsets.

I guess in the end I'm just arguing for the registry path to default to fetching one blob at a time, or maybe two.

Or maybe it's fine, and we can just make it configurable later if we need to.

(Some environment variable to control parallelism in composefs at a coarse grained level is probably enough, or perhaps just `--jobserver-fd` like https://make.mad-scientist.net/papers/jobserver-implementation/  )

alexlarsson: Ok, i will make the skopeo case do a limited parallelism, and keep the current parallelism in the oci_layout.rs case. Then I will make the oci_layout.rs case handle oci-archive as well as oci-dirs. That seems like a good approach to me.

alexlarsson: So, I did the parallelism limitation. However, I think the only reasonable approach to take the non-skopeo codepath for oci archives is to change ocidir-rs, which is gonna require some work. I think we can do that in a separate MR, oci-archive: works as is via skopeo anyway.

crates/composefs-oci/src/test_util.rs
```diff
@@ -845,6 +696,60 @@ pub fn ensure_erofs_for_image(
...
 }
 
+/// Build an OCI layout directory on disk from dumpfile layer strings.
+///
+/// Each entry in `layers` is a dumpfile string (same format as
+/// [`create_multi_layer_image`] uses).  Returns a [`tempfile::TempDir`]
+/// containing a valid OCI layout that can be read by `oci-delta`,
+/// `import_oci_layout`, or any OCI-aware tool.
+#[cfg(test)]
+pub fn build_oci_layout(layers: &[&str]) -> tempfile::TempDir {
```
cgwalters, line 706: I have a gut feeling this is like the 6th copy of this, may overlap with https://github.com/composefs/composefs-rs/pull/312

cgwalters (approved):

alexlarsson:
This is the start of the work to support skopeo-less oci-archive deltas: 
 https://github.com/bootc-dev/ocidir-rs/pull/78

cgwalters:
Joseph and I had a live chat on this, I think it really makes sense for us to support registry discovery of these (via referrers API with fallback to cosign-style sha digest discovery).

xref https://github.com/podman-container-tools/container-libs/issues/221#issuecomment-4856452838 - basically we are probably going to need to add support for accessing referrers via skopeo

PR #307, opened 2026-06-05
Fix: Avoid a hard panic and use a recoverable error instead

henrywang:
Issue: Arch Linux container build where pacman -Syu in a downstream stage installs a newer kernel version
(with nativevmlinuz in `/usr/lib/modules/<kver>/`) but no corresponding `initramfs`.

`UsrLibModulesVmlinuz::load_all` calls `self.initramfs.unwrap()` which panics instead of returning an error when a kernel directory has vmlinuz but no initramfs

fix: change `unwrap()` to `ok_or_else(|| anyhow!("no initramfs for kernel {kver}"))` and propagate

Two tests:
- `test_into_type1_with_initramfs` — verifies the right path produces entries with the correct vmlinuz and initramfs
- `test_into_type1_without_initramfs_returns_error` — verifies None initramfs returns an error containing "no initramfs" rather than panic

### DIFF
--- a/crates/composefs-boot/src/bootloader.rs
+++ b/crates/composefs-boot/src/bootloader.rs
@@ -457,9 +457,17 @@ impl<ObjectID: FsVerityHashValue> UsrLibModulesVmlinuz<ObjectID> {
     ///
     /// # Returns
     ///
-    /// A Type1Entry with generated BLS configuration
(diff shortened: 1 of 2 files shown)


Johan-Liebert1 (dismissed):
Thanks for this

Johan-Liebert1:
Needs `cargo fmt`

Johan-Liebert1 (approved):

PR #308, opened 2026-06-05
Add composefs-run a minimal crun based container starter

alexlarsson:
This is an experiment to see how a composefs native podman run would look. Its around 700 lines, and depends only on crun. It supports a pretty small subset of the "podman run" options, but the ones it has actually cover a great part of the things in wide use.

All containers are stateless on the host. There is no database of running containers, etc. All the state for the container are in a tmpfs that is mounted in a namespace and freed automatically on container exit. By default the writable layer of the container is in thet tmpfs as well, but you can pass separate
--overlayfs-upperdir/workdir options to store the state somewhere.

**Note**: I'm not sure this is something we want, and if we do, if this is the right places for it (in a separate binary). I'm just putting this out as an idea.

### DIFF
--- a/crates/composefs-run/src/main.rs
+++ b/crates/composefs-run/src/main.rs
@@ -1,0 +1,742 @@
+//! Minimal container runner using composefs images and crun.
+
+use std::fs;
+use std::io::IsTerminal;
+use std::os::unix::process::CommandExt;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use anyhow::{Context, Result, ensure};
+use clap::Parser;
+use composefs::fsverity::{FsVerityHashValue, Sha256HashValue, Sha512HashValue};
+use composefs::mount::MountOptions;
+use composefs::repository::{Repository, read_repo_algorithm};
+use composefs_oci::OciDigest;
+use oci_spec::runtime::{
+    Capability, LinuxBuilder, LinuxCapabilitiesBuilder, LinuxNamespace, LinuxNamespaceType,
+    MountBuilder, ProcessBuilder, RootBuilder, Spec, UserBuilder,
+};
+use rustix::fs::{CWD, Mode, OFlags};
+use rustix::mount::{
+    FsMountFlags, FsOpenFlags, MountAttrFlags, MountPropagationFlags, fsconfig_create, fsmount,
+    fsopen,
+};
+
+/// Minimal container runner using composefs images and crun.
+#[derive(Debug, Parser)]
+#[clap(name = "composefs-run")]
+struct Cli {
+    /// Path to the composefs repository
+    #[clap(long, default_value = "/sysroot/composefs")]
+    repo: PathBuf,
+
+    /// Mount the rootfs read-only
+    #[clap(long)]
+    read_only: bool,
+
+    /// When --read-only is set, mount tmpfs on /dev/shm, /run, /tmp, /var/tmp
+    #[clap(long, default_value = "true", value_parser = clap::builder::BoolishValueParser::new())]
+    read_only_tmpfs: bool,
+
+    /// Persistent upper layer directory for overlayfs (default: tmpfs)
+    #[clap(long = "overlay-upperdir", requires = "overlay_workdir")]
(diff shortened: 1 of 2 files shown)


alexlarsson:
In use:
```
$ cfsctl --repo repo oci pull docker://quay.io/centos/centos:10 cs10
$ sudo composefs-run --repo repo -it cs10 -- head -2 /etc/os-release 
NAME="CentOS Stream"
VERSION="10 (Coughlan)"
```

crates/composefs-run/src/main.rs
```diff
@@ -0,0 +1,742 @@
...
+        .context("unshare(CLONE_NEWNS) — are you running as root?")?;
+
+    rustix::mount::mount_change(
+        "/",
+        MountPropagationFlags::REC | MountPropagationFlags::PRIVATE,
+    )
+    .context("Making / recursively private")?;
+
+    // Attach the bundle tmpfs and composefs rootfs mount
+    let bundle_dir = "/run/composefs-run";
```
cgwalters, line 257: I don't think this hardcoded path is a good idea.

When I looked at the OCI runtime spec...I wondered why the heck it isn't supported to pass a file descriptor for the rootfs mount in a clear way.

Anyways surely at least it's work-aroudable via passing a `/proc/self/fd/N`?

crates/composefs-run/src/main.rs
```diff
@@ -0,0 +1,742 @@
...
+    let pidfile_path = match pidfile {
+        Some(pf) => pf.to_owned(),
+        None => {
+            tmp_pidfile = format!("{bundle_dir}/container.pid");
+            PathBuf::from(&tmp_pidfile)
+        }
+    };
+
+    // crun create: set up the container but don't start it yet
+    let status = Command::new("crun")
```
cgwalters, line 360: Presumably this could also use https://github.com/youki-dev/youki - probably worth having minimal support for configuring it

crates/composefs-run/src/main.rs
```diff
@@ -0,0 +1,742 @@
...
+    println!("{}", pid.trim());
+
+    Ok(())
+}
+
+fn resolve_image<ObjectID: FsVerityHashValue>(
+    repo: &Repository<ObjectID>,
+    image: &str,
+) -> Result<(ObjectID, composefs_oci::OpenConfig<ObjectID>)> {
+    let img = if let Some(digest_str) = image.strip_prefix('@') {
```
cgwalters, line 393: Let's use https://docs.rs/oci-spec/latest/oci_spec/distribution/struct.Reference.html ?

cgwalters (commented):
I did a quick PoC of a similar thing in https://github.com/composefs/composefs-rs/pull/306/changes/c0bac64b77e7a2fa6e0ea4c7cb0d16e71929957c

~In Python primarily to demonstrate the cross-process thing via varlink; I think it'd make sense to do the same thing here~ sorry that's a lie that was a previous iteration

crates/composefs-run/src/main.rs
alexlarsson: I tried to avoid it using /proc/self/fd/N but it didn't work for pivot root in crun. But, we could really use *any* pre-existing directory for this really.

cgwalters:
> Note: I'm not sure this is something we want, and if we do, if this is the right places for it (in a separate binary). I'm just putting this out as an idea.

I think it'd make total sense to have a separate `composefs-examples` repo? WDYT?

alexlarsson:
@cgwalters composefs-examples sounds like a good idea to me.

alexlarsson:
I created a standalone version of this as cfsrun:
https://github.com/alexlarsson/composefs-run

It now also support networking, rootless (using fuse), etc. Its really a pretty complete version of podman run for local use. I think it makes sense to put this in the composefs org?

cgwalters:
Yep you're an Owner in this org so I think you can just transfer from my PoV.

That said I'd probably make very clear in README.md it's just a PoC at this time.

cgwalters:
Closing since I think we're in agreement to have this in another repo

PR #312, opened 2026-06-09
test_util: expose and extend OCI image builder API

cgwalters:
Expose the core building blocks so downstream crates (e.g. bootc) can use them under feature = "test":

- dumpfile_to_tar() and create_multi_layer_image() made pub
- pub const LAYER_SELINUX: a minimal targeted policy tree (/etc/selinux/config + file_contexts for /, /usr, /etc, /boot, /var)

Replace the ad-hoc create_bootable_image(version: u32) / create_selinux_bootable_image sprawl with a clean builder:

  OsImage::bootable(KernelVersion::V1)
      .with_selinux()
      .with_layer("/usr/lib/os-release 0 100644 ...")
      .build_oci(repo, Some("test:v1"))
      .await

OsProfile (Minimal / Bootable { version: KernelVersion }) and OsFeatures { selinux } replace the growing list of separate functions. with_layer(dumpfile) appends an arbitrary dumpfile layer on top of all profile and feature layers — the primary hook for callers that need to inject specific files without assembling a full image from scratch. build_filesystem(repo) returns a FileSystem<Sha256HashValue> suitable for passing to transform_for_boot or selabel.

The existing create_base_image / create_bootable_image free functions are kept as thin wrappers for zero call-site churn.

Assisted-by: OpenCode (Claude Sonnet 4.6)

### DIFF
--- a/crates/composefs-oci/src/test_util.rs
+++ b/crates/composefs-oci/src/test_util.rs
@@ -84,7 +84,7 @@ fn append_with_xattrs<W: std::io::Write>(
 ///
 /// External files (`Item::Regular` with no inline content) get
 /// deterministic pseudo-random data from a seeded RNG (keyed on file size).
-fn dumpfile_to_tar(dumpfile: &str) -> Vec<u8> {
+pub fn dumpfile_to_tar(dumpfile: &str) -> Vec<u8> {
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

PR #313, opened 2026-06-10
fuse: Add option for allow_other

alexlarsson:
This is not always usable in rootless applications, so allow it to be disabled.

Actually, its not even a great default, so disable by default.


giuseppe (dismissed):
LGTM,

could you run `cargo fmt` and add a `--signoff`?

alexlarsson:
@giuseppe done

giuseppe (approved):
LGTM

PR #314, opened 2026-06-15
Release 0.7.0

cgwalters:
We had some desync between the crates.io and Github releases, hence the bump from 0.4 to 0.7 in the Cargo metadata.

### DIFF
--- a/.github/workflows/bootc-revdep.yml
+++ b/.github/workflows/bootc-revdep.yml
@@ -22,6 +22,7 @@ permissions:
 jobs:
   bootc-test:
     name: Build and test bootc with local composefs-rs
+    if: never()
     runs-on: ubuntu-24.04
     timeout-minutes: 120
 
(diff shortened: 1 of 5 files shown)


alexlarsson (dismissed):
I think this is fine. But I wonder if at some point we should bump the version above the C version number to make the replacement commands look "newer".

cgwalters:
> I think this is fine. But I wonder if at some point we should bump the version above the C version number to make the replacement commands look "newer".

Definitely, I think when we're ready to pull the trigger on that full replacement, which I think mostly blocks on the shared library and then small mechanics like copying the man pages over etc.

alexlarsson:
When we're at that point, maybe we should just bite the bullet and call it 2.0?

cgwalters:
OK basically we now have full CI interlock in the revdep test - it's really annoying to fix this because of the semver version handling in cargo. Ideally we sprint to 1.0 and stabilize the Rust API and then this all gets easier.

giuseppe (approved):
LGTM

PR #315, opened 2026-06-17
Use the new ocidir-rs archive support

alexlarsson:
With this we can pull (both regular and delta) images directly from an oci-archive without using the container image proxy.

NOTE: This is draft, because the first commit updates some dependencies to random commits. We really need a new release of oci-dir and containers-image-proxy for this to be clean.

### DIFF
--- a/crates/composefs-oci/src/oci_layout.rs
+++ b/crates/composefs-oci/src/oci_layout.rs
@@ -19,7 +19,6 @@ use std::io::Read;
 use std::path::Path;
 use std::pin::Pin;
 use std::sync::Arc;
-use std::task::{Context as TaskContext, Poll};
 use std::thread::available_parallelism;
 
 use anyhow::{Context, Result};
@@ -30,8 +29,10 @@ use containers_image_proxy::oci_spec::image::{
 use fn_error_context::context;
 use ocidir::prelude::*;
 use ocidir::{OciArchive, OciDir, ResolvedManifest};
+use tokio::io::DuplexStream;
 use tokio::sync::Semaphore;
 use tokio::task::JoinSet;
+use tokio_util::io::SyncIoBridge;
 use tracing::debug;
 
 use composefs::fsverity::FsVerityHashValue;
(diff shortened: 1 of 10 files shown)


cgwalters:
Looks sane offhand, but let's get releases of the dependent crates out.

alexlarsson:
@cgwalters I really thought there would be something like the BlockingReader helper somewhere already, but I was unable to find it...

crates/composefs-oci/src/oci_layout.rs
```diff
@@ -42,6 +44,87 @@ use crate::{ImportStats, config_identifier, layer_identifier};
 
 use crate::skopeo::PullResult;
 
+const READ_BUF_SIZE: usize = 128 * 1024;
+
+/// Adapts a synchronous `Read` into a `tokio::io::AsyncRead` by dispatching
+/// reads to [`tokio::task::spawn_blocking`], similar to `tokio::fs::File`.
```
cgwalters: This one is interesting; way back in the past I hit this in the other direction, see https://users.rust-lang.org/t/best-practices-for-bridging-async-and-sync-particularly-read-write/58994 and we ended up adding some tokio helpers.

This case is the inverse though.

I think implementing the low-level `poll_read` API requires a lot of thought about correctness.

In this case, instead of a `spawn_blocking` per read call, what I think would be more efficient is basically having a pipe:

- spawn exactly one spawn_blocking task that uses sync IO and writes to a channel (stream)
- A stream of bytes can be converted to AsyncRead https://docs.rs/tokio-stream/latest/tokio_stream/#conversion-to-and-from-asyncreadasyncwrite

alexlarsson: Ok, I did something like this, as a separate commit so you can look at it. For the final version I'll delete the BlockingReader commit and squash this into the other commit.

alexlarsson:
I refreshed this, including some fixes for test failures and (hopefully) enabling the delta tests in CI.

alexlarsson:
rebased on main

giuseppe (approved):
LGTM

PR #321, opened 2026-06-23
crates: add composefs-capi cdylib crate

giuseppe:
It provides the libcomposefs C API.

Add a new crate that produces a drop-in libcomposefs.so shared library backed by the Rust composefs implementation. This enables C-only consumers to use the Rust codebase without source changes.

The crate ships the original C headers unchanged and includes a pkg-config template.

I've not reviewed carefully the code generated by Claude, before I do it, I'd like to gather some comments, and if this is a viable option to replace libcomposefs.

@alexlarsson @cgwalters PTAL

### DIFF
--- a/crates/composefs-capi/include/libcomposefs/private/erofs_fs.h
+++ b/crates/composefs-capi/include/libcomposefs/private/erofs_fs.h
@@ -1,0 +1,461 @@
+/* SPDX-License-Identifier: GPL-2.0-only OR Apache-2.0 */
+/*
+ * EROFS (Enhanced ROM File System) on-disk format definition
+ *
+ * Copyright (C) 2017-2018 HUAWEI, Inc.
+ *             https://www.huawei.com/
+ * Copyright (C) 2021, Alibaba Cloud
+ */
+#ifndef __EROFS_FS_H
+#define __EROFS_FS_H
+
+#define EROFS_SUPER_OFFSET      1024
+
+#define EROFS_FEATURE_COMPAT_SB_CHKSUM          0x00000001
+#define EROFS_FEATURE_COMPAT_MTIME              0x00000002
+#define EROFS_FEATURE_COMPAT_XATTR_FILTER	0x00000004
+
+/*
+ * Any bits that aren't in EROFS_ALL_FEATURE_INCOMPAT should
+ * be incompatible with this kernel version.
+ */
+#define EROFS_FEATURE_INCOMPAT_ZERO_PADDING	0x00000001
+#define EROFS_FEATURE_INCOMPAT_COMPR_CFGS	0x00000002
+#define EROFS_FEATURE_INCOMPAT_BIG_PCLUSTER	0x00000002
+#define EROFS_FEATURE_INCOMPAT_CHUNKED_FILE	0x00000004
+#define EROFS_FEATURE_INCOMPAT_DEVICE_TABLE	0x00000008
+#define EROFS_FEATURE_INCOMPAT_COMPR_HEAD2	0x00000008
+#define EROFS_FEATURE_INCOMPAT_ZTAILPACKING	0x00000010
+#define EROFS_FEATURE_INCOMPAT_FRAGMENTS	0x00000020
+#define EROFS_FEATURE_INCOMPAT_DEDUPE		0x00000020
+#define EROFS_FEATURE_INCOMPAT_XATTR_PREFIXES	0x00000040
+#define EROFS_ALL_FEATURE_INCOMPAT		\
+	(EROFS_FEATURE_INCOMPAT_ZERO_PADDING | \
+	 EROFS_FEATURE_INCOMPAT_COMPR_CFGS | \
+	 EROFS_FEATURE_INCOMPAT_BIG_PCLUSTER | \
+	 EROFS_FEATURE_INCOMPAT_CHUNKED_FILE | \
+	 EROFS_FEATURE_INCOMPAT_DEVICE_TABLE | \
+	 EROFS_FEATURE_INCOMPAT_COMPR_HEAD2 | \
+	 EROFS_FEATURE_INCOMPAT_ZTAILPACKING | \
+	 EROFS_FEATURE_INCOMPAT_FRAGMENTS | \
+	 EROFS_FEATURE_INCOMPAT_DEDUPE | \
+	 EROFS_FEATURE_INCOMPAT_XATTR_PREFIXES)
+
+#define EROFS_SB_EXTSLOT_SIZE	16
+
+struct erofs_deviceslot {
+	u8 tag[64];		/* digest(sha256), etc. */
+	__le32 blocks;		/* total fs blocks of this device */
+	__le32 mapped_blkaddr;	/* map starting at mapped_blkaddr */
+	u8 reserved[56];
+};
+#define EROFS_DEVT_SLOT_SIZE	sizeof(struct erofs_deviceslot)
+
+/* erofs on-disk super block (currently 128 bytes) */
+struct erofs_super_block {
+	__le32 magic;           /* file system magic number */
+	__le32 checksum;        /* crc32c(super_block) */
+	__le32 feature_compat;
+	__u8 blkszbits;         /* filesystem block size in bit shift */
+	__u8 sb_extslots;	/* superblock size = 128 + sb_extslots * 16 */
+
+	__le16 root_nid;	/* nid of root directory */
+	__le64 inos;            /* total valid ino # (== f_files - f_favail) */
+
+	__le64 build_time;      /* compact inode time derivation */
+	__le32 build_time_nsec;	/* compact inode time derivation in ns scale */
+	__le32 blocks;          /* used for statfs */
+	__le32 meta_blkaddr;	/* start block address of metadata area */
+	__le32 xattr_blkaddr;	/* start block address of shared xattr area */
+	__u8 uuid[16];          /* 128-bit uuid for volume */
+	__u8 volume_name[16];   /* volume name */
+	__le32 feature_incompat;
+	union {
+		/* bitmap for available compression algorithms */
+		__le16 available_compr_algs;
+		/* customized sliding window size instead of 64k by default */
+		__le16 lz4_max_distance;
+	} __packed u1;
+	__le16 extra_devices;	/* # of devices besides the primary device */
+	__le16 devt_slotoff;	/* startoff = devt_slotoff * devt_slotsize */
+	__u8 dirblkbits;	/* directory block size in bit shift */
+	__u8 xattr_prefix_count;	/* # of long xattr name prefixes */
+	__le32 xattr_prefix_start;	/* start of long xattr prefixes */
+	__le64 packed_nid;	/* nid of the special packed inode */
+	__u8 xattr_filter_reserved; /* reserved for xattr name filter */
+	__u8 reserved2[23];
+};
+
+/*
+ * EROFS inode datalayout (i_format in on-disk inode):
+ * 0 - uncompressed flat inode without tail-packing inline data:
+ * 1 - compressed inode with non-compact indexes:
+ * 2 - uncompressed flat inode with tail-packing inline data:
+ * 3 - compressed inode with compact indexes:
+ * 4 - chunk-based inode with (optional) multi-device support:
+ * 5~7 - reserved
+ */
+enum {
+	EROFS_INODE_FLAT_PLAIN			= 0,
+	EROFS_INODE_COMPRESSED_FULL		= 1,
+	EROFS_INODE_FLAT_INLINE			= 2,
+	EROFS_INODE_COMPRESSED_COMPACT		= 3,
+	EROFS_INODE_CHUNK_BASED			= 4,
+	EROFS_INODE_DATALAYOUT_MAX
+};
+
+static inline bool erofs_inode_is_data_compressed(unsigned int datamode)
+{
+	return datamode == EROFS_INODE_COMPRESSED_COMPACT ||
+		datamode == EROFS_INODE_COMPRESSED_FULL;
+}
+
+/* bit definitions of inode i_format */
+#define EROFS_I_VERSION_MASK            0x01
+#define EROFS_I_DATALAYOUT_MASK         0x07
+
+#define EROFS_I_VERSION_BIT             0
+#define EROFS_I_DATALAYOUT_BIT          1
+#define EROFS_I_ALL_BIT			4
+
+#define EROFS_I_ALL	((1 << EROFS_I_ALL_BIT) - 1)
+
+/* indicate chunk blkbits, thus 'chunksize = blocksize << chunk blkbits' */
+#define EROFS_CHUNK_FORMAT_BLKBITS_MASK		0x001F
+/* with chunk indexes or just a 4-byte blkaddr array */
+#define EROFS_CHUNK_FORMAT_INDEXES		0x0020
+
+#define EROFS_CHUNK_FORMAT_ALL	\
+	(EROFS_CHUNK_FORMAT_BLKBITS_MASK | EROFS_CHUNK_FORMAT_INDEXES)
+
+/* 32-byte on-disk inode */
+#define EROFS_INODE_LAYOUT_COMPACT	0
+/* 64-byte on-disk inode */
+#define EROFS_INODE_LAYOUT_EXTENDED	1
+
+struct erofs_inode_chunk_info {
+	__le16 format;		/* chunk blkbits, etc. */
+	__le16 reserved;
+};
+
+union erofs_inode_i_u {
+	/* total compressed blocks for compressed inodes */
+	__le32 compressed_blocks;
+
+	/* block address for uncompressed flat inodes */
+	__le32 raw_blkaddr;
+
+	/* for device files, used to indicate old/new device # */
+	__le32 rdev;
+
+	/* for chunk-based files, it contains the summary info */
+	struct erofs_inode_chunk_info c;
+};
+
+/* 32-byte reduced form of an ondisk inode */
+struct erofs_inode_compact {
+	__le16 i_format;	/* inode format hints */
+
+/* 1 header + n-1 * 4 bytes inline xattr to keep continuity */
+	__le16 i_xattr_icount;
+	__le16 i_mode;
+	__le16 i_nlink;
+	__le32 i_size;
+	__le32 i_reserved;
+	union erofs_inode_i_u i_u;
+
+	__le32 i_ino;		/* only used for 32-bit stat compatibility */
+	__le16 i_uid;
+	__le16 i_gid;
+	__le32 i_reserved2;
+};
+
+/* 64-byte complete form of an ondisk inode */
+struct erofs_inode_extended {
+	__le16 i_format;	/* inode format hints */
+
+/* 1 header + n-1 * 4 bytes inline xattr to keep continuity */
+	__le16 i_xattr_icount;
+	__le16 i_mode;
+	__le16 i_reserved;
+	__le64 i_size;
+	union erofs_inode_i_u i_u;
+
+	__le32 i_ino;		/* only used for 32-bit stat compatibility */
+	__le32 i_uid;
+	__le32 i_gid;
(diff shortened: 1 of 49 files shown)


crates/composefs-ffi/src/fsverity.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn lcfs_compute_fsverity_from_fd(digest: *mut u8, fd: c_int) -> c_int {
+    if digest.is_null() || fd < 0 {
+        set_errno(libc::EINVAL);
+        return -1;
+    }
+
+    let dup_fd = unsafe { libc::dup(fd) };
```
cgwalters: I think it's cleaner to use https://doc.rust-lang.org/stable/std/os/fd/struct.BorrowedFd.html#method.borrow_raw

crates/composefs-capi/src/fsverity.rs
```diff
@@ -0,0 +1,256 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn lcfs_compute_fsverity_from_data(
+    digest: *mut u8,
+    data: *mut u8,
+    data_len: usize,
+) -> c_int {
+    if digest.is_null() || data.is_null() {
+        set_errno(libc::EINVAL);
+        return -1;
+    }
```
cgwalters, line 33: More elegant I think to use e.g. https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#method.new like

```
let Some(digest) = NonNull::new(digest) else { set_errno(libc::EINVAL); return -1 };
```
could wrap in a macro too

crates/composefs-ffi/src/image.rs
```diff
@@ -0,0 +1,500 @@
...
+pub unsafe extern "C" fn lcfs_write_to(
+    root: *mut FfiNode,
+    options: *mut LcfsWriteOptions,
+) -> c_int {
+    if root.is_null() || options.is_null() {
+        set_errno(libc::EINVAL);
+        return -1;
+    }
+
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| unsafe {
```
cgwalters: I would personally prefer to compile the Rust code with panic=abort. That's what we do in rpm-ostree.

crates/composefs-ffi/src/image.rs
```diff
@@ -0,0 +1,500 @@
...
+            }
+
+            let child = lcfs_load_node_from_file(dirfd, entry_name.as_ptr(), buildflags);
+            if child.is_null() {
+                continue;
+            }
+
+            // If it's a directory, recurse
+            if ((*child).mode & libc::S_IFMT) == libc::S_IFDIR {
+                let child_fd = libc::openat(
```
cgwalters: I know this is a translation, but there's nothing that stops us from using say rustix here instead so we get safety (including `Drop` handling).

crates/composefs-ffi/src/image.rs
```diff
@@ -0,0 +1,500 @@
...
+            None => {
+                set_errno(libc::EIO);
+                ptr::null_mut()
+            }
+        }
+    }
+}
+
+unsafe fn build_dir_recursive(dirfd: c_int, buildflags: c_int) -> Option<*mut FfiNode> {
+    unsafe {
```
cgwalters: I am thinking it might actually be cleaner to have this whole function be implemented by using the rust-side Filesystem API and then having a single mapping between Filesystem -> FfiNode ?

crates/composefs-ffi/src/mount.rs
```diff
@@ -0,0 +1,166 @@
...
+    options: *mut LcfsMountOptions,
+) -> c_int {
+    if path.is_null() || mountpoint.is_null() {
+        set_errno(libc::EINVAL);
+        return -1;
+    }
+
+    unsafe {
+        let path_cstr = CStr::from_ptr(path);
+        let path_str = match path_cstr.to_str() {
```
cgwalters: We don't need to require the paths are utf8

crates/composefs-ffi/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    uid: u32,
+    gid: u32,
+    nlink: u32,
+    rdev: u64,
+    size: u64,
+    mtime_sec: i64,
+    mtime_nsec: u32,
+    digest: [u8; LCFS_DIGEST_SIZE],
+    digest_set: bool,
+    hardlink_target: *mut FfiNode,
```
cgwalters: This case can use `Option<NonNull<FfiNode>>`

crates/composefs-ffi/src/mount.rs
```diff
@@ -0,0 +1,166 @@
...
+            Ok(s) => s,
+            Err(_) => {
```
cgwalters: Here and elsewhere I think it'd be cleaner to have a helper macro or closure which handles this Err -> errno+return. I think we have code for that in rpm-ostree.

crates/composefs-ffi/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "composefs-ffi"
```
cgwalters: I'd kind of prefer the name `-capi`, there's not a really strong convention for this though.

cgwalters (commented):
Thanks for this! I took a quick skim.

One thing I think that would be a great test: run tests for the original C `mkcomposefs` and `composefs-info` from the C codebase using this shared library. I think we could hack that up in an integration test container.

Also a bigger and even more important test: Change the bootc revdep test to drop this shared library into the target container - then the existing ostree -> composefs logic would run through it.

crates/composefs/src/tree.rs
```diff
@@ -17,7 +17,15 @@ pub enum RegularFile<ObjectID: FsVerityHashValue> {
     /// File stored externally, referenced by fsverity hash and size.
     ///
     /// The tuple contains (fsverity hash, file size in bytes).
+    /// The fsverity digest is embedded in the overlay metacopy xattr.
     External(ObjectID, u64),
+    /// Like `External`, but without embedding the fsverity digest in the
+    /// overlay metacopy xattr.  Used by the C API when the caller set a
+    /// content-address payload but did not explicitly set a verified digest.
+    ExternalNoVerity(ObjectID, u64),
```
cgwalters, line 25: Is this coming from dumpfiles in the C test cases? I don't think it's something `mkcomposefs` would do right?

I guess in the end, if one is intending to use composefs without enforcing fsverity, it is technically OK...but OTOH, a caller can just always pass a verity digest and then just mount without the `verity=require` option for overlayfs.

Oh you know what, I bet libostree is doing this case if we don't have fsverity enabled on the object store. OK so yeah we probably do need to handle this.

crates/composefs/src/tree.rs
```diff
@@ -17,7 +17,15 @@ pub enum RegularFile<ObjectID: FsVerityHashValue> {
...
+    /// File with declared size but no content or external reference.
+    /// Produces ChunkBased layout with null chunk indices.
+    Sparse(u64),
```
cgwalters, line 28: Hmmm...this one seems just nonsensical. I can imagine the C library allows constructing it, and the C dumpfile parser probably does too, but the file would just be unreadable at runtime right? Or maybe it appears as a big chunk of zeroes?

I guess again if the C library technically allowed this, we probably need to as well...

cgwalters:
OK, I spent some tokens on crosschecking with the C impl, can you squash in https://github.com/cgwalters/composefs-rs/commit/34ee730325409074e5c539b3ee6634d2ce198194 - I think that's all correct.

crates/composefs/src/tree.rs
alexlarsson: Yes, i think its just ends up being read as zeros. I'm not sure why we would want to use this though, I don't see any use for it. Maybe just an accident of the API?

alexlarsson: But, yeah, if the C api allows it, and its not too much work to support...

crates/composefs-oci/src/delta.rs
```diff
@@ -130,14 +130,20 @@ impl<ObjectID: FsVerityHashValue> ComposeFsDataSource<ObjectID> {
...
                     .repo
                     .open_object(id)
                     .with_context(|| format!("Opening source object for {}", path.display()))?;
                 CurrentFile::External(File::from(fd))
             }
+            RegularFile::Sparse(_) => {
+                anyhow::bail!(
+                    "Sparse file not supported as delta source: {}",
+                    path.display()
+                );
```
alexlarsson, line 145: I mean, we could easily create a sparse tmpfile fd of the right lenght with just a truncate()  call here.

crates/composefs/src/erofs/format.rs
```diff
@@ -263,7 +263,7 @@ impl std::ops::BitOr<u32> for FileType {
     type Output = ModeField;
 
     fn bitor(self, permissions: u32) -> ModeField {
-        ModeField(self | (permissions as u16))
+        ModeField(self | (permissions as u16 & !S_IFMT))
```
alexlarsson, line 267: I think this is likely right. If we want to or some new permissions we don't want to inherit some new mode. But, maybe a comment about this?

crates/composefs/src/erofs/reader.rs
```diff
@@ -1837,8 +1889,19 @@ fn populate_directory<ObjectID: FsVerityHashValue>(
...
+                        tree::LeafContent::Regular(tree::RegularFile::ExternalNoVerity(
+                            id,
+                            child_inode.size(),
+                        ))
+                    } else if child_inode.data_layout()? == DataLayout::ChunkBased {
+                        tree::LeafContent::Regular(tree::RegularFile::Sparse(child_inode.size()))
                     } else {
-                        if img.composefs_restricted {
+                        if img.composefs_restricted
+                            && img.header.composefs_version == COMPOSEFS_VERSION
```
alexlarsson: I see the C code doesn't enforce this limit, so this is correct, but can you add a comment to the MAX_INLINE_CONTENT that it only applies to validating v2 images.

crates/composefs/src/erofs/reader.rs
```diff
@@ -1854,16 +1917,19 @@ fn populate_directory<ObjectID: FsVerityHashValue>(
                     }
                 }
                 S_IFLNK => {
-                    let target_data = child_inode.inline().unwrap_or(&[]);
-                    if target_data.len() > crate::SYMLINK_MAX {
+                    let target_data = extract_all_file_data(img, &child_inode)?;
+                    if img.composefs_restricted
+                        && img.header.composefs_version == COMPOSEFS_VERSION
+                        && target_data.len() > crate::SYMLINK_MAX
```
alexlarsson: Same here, SYMLINK_MAX_V2

crates/composefs/src/erofs/writer.rs
```diff
@@ -644,50 +651,70 @@ impl<'a> Directory<'a> {
 /// 2. Clamp to at least BLOCK_BITS (12)
 /// 3. Clamp to at most BLOCK_BITS + 31 (max representable)
 /// 4. Return chunkbits - BLOCK_BITS
-fn compute_chunk_format(file_size: u64) -> u32 {
+fn compute_chunk_bitsize(file_size: u64) -> u32 {
     const BLOCK_BITS: u32 = format::BLOCK_BITS as u32;
     const CHUNK_FORMAT_BLKBITS_MASK: u32 = 0x001F; // 31
 
-    // Compute the chunkbits to use for the file size.
-    // We want as few chunks as possible, but not an unnecessarily large chunk.
```
alexlarsson, line 652: Bunch of comments here are unnecessarily removed (which claude seems to do all the time...)

crates/composefs/src/erofs/writer.rs
```diff
@@ -1806,6 +2004,69 @@ type PreparedInodes<'a, ObjectID> = (Vec<Inode<'a, ObjectID>>, Vec<XAttr>, (u64,
 
 /// Shared setup for all `mkfs_erofs_*` entry points.
 ///
+/// Epoch1: promote symlink inodes to data-block layout when the inode header +
+/// xattrs + target would fill a full block. Must run after share_xattrs (so
+/// xattr sizes are final) and after calculate_min_mtime (so compact/extended
+/// is deterministic).
+fn fixup_epoch1_data_blocks<ObjectID: FsVerityHashValue>(
```
alexlarsson, line 2022: There is so many details here with the padding, block/inline splitting, etc that I fear that a regular review just cannot be trusted for this. We already have test_mkcomposefs_vs_c_mkcomposefs which compares that a basic image gets the same thing, I think we need to extend this to a more fuzz-like thing that generates a bunch of random images with different file sizes, etc, to try to shake out any issues in this.

crates/composefs-capi/include/libcomposefs/private/erofs_fs.h
```diff
@@ -0,0 +1,461 @@
+/* SPDX-License-Identifier: GPL-2.0-only OR Apache-2.0 */
```
alexlarsson, line 1: I wonder if we should put the non-public headers in a separate directory?

crates/composefs-capi/install.sh
```diff
@@ -0,0 +1,46 @@
...
+ln -sf "libcomposefs.so.${SOVERSION}" "${LIBDIR}/libcomposefs.so"
+
+# Static library
+if [ -f "$AFILE" ]; then
+    install -m 644 "$AFILE" "${LIBDIR}/libcomposefs.a"
+fi
+
+# Headers
+SCRIPTDIR="$(cd "$(dirname "$0")" && pwd)"
+for h in "${SCRIPTDIR}"/include/libcomposefs/*.h; do
```
alexlarsson, line 37: Not all the headers we have locally should be public, they are not in the C version.

crates/composefs-capi/install.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+set -eu
+
+PREFIX="${1:-/usr/local}"
+LIBDIR="${PREFIX}/lib"
```
alexlarsson: Shouldn't we allow overrides for this too for lib64, etc?

crates/composefs-capi/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+
+const LCFS_DIGEST_SIZE: usize = 32;
+
+pub(crate) struct FfiXattr {
+    key: CString,
+    value: Vec<u8>,
+}
+
+pub(crate) struct FfiNode {
+    ref_count: i32,
```
alexlarsson: Would it not be possible to use an actual Rc<FfiNode> for this? Maybe that just overcomplicates things internally.

alexlarsson: Yeah, it probably just makes it more likely that we'll get some behavioral differences.

alexlarsson:
I didn't do a super detailed review, but left some comments. Overall I think this is a viable approach, although i have some overall ideas.

First of all, (and i mentioned this in a comment), we really need to machine-verify that these block allocation and inline behavior produces the same result in all cases. We should have some code to generate random dump files with a wide variety of file sizes and stuff, and then feed it into both implementation and validate we get the exact same results.

Secondly, the primary C consumer of the library is ostree, we need to test it with the new version, and specifically we need to ensure it produces the same bitwise identical composefs images for a given ostree commit.

Also, I'm not sure some of the rust code in libcomposefs-capi is actually cleaner than the C code. In particular the code around managing lcfs_node_s trees. I wonder if for some of this code it wouldn't be better to just take the C code directly, and then just have the "meat" in pure rust. I dunno...

crates/composefs/src/erofs/writer.rs
cgwalters: There's actually several of those tests - and I think `test_v1_binary_identical_to_c_mkcomposefs` is already covering this.

proptests aren't quite the same thing as fuzzing, but IME get us 90% of the way there cheaply. The challenge is to really do intelligent coverage-guided fuzzing we need it all in the same process so we'd need a harness that calls into the *original* C library through Rust bindings or so.

What is likely missing though here is to extend `LeafContentSpec` to cover both the new cases (external-no-verity and sparse-zeroes).

crates/composefs-ctl/src/mountcomposefs.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+    result
+}
+
+struct ParsedOption {
+    key: String,
+    value: Option<String>,
+}
+
+fn parse_mount_options(options: &str) -> Vec<ParsedOption> {
```
cgwalters, line 59: We could use clap right?

crates/composefs-ioctls/src/mount.rs
```diff
@@ -0,0 +1,58 @@
...
+    userns_fd: u64,
+}
+
+unsafe extern "C" {
+    fn syscall(num: std::ffi::c_long, ...) -> std::ffi::c_long;
+}
+
+/// Applies an ID mapping from a user namespace to a mount.
+pub fn mount_setattr_idmap(
+    mount_fd: BorrowedFd<'_>,
```
cgwalters: Fine as is, but a bit more typical to use `impl AsFd` for these

giuseppe:
@alexlarsson @cgwalters I've added `mount.composefs` and man pages to this PR, so this should be everything needed to replace `composefs` with this repo

crates/composefs/src/mount.rs
```diff
@@ -170,24 +190,38 @@ impl MountOptions {
...
     fsconfig_set_string(overlayfs.as_fd(), "redirect_dir", "on")?;
-    if enable_verity {
-        fsconfig_set_string(overlayfs.as_fd(), "verity", "require")?;
+    match verity {
+        VerityRequirement::Disabled => {}
+        VerityRequirement::Required => {
+            fsconfig_set_string(overlayfs.as_fd(), "verity", "require")?;
+        }
+        VerityRequirement::Try => {
+            match fsconfig_set_string(overlayfs.as_fd(), "verity", "require") {
```
cgwalters, line 213: I wonder though...was this ever really working? AFAICS, this is basically about just the raw overlayfs verity=require support, and I think at this point most systems that care about composefs should have that.

Those that don't really just should be specifying enable_verity: false.

Actually something notable - since this fsmount code was written, at the repository level we now know directly from `meta.json` whether fsverity is enabled or not. I think we could probably wire this up better in the future - if the repository's verity is disabled, then we should default to disabling it for mounts too. Well, except for the sealed UKI case, where source of truth has to be the `composefs=` karg.

crates/composefs-oci/src/delta.rs
cgwalters: Yeah; OTOH no one should be using this anyways

cgwalters:
> @alexlarsson @cgwalters I've added mount.composefs and man pages to this PR, so this should be everything needed to replace composefs with this repo

Yeah, though I think we can strengthen testing here by copying in `composefs.spec` and then changing how the `bootc/Justfile` works to *also* support injecting the built RPM, entirely replacing the C implementation. That would give us better coverage here.

giuseppe:
> > @alexlarsson @cgwalters I've added mount.composefs and man pages to this PR, so this should be everything needed to replace composefs with this repo
> 
> Yeah, though I think we can strengthen testing here by copying in `composefs.spec` and then changing how the `bootc/Justfile` works to _also_ support injecting the built RPM, entirely replacing the C implementation. That would give us better coverage here.

do you prefer it as part of this PR or as a follow-up?

crates/composefs-ctl/src/mountcomposefs.rs
```diff
@@ -0,0 +1,272 @@
+//! mount.composefs - Mount helper for composefs images.
+//!
+//! This is a Rust reimplementation of the C mount.composefs tool, providing
```
cgwalters, line 3: I think ideally in the future we bridge this to the varlink API, and we also should extend it to support `--repo`.

cgwalters (approved):
From my PoV I think since (like composefs-ostree) this is basically mostly just a new opt-in crate for the C library, it should be good to merge and continue to iterate from there!

PR #306, opened 2026-06-03
Some work on FUSE

cgwalters:
See commits for details.

PR #328, opened 2026-06-24
Support fuse with overlayfs

alexlarsson:
This adds support for a fuse version that only serves the erofs metadata, and helpers to mount an overlayfs using this with userxattrs. This allows rootless mounting of a composefs image where file content (but not metadata) is directly accessed from the regular fs by the kernel and should perform similar to the rootful overlayfs mount.

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -23,6 +23,8 @@ pub use composefs_http;
 pub use composefs_oci;
 
 pub mod composefs_info;
+#[cfg(feature = "fuse")]
+pub mod fuse;
 pub mod mkcomposefs;
 pub mod mountcomposefs;
 /// Varlink RPC service exposing repository operations over a Unix socket.
@@ -441,15 +443,8 @@ enum OciCommand {
         /// Mount the bootable variant instead of the regular EROFS image
         #[arg(long)]
         bootable: bool,
-        /// Writable upper layer directory for overlayfs
-        #[arg(long, requires = "workdir")]
-        upperdir: Option<PathBuf>,
-        /// Work directory for overlayfs (required with --upperdir)
-        #[arg(long, requires = "upperdir")]
-        workdir: Option<PathBuf>,
-        /// Mount read-write (requires --upperdir)
-        #[arg(long, requires = "upperdir")]
-        read_write: bool,
+        #[clap(flatten)]
+        mount_opts: MountOpts,
     },
     /// Compute the composefs image ID of a stored OCI image's rootfs
     ///
@@ -525,15 +520,8 @@ enum OstreeCommand {
         commit: String,
         /// Target mountpoint
         mountpoint: String,
-        /// Writable upper layer directory for overlayfs
-        #[arg(long, requires = "workdir")]
-        upperdir: Option<PathBuf>,
-        /// Work directory for overlayfs (required with --upperdir)
-        #[arg(long, requires = "upperdir")]
-        workdir: Option<PathBuf>,
-        /// Mount read-write (requires --upperdir)
-        #[arg(long, requires = "upperdir")]
-        read_write: bool,
+        #[clap(flatten)]
+        mount_opts: MountOpts,
     },
     /// Dump the filesystem of an ostree commit as a composefs dumpfile to stdout
     Dump {
@@ -583,6 +571,64 @@ struct FsReadOptions {
     /// Don't copy /usr metadata to root directory (use if root already has well-defined metadata)
     #[clap(long)]
     no_propagate_usr_to_root: bool,
+}
+
+/// Common options for mount commands (shared across regular, OCI, and ostree mount).
+#[derive(Debug, Parser)]
+struct MountOpts {
+    /// Mount mode: auto, yes (force FUSE), or no (force kernel)
+    #[cfg(feature = "fuse")]
+    #[arg(long, value_enum, default_value_t)]
+    fuse: FuseMode,
+    /// Run FUSE server in the foreground (don't daemonize)
+    #[cfg(feature = "fuse")]
+    #[arg(long)]
+    foreground: bool,
+    /// Writable upper layer directory for overlayfs
+    #[arg(long, requires = "workdir")]
+    upperdir: Option<PathBuf>,
+    /// Work directory for overlayfs (required with --upperdir)
+    #[arg(long, requires = "upperdir")]
+    workdir: Option<PathBuf>,
+    /// Mount read-write (requires --upperdir)
+    #[arg(long, requires = "upperdir")]
+    read_write: bool,
+}
+
+impl MountOpts {
+    fn to_mount_options(&self) -> Result<composefs::mount::MountOptions> {
+        get_mount_options(
+            self.upperdir.as_deref(),
+            self.workdir.as_deref(),
+            self.read_write,
+        )
+    }
+
+    fn mount_image<ObjectID: FsVerityHashValue>(
+        &self,
+        repo: &Arc<Repository<ObjectID>>,
+        image_name: &str,
+        mountpoint: &str,
+    ) -> Result<()> {
+        let mount_options = self.to_mount_options()?;
+
+        #[cfg(feature = "fuse")]
+        if let mode @ (MountMode::Fuse | MountMode::FuseOverlay) =
+            detect_mount_mode(self.fuse, self.upperdir.is_some())
+        {
+            return run_fuse_mount(
+                repo,
+                image_name,
+                mountpoint,
+                mode,
+                mount_options,
+                self.foreground,
+            );
+        }
+
+        repo.mount_at(image_name, mountpoint, &mount_options)?;
+        Ok(())
+    }
 }
 
 #[derive(Debug, Subcommand)]
@@ -652,15 +698,8 @@ enum Command {
         name: String,
         /// the mountpoint
         mountpoint: String,
-        /// Writable upper layer directory for overlayfs
-        #[arg(long, requires = "workdir")]
-        upperdir: Option<PathBuf>,
-        /// Work directory for overlayfs (required with --upperdir)
-        #[arg(long, requires = "upperdir")]
-        workdir: Option<PathBuf>,
-        /// Mount read-write (requires --upperdir)
-        #[arg(long, requires = "upperdir")]
-        read_write: bool,
+        #[clap(flatten)]
+        mount_opts: MountOpts,
     },
     /// Read rootfs located at a path, add all files to the repo, then create the composefs image of the rootfs,
     /// commit it to the repo, and print its image object ID
@@ -809,6 +848,9 @@ fn get_mount_options(
     options.set_read_write(read_write);
     Ok(options)
 }
+
+#[cfg(feature = "fuse")]
+use fuse::{FuseMode, MountMode, detect_mount_mode, run_fuse_mount};
 
 #[cfg(feature = "oci")]
 pub(crate) fn verity_opt<ObjectID>(opt: &Option<String>) -> Result<Option<ObjectID>>
@@ -1388,12 +1430,8 @@ where
                 ref image,
                 ref mountpoint,
                 bootable,
-                ref upperdir,
-                ref workdir,
-                read_write,
+                ref mount_opts,
             } => {
-                let mount_options =
-                    get_mount_options(upperdir.as_deref(), workdir.as_deref(), read_write)?;
                 let img = if image.starts_with("sha256:") {
                     let digest: composefs_oci::OciDigest =
                         image.parse().context("Parsing manifest digest")?;
@@ -1416,7 +1454,7 @@ where
                         ),
                     }
                 };
-                repo.mount_at(&erofs_id.to_hex(), mountpoint.as_str(), &mount_options)?;
+                mount_opts.mount_image(&repo, &erofs_id.to_hex(), mountpoint.as_str())?;
             }
             OciCommand::ComputeId { config_opts } => {
                 let fs = load_filesystem_from_oci_image(&repo, config_opts)?;
@@ -1699,14 +1737,10 @@ where
             OstreeCommand::Mount {
                 ref commit,
                 ref mountpoint,
-                ref upperdir,
-                ref workdir,
-                read_write,
+                ref mount_opts,
             } => {
-                let mount_options =
-                    get_mount_options(upperdir.as_deref(), workdir.as_deref(), read_write)?;
                 let image_id = composefs_ostree::get_image_ref(&repo, commit)?;
-                repo.mount_at(&image_id.to_hex(), mountpoint.as_str(), &mount_options)?;
+                mount_opts.mount_image(&repo, &image_id.to_hex(), mountpoint.as_str())?;
             }
             OstreeCommand::Dump { ref commit_name } => {
                 let fs = composefs_ostree::create_filesystem(&repo, commit_name)?;
@@ -1766,13 +1800,9 @@ where
         Command::Mount {
             name,
             mountpoint,
-            ref upperdir,
-            ref workdir,
-            read_write,
+            ref mount_opts,
         } => {
-            let mount_options =
-                get_mount_options(upperdir.as_deref(), workdir.as_deref(), read_write)?;
-            repo.mount_at(&name, &mountpoint, &mount_options)?;
+            mount_opts.mount_image(&repo, &name, &mountpoint)?;
         }
         Command::ImageObjects { name } => {
             let objects = repo.objects_for_image(&name)?;
--- a/crates/composefs-fuse/src/lib.rs
+++ b/crates/composefs-fuse/src/lib.rs
@@ -883,6 +883,135 @@ pub fn mount_fuse(dev_fuse: impl AsFd, options: &FuseMountOptions) -> anyhow::Re
     )?)
 }
 
+/// Options controlling how an overlayfs is created on top of a FUSE mount.
+#[derive(Debug, Default)]
+#[non_exhaustive]
+pub struct OverlayMountOptions {
+    overlay_xattr: OverlayXattrMode,
+    upperdirs: Option<(OwnedFd, OwnedFd)>,
+    read_write: bool,
+    enable_verity: bool,
+}
+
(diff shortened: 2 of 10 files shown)


cgwalters:
Maybe some overlap with https://github.com/composefs/composefs-rs/pull/306 ?

alexlarsson:
@cgwalters I don't think there is necessary an overlap, except perhaps that the passthrough support is intended to solve the same performance issue (but is not useful as it is root only). They are however complementary, in that the readdir+ and multi-threading will increase metadata performance.

alexlarsson:
That said, there might be code conflicts in the two, i'll have a look at that.

crates/composefs-fuse/src/lib.rs
```diff
@@ -32,15 +34,32 @@ use rustix::{
...
+const XATTR_USER_OVERLAY_OPAQUE: &[u8] = b"user.overlay.opaque";
+const XATTR_USER_OVERLAY_WHITEOUT: &[u8] = b"user.overlay.whiteout";
+const XATTR_USER_OVERLAY_PREFIX: &[u8] = b"user.overlay.";
+const XATTR_USER_OVERLAY_ESCAPED_PREFIX: &[u8] = b"user.overlay.overlay.";
+
+fn escape_xattr_name(name: &[u8]) -> Vec<u8> {
+    if let Some(rest) = name.strip_prefix(XATTR_USER_OVERLAY_PREFIX) {
+        [XATTR_USER_OVERLAY_ESCAPED_PREFIX, rest].concat()
+    } else {
+        name.to_vec()
```
cgwalters: Could use Cow 🐮 here

crates/composefs-fuse/src/lib.rs
```diff
@@ -352,8 +437,26 @@ impl<ObjectID: FsVerityHashValue> Filesystem for TreeFuse<'_, ObjectID> {
...
+            } else if value.len() > size as usize {
+                return reply.error(Errno::RANGE.raw_os_error());
+            }
+            return reply.data(&value);
+        }
+
+        let lookup_name = if self.follow_redirects {
+            name.to_os_string()
+        } else {
+            OsStr::from_bytes(&escape_xattr_name(name_bytes)).to_os_string()
```
cgwalters: I think this is an unnecessary copy

crates/composefs-fuse/src/lib.rs
```diff
@@ -536,23 +657,43 @@ pub fn mount_fuse(dev_fuse: impl AsFd, options: &FuseMountOptions) -> anyhow::Re
...
+#[non_exhaustive]
+pub struct ServeFuseOptions {
+    follow_redirects: bool,
+}
+
+impl ServeFuseOptions {
+    /// When true, the server follows overlay redirects and serves file
+    /// content from the repository. When false, it synthesizes
+    /// `user.overlay.*` xattrs for use as an overlayfs lower layer.
+    /// Defaults to false.
```
cgwalters: But actually...I am not sure we considered this at all but - is there any reason at all to have the direct serving? I am not sure there is a strong one...I think the FUSE implementation was kind of intended mainly for unprivileged mounts, but since nowadays overlay is allowed in user namespaces, I think we should automatically set "follow_redirects" = false if running unprivileged right?

Actually to say it a different way - don't we *always* want to just always use overlayfs, and only have the FUSE for unprivileged EROFS? In that case, when running privileged, we should *not* use `user.overlay` right? Or I guess we still can, but there's no reason to?

cgwalters (commented):
OK, now that I understand what this is doing - pretty cool! This makes a ton of sense.
 

One thing I did in the other PR is add some more integration tests for the FUSE path, which would probably be quite useful here.

crates/composefs-fuse/src/lib.rs
alexlarsson: Honestly I'm not sure here. There is one way that `follow_redirects` mode is useful, in that it allows any non-root user to mount a composefs image in the "root" user namespace. 

Like, if you're in the shell you can just mount a cfs and have some other program look into the mount. You cannot do that with an overlay mount. For that you need to be in a new user namespace (where you have cap_sysadmin) so you can do the mount. And if you create a new user+mnt namespace programs in the root namespace can't see your mount. So, non-folllow_redirect is primarily useful for container like tools, wheras follow_redirect is for traditional commandline work.

So, I do think it would make sense to support both of these modes. I wonder though if the current composefs fuse implementation is ideal for this kind of use. It starts by parsing the entire erofs into a filesystem tree, which adds quite some latency. An implementation that just reads the erofs file into memory and does metadata lookups directly from that would probably be faster at startup. And, for a metadata-only implementation (non-follow-redirects) that would probably be pretty easy to implement, as all you have to do is read inode info. 

I guess that could be later work though.

giuseppe: +1 to make the metadata-only serving the default.

cgwalters: > There is one way that follow_redirects mode is useful, in that it allows any non-root user to mount a composefs image in the "root" user namespace.

Yeah, but I'm not sure of any kind of "production" use case for that. Debugging can equally well just use APIs to inspect things.

follow_redirects is a confusing name, how about `--standalone` for the non-overlayfs case?

> It starts by parsing the entire erofs into a filesystem tree, which adds quite some latency. An implementation that just reads the erofs file into memory and does metadata lookups directly from that would probably be faster at startup. 

Yeah, agreed, we should do it that way.

giuseppe: > > It starts by parsing the entire erofs into a filesystem tree, which adds quite some latency. An implementation that just reads the erofs file into memory and does metadata lookups directly from that would probably be faster at startup.
> 
> Yeah, agreed, we should do it that way.

can we mmap'it so multiple containers using the same erofs will share the memory?

alexlarsson:
Ok I did some more work on this, picking up a bunch of changes from #306 with some changes. This now has multi-threaded fuse, mount apis to support mounting direct as well as via overlayfs with and without userxattrs, as well as verity=require. It also has mount CLI options for image/oci/ostree that mostly does the right thing by default (for example, it will use a non-overlay approach if we don't have cap_sysadmin).

I didn't look at the varlink part yet. Also this is still based on serving a filesystem, rather than directly serving from an erofs image.

crates/composefs-ctl/src/lib.rs
```diff
@@ -441,6 +445,18 @@ enum OciCommand {
         /// Mount the bootable variant instead of the regular EROFS image
         #[arg(long)]
         bootable: bool,
+        /// Force FUSE mount instead of auto-detecting
+        #[cfg(feature = "fuse")]
+        #[arg(long, conflicts_with = "no_fuse")]
+        fuse: bool,
+        /// Force kernel mount instead of auto-detecting
+        #[cfg(feature = "fuse")]
+        #[arg(long, conflicts_with = "fuse")]
```
cgwalters: I think an enum is better `--fuse=auto|yes|no`. Could also have a `#[clap(flatten)]` shared struct to dedup

crates/composefs-fuse/src/lib.rs
alexlarsson: I'd think so.

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+enum MountMode {
+    Kernel,
+    Fuse,
+    FuseOverlay,
+}
+
+#[cfg(feature = "fuse")]
+fn in_init_user_namespace() -> bool {
+    std::fs::read_to_string("/proc/self/uid_map")
+        .map(|s| s.trim() == "0          0 4294967295")
```
cgwalters: This feels hacky.

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
     Ok(options)
 }
 
+#[cfg(feature = "fuse")]
```
cgwalters, line 852: I think it'd be nicer to have a separate `mod fuse` with all this stuff so we don't need a lot of `#[cfg`

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+fn has_cap_sys_admin() -> bool {
+    if let Ok(caps) = capabilities(None) {
+        caps.effective.contains(CapabilitySet::SYS_ADMIN)
+    } else {
+        false
+    }
+}
+
+#[cfg(feature = "fuse")]
+fn detect_mount_mode(force_fuse: bool, no_fuse: bool, has_upper: bool) -> MountMode {
```
cgwalters: instead of the first two bools an enum would be way clearer

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+    let use_fuse = if force_fuse {
+        true
+    } else if no_fuse {
+        false
+    } else {
+        !(getuid().is_root() && in_init_user_namespace())
```
cgwalters: Hmm do we really need to check the userns? I think we could just always check for `has_cap_sys_admin`?

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+    if let Some((upper_fd, work_fd)) = mount_options.into_overlay() {
+        cmd.arg("--upper-fd").arg(upper_fd.as_raw_fd().to_string());
+        cmd.arg("--work-fd").arg(work_fd.as_raw_fd().to_string());
+        clear_cloexec(&upper_fd);
+        clear_cloexec(&work_fd);
+        std::mem::forget(upper_fd);
+        std::mem::forget(work_fd);
+    }
+
+    clear_cloexec(&image_fd);
```
cgwalters: I think https://docs.rs/cap-std-ext/latest/cap_std_ext/cmdext/struct.CmdFds.html is better than this

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+    #[arg(long)]
+    mountpoint: String,
+    #[arg(long)]
+    image_fd: i32,
+    #[arg(long)]
+    repo_fd: i32,
+    #[arg(long)]
+    ready_fd: i32,
+    #[arg(long, value_parser = ["fuse", "fuse-overlay"])]
+    mode: String,
```
cgwalters: Clearer as an `enum` right?

crates/composefs-ctl/src/lib.rs
```diff
@@ -810,6 +850,298 @@ fn get_mount_options(
...
+}
+
+/// Entry point for the internal FUSE server process, called from main()
+/// before the tokio runtime is created.
+#[cfg(feature = "fuse")]
+#[allow(unsafe_code)]
+pub fn run_internal_fuse_serve(args: InternalFuseServeArgs) -> Result<()> {
+    use std::os::fd::FromRawFd;
+
+    let image_fd = unsafe { std::os::fd::OwnedFd::from_raw_fd(args.image_fd) };
```
cgwalters: Perhaps we could use the socket activation protocol to add a little bit more safety?

cgwalters (changes requested):
Cool, overall looking good

crates/composefs-fuse/src/lib.rs
alexlarsson: At least if the image is fs-verity, because then we can trust that it doesn't change under our back.

crates/composefs-ctl/src/lib.rs
giuseppe: wouldn't that be enough in a rootless environment?

crates/composefs-ctl/src/lib.rs
alexlarsson: I think this is the canonical way to see if you're in the init user namespace. @giuseppe is there something better?

crates/composefs-fuse/src/lib.rs
cgwalters: Yep agreed

crates/composefs-ctl/src/lib.rs
giuseppe: we have a similar function in containers/storage:

```
// hasFullUsersMappings checks whether the current user namespace has all the IDs mapped.
func hasFullUsersMappings() (bool, error) {
        content, err := os.ReadFile("/proc/self/uid_map")
        if err != nil {
                return false, err
        }
        // The kernel rejects attempts to create mappings where either starting
        // point is (u32)-1: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/user_namespace.c?id=af3e9579ecfb#n1006 .
        // So, if the uid_map contains 4294967295, the entire IDs space is available in the
        // user namespace, so it is likely the initial user namespace.
        return bytes.Contains(content, []byte("4294967295")), nil
}
```

it will fail if you do something like: `unshare --map-users 0:0:4294967295 ...` but I guess that is not a common configuration

alexlarsson:
Ok, completely new fuse reimplementation. Much smaller and should start faster. I'd like to do some more reviewing of it next week, and handle some of the comments above too. But, it looks pretty sweet to me.

crates/composefs-fuse/src/lib.rs
```diff
@@ -208,281 +251,540 @@ enum OpenHandle {
...
+}
 
-        let LeafContent::Symlink(target) = &leaf.content else {
-            return reply.error(Errno::INVAL.raw_os_error());
+impl Filesystem for ComposefsFuse {
+    fn statfs(&self, _req: &Request, _ino: INodeNo, reply: fuser::ReplyStatfs) {
+        reply.statfs(0, 0, 0, 0, 0, 4096, 255, 4096);
+    }
+
+    fn forget(&self, _req: &Request, _ino: INodeNo, _nlookup: u64) {}
```
giuseppe, line 438: pre-existing, but what is the memory cost of keeping all the inodes in memory?

If you use a ~big image like bootc, how much memory is consumed by the fuse file system?

crates/composefs-fuse/src/lib.rs
```diff
@@ -208,281 +251,540 @@ enum OpenHandle {
...
     }
 }
 
-/// Opens /dev/fuse.
+fn errno_to_fuser(errno: rustix::io::Errno) -> fuser::Errno {
+    fuser::Errno::from(std::io::Error::from_raw_os_error(errno.raw_os_error()))
+}
+
+/// Check if an fd has fs-verity enabled, meaning its contents cannot change.
+fn is_safe_to_mmap(fd: &impl AsFd) -> bool {
```
cgwalters, line 802: We could argue to move this check into the memmap2 crate, then we wouldn't need any unsafe here.

It's the same with sealed memfds

cgwalters (dismissed):
Overall looks good to me; you didn't address a few of my comments, was that intentional?

crates/composefs-fuse/src/lib.rs
alexlarsson: You mean the kernel inodes? They are not kept forever, if there is memory pressure they are flushed out just like any other filesystems in-memory inodes. Also they are lazily instantiated.

As for the erofs image itself for quay.io/fedora/fedora-bootc:42, that is 24M, and with the new fuse backend we shouldn't use all that much memory over that.

alexlarsson:
> Overall looks good to me; you didn't address a few of my comments, was that intentional?

I mean to look at those comment next week, as they were mainly in a different area than this change. I just wanted people to have a look at the new approach earlier.

crates/composefs-fuse/src/lib.rs
giuseppe: no nevermind, I see there is no extra memory used for the FUSE inodes, only the information returned from the EROFS mmap'ed file.  Cool!

crates/composefs-ctl/src/lib.rs
alexlarsson: The point of this check is that if we're in the root user namespace but not root or with cap_sys_admin we can still use the fuse mount (but not via overlayfs) because we have the setuid fusermount binary to help us.

crates/composefs-fuse/src/lib.rs
alexlarsson: https://github.com/RazrFalcon/memmap2-rs/pull/173

crates/composefs-ctl/src/lib.rs
alexlarsson: done

crates/composefs-ctl/src/lib.rs
alexlarsson: done

crates/composefs-ctl/src/lib.rs
alexlarsson: done

crates/composefs-ctl/src/lib.rs
alexlarsson: done

alexlarsson:
Hmm, the integration test is failing on cs9, because we need this commit:
https://github.com/torvalds/linux/commit/5ef7bcdeecc982ae17d13b682a85123c7d74b200

This is in upstream 6.16, and backported to CS10, but not in CS9. I think we need to support the non-overlayfs codepath as well, for backwards compat.

cgwalters:
Ah. Hmm...there's probably no reasonable way for us to detect precisely when a kernel has that support right? We can just try to mount and see if we get `EINVAL`?

I guess our options are:

- Just fail and let the invoking process retry as a pure FUSE mount
- Have `--fuse=overlay-if-possible` where we just fall back to fuse on any EINVAL

?

For now our integration test could then special case c9s and pass `overlay-if-possible`

alexlarsson:
@cgwalters I added some code that detects when its possible ahead of time. Its not free (but not tooo bad), so i also added a pre-6.16 option so you can avoid it.

I first tried a "try overlay and then fallback", but its just way to icky, as it needs a full fuse server daemon process set up, etc.

cgwalters (approved):

alexlarsson:
@cgwalters I wonder if there is some more generic way to handle pre-6.15 and pre-6.16 so that enabling one implies all lower versions.

PR #329, opened 2026-06-24
selabel: Use hybrid DFA+PCRE2 fallback for regex matching

cgwalters:
Originally (before 835c9865c) we were compiling all the regexps into a single DFA, which allowed huge amounts of optimization.

In that commit we added support for lookbehinds, but this is dramatically (easily 5000x) slower to evaluate each regexp individually.

Since the default Fedora SELinux policy doesn't have any lookbehinds, let's first try the previous strategy of building a single DFA with all regexps at once.

If that fails, then our fallback strategy is to parse each one, and ones we can parse with the main regex_syntax crate we put into the single DFA, the others go into individual PCRE regexps.

This gives us nearly all the performance back.

Assisted-by: OpenCode (claude-opus-4-6)


giuseppe (approved):
LGTM

PR #330, opened 2026-06-26
erofs: Switch default format version from V2 to V1

### LINKED ISSUE #327: Switch over to v1 format by default
As of [the 0.7 release](https://github.com/composefs/composefs-rs/releases/tag/v0.7.0) we support generating an EROFS v1 format. Let's make that our new default.

cgwalters:
This is one of the goals in getting to composefs-rs 1.0.

V1 is compatible with both C mkcomposefs 1.0.8 and composefs-rs, making it the better default for interoperability. V2 remains available for users who need it.

For backward compatibility, old repositories that lack an explicit erofs_formats field in meta.json continue to use V2 via a separate serde default (legacy_erofs_default), ensuring they are not silently reinterpreted as V1.

Tests that validate specific V2 behavior (directory entry assertions without whiteout stubs, V2-only OCI image refs, digest stability pinning) now use V2 explicitly.

Closes: https://github.com/composefs/composefs-rs/issues/327
Assisted-by: opencode (Claude claude-opus-4-6)

### DIFF
--- a/crates/composefs-integration-tests/src/tests/digest_stability.rs
+++ b/crates/composefs-integration-tests/src/tests/digest_stability.rs
@@ -21,9 +21,9 @@ struct ContainerImage {
     /// unavailable (e.g. a PR that adds a new mirror entry before it has been
     /// pushed).  Should be pinned by digest for reproducibility.
     upstream_ref: &'static str,
-    /// Expected composefs image ID without `--bootable` (V2/default EROFS).
+    /// Expected composefs image ID without `--bootable` (V2 EROFS).
     expected_id: &'static str,
-    /// Expected composefs image ID with `--bootable` (V2/default EROFS), or
+    /// Expected composefs image ID with `--bootable` (V2 EROFS), or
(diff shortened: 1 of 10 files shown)


crates/composefs-integration-tests/src/tests/digest_stability.rs
```diff
@@ -386,109 +389,117 @@ fn check_digest_equivalence(image: &ContainerImage) -> Result<()> {
...
-        eprintln!("  MISMATCH detected — capturing dumpfiles for diff...");
+        let repo_dir = tempfile::tempdir()?;
+        let repo = repo_dir.path();
+        cmd!(
+            sh,
+            "{cfsctl} --insecure --repo {repo} init --erofs-version {version}"
+        )
+        .read()?;
 
-        // cstor dumpfile: `cfsctl oci dump --bootable @<config>`
```
alexlarsson, line 434: There are some random deleted comments here.

alexlarsson (approved):
Some imho unnecessary comment deletions, but otherwise lgtm

PR #331, opened 2026-06-26
Make rustdoc hold most documentation

cgwalters:
I want composefs to have a website with useful docs. After some reflection, what I think is the least effort and most clear right now is just to move things into the Rust documentation.

So things like the repository format, splitstream definition now live there. The pattern is that we have `#[cfg(doc)] mod foo;` for things that only need to live in docs.

Add information about varlink there too.

Assisted-by: OpenCode (claude-opus-4-6)


crates/composefs/src/splitstream_format.rs
```diff
@@ -0,0 +1,164 @@
...
+//! `U64` is an unsigned 64bit little-endian integer.
+//!
+//! ### File ranges ("sections")
+//!
+//! The file format consists of a fixed-sized header at the start of the file plus
+//! a number of sections located at arbitrary locations inside of the file.  All of
+//! these sections are referred to by a 64-bit `[start..end)` range expressed in
+//! terms of overall byte offsets within the complete file.
+//!
+//! ```text
```
giuseppe, line 59: should this be `text` instead of `rust`?

crates/composefs-boot/src/design.rs
```diff
@@ -0,0 +1,90 @@
...
+//! `composefs.digest=?v1-sha256-12:<digest>` or `composefs=?<digest>`) makes
+//! fs-verity verification optional. The system will boot even when the underlying
+//! filesystem does not support fs-verity or the image has no verity metadata
+//! attached. This mode exists for development and testing only; it must not be used
+//! in production.
+//!
+//! ## On-disk layout
+//!
+//! The composefs repository must be present at `/sysroot/composefs` with the
+//! standard layout (see the `composefs` crate's `repository` module).
```
giuseppe: Claude suggests to use an intra-doc link like [composefs::repository_format] here

cgwalters:
Ah, the random proptest run in this PR caught a bug, https://github.com/composefs/composefs-rs/pull/332

crates/composefs/src/splitstream_format.rs
cgwalters: It's more pseudocode, so I think `text` is OK.

giuseppe (approved):
LGTM

PR #332, opened 2026-06-26
erofs: Fix proptest check for compatibility with whiteouts

cgwalters:
test_v1_binary_identical_to_c_mkcomposefs hardcoded FormatVersion::V0 when calling the Rust writer, but C mkcomposefs auto-bumps from version 0 to 1 when it encounters a whiteout (chardev rdev=0) in the input tree. When the random generator produced a CharacterDevice(0), the composefs_version header byte differed, causing a spurious failure.

The companion test test_v1_binary_identical_unusual_content already handled this correctly. Extract the shared comparison logic into assert_binary_identical_to_c() so both tests use it.

Assisted-by: opencode (claude-opus-4-6)

### DIFF
--- a/crates/composefs/src/erofs/reader.rs
+++ b/crates/composefs/src/erofs/reader.rs
@@ -3390,20 +3390,55 @@ mod tests {
             bytes
         }
 
-        /// Verify that our Rust V1 writer produces byte-for-byte identical EROFS images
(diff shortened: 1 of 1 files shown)


giuseppe (dismissed):
LGTM

alexlarsson (approved):

PR #336, opened 2026-06-30
packaging: Add RPM spec file

giuseppe:
More details in each commit.

Manually tested on Fedora 45 Rawhide.


cgwalters (approved):
Awesome!

Next step: Wire up packit

PR #338, opened 2026-06-30
More work on ostree

alexlarsson:
This adds some fixes and cleanups to the ostree support, adds a list-refs command, and then adds support for ostree summaries and static deltas.

Also, there is a fix to FsVerityHasher that may be important to non-ostree users as well.

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -558,6 +558,14 @@ enum OstreeCommand {
     /// List all ostree commits in the repository
     #[clap(name = "images")]
     ListCommits,
+    /// List refs available in a remote ostree repository
+    ListRefs {
+        /// URL of the remote ostree repository
+        ostree_repo_url: String,
+        /// Summary index subset key (defaults to system architecture)
+        #[clap(long)]
+        subset: Option<String>,
+    },
 }
 
 /// Common options for reading a filesystem from a path
@@ -1777,6 +1785,27 @@ where
                     table.set_header(["NAME", "COMMIT"]);
                     for c in commits {
                         table.add_row([c.name.as_str(), &c.commit_id]);
+                    }
+                    println!("{table}");
+                }
+            }
+            OstreeCommand::ListRefs {
+                ref ostree_repo_url,
+                ref subset,
+            } => {
+                let mut ostree_repo = composefs_ostree::RemoteRepo::new(&repo, ostree_repo_url)?;
+                if let Some(s) = subset {
+                    ostree_repo = ostree_repo.with_summary_subset(s);
+                }
+                let refs = ostree_repo.list_remote_refs().await?;
+                if refs.is_empty() {
+                    println!("No refs found");
+                } else {
+                    let mut table = Table::new();
+                    table.load_preset(UTF8_FULL);
+                    table.set_header(["REF", "COMMIT"]);
+                    for (name, checksum) in &refs {
+                        table.add_row([name.as_str(), &hex::encode(checksum)]);
                     }
                     println!("{table}");
                 }
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -47,11 +47,11 @@ use anyhow::{Context as _, Result};
 use clap::{Parser, Subcommand, ValueEnum};
 #[cfg(any(feature = "oci", feature = "ostree"))]
 use comfy_table::{Table, presets::UTF8_FULL};
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use rustix::fs::{CWD, Mode, OFlags};
 
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 use composefs::progress::{
     ComponentId, ProgressEvent, ProgressReporter, ProgressUnit, SharedReporter,
 };
@@ -81,13 +81,13 @@ use composefs::{
 /// Renders per-component progress bars via [`MultiProgress`].  When a component
 /// completes or is skipped the bar is removed; human-readable messages are
 /// printed above the bar group via [`MultiProgress::println`].
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 struct IndicatifReporter {
     multi: MultiProgress,
     bars: Mutex<HashMap<ComponentId, ProgressBar>>,
 }
 
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 impl IndicatifReporter {
     fn new() -> Self {
         IndicatifReporter {
@@ -102,14 +102,14 @@ impl IndicatifReporter {
     }
 }
 
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 impl std::fmt::Debug for IndicatifReporter {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("IndicatifReporter").finish_non_exhaustive()
     }
 }
 
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
 impl ProgressReporter for IndicatifReporter {
     fn report(&self, event: ProgressEvent) {
         match event {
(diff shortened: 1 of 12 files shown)


crates/composefs/src/fsverity/digest.rs
```diff
@@ -80,13 +81,32 @@ impl<H: FsVerityHashValue, const LG_BLKSZ: u8> FsVerityHasher<H, LG_BLKSZ> {
...
     }
 
-    /// Add a block of data to the hasher.
+    /// Write arbitrary-sized data into the hasher.
     ///
-    /// For correct results, data should be provided in block-sized chunks (4KB)
-    /// except for the final chunk which may be smaller.
+    /// Internally buffers data and feeds complete blocks to the Merkle
+    /// tree. Any remaining partial block is flushed by [`digest()`].
+    pub fn write(&mut self, mut data: &[u8]) {
```
cgwalters: may as well `impl Write` then?

crates/composefs/src/fsverity/digest.rs
```diff
@@ -80,13 +81,32 @@ impl<H: FsVerityHashValue, const LG_BLKSZ: u8> FsVerityHasher<H, LG_BLKSZ> {
             layers: vec![],
             value: None,
             n_bytes: 0,
+            partial: Vec::new(),
```
cgwalters: I suspect https://docs.rs/bytes/latest/bytes/struct.BytesMut.html would make this code cleaner

crates/composefs/src/fsverity/digest.rs
```diff
@@ -80,13 +81,32 @@ impl<H: FsVerityHashValue, const LG_BLKSZ: u8> FsVerityHasher<H, LG_BLKSZ> {
...
+                let block = std::mem::take(&mut self.partial);
+                self.add_block(&block);
+            }
+        }
+    }
+
+    /// Add a complete block of data to the hasher.
+    ///
+    /// The data must be exactly `BLOCK_SIZE` bytes, except for the
+    /// final block which may be smaller. Prefer [`write()`] which
```
cgwalters: I think it'd be cleaner if we had a dedicated API to provide the final block that was partial. Then this could be `fn add_block(&mut self, data: &[u8; BLOCK_SIZE])` or so

crates/composefs/src/fsverity/digest.rs
cgwalters: Also BTW I think it would help to have proptest-based unit tests that write random sizes and verify the digest matches, I did that type of thing in https://github.com/composefs/tar-core

crates/composefs/src/fsverity/digest.rs
alexlarsson: This is not possible because BLOCK_SIZE is a geneirc. I added an assert at least to catch this in the future.

crates/composefs/src/fsverity/digest.rs
alexlarsson: done

crates/composefs/src/fsverity/digest.rs
alexlarsson: Not sure how you meant it to be used.

cgwalters: A `BytesMut` is a special cased more powerful version of `Vec<u8>` basically that is designed specifically for "buffer" use cases, but anyways not blocking

crates/composefs-ctl/src/lib.rs
```diff
@@ -47,11 +47,11 @@ use anyhow::{Context as _, Result};
 use clap::{Parser, Subcommand, ValueEnum};
 #[cfg(any(feature = "oci", feature = "ostree"))]
 use comfy_table::{Table, presets::UTF8_FULL};
-#[cfg(any(feature = "oci", feature = "http"))]
+#[cfg(any(feature = "oci", feature = "http", feature = "ostree"))]
```
cgwalters, line 50: I think we can consolidate to a single `#[cfg(content_fetching)]` or so

crates/composefs-ostree/src/repo.rs
```diff
@@ -111,12 +113,203 @@ pub trait OstreeRepo<ObjectID: FsVerityHashValue>: Send + Sync {
     ) -> impl Future<Output = Result<(AlignedBuf, Option<ObjectID>)>> + Send;
 }
 
+const OSTREE_SUMMARY_CONTENT_TYPE: u64 = u64::from_le_bytes(*b"osummary");
```
cgwalters, line 116: This looks...weird

crates/composefs-ostree/src/repo.rs
```diff
@@ -111,12 +113,203 @@ pub trait OstreeRepo<ObjectID: FsVerityHashValue>: Send + Sync {
...
+        ),
+        checksum: info.checksum,
+    };
+
+    let mut ss = repo.create_stream(OSTREE_SUMMARY_CONTENT_TYPE)?;
+    ss.write_inline(header.as_bytes());
+    ss.write_inline(etag_bytes);
+    ss.write_inline(lm_bytes);
+    ss.write_inline(summary_data);
+    repo.write_stream(ss, stream_id, None)?;
```
cgwalters, line 229: Nonblocking but just thinking: I'd like to make it feel nice again to store externally defined content without touching composefs core. I could imagine for example this use case is basically "store summary as OCI artifact" for example.

(Of course we could make that even more formal outside of this project)

It is basically arguing for composefs-ostree to depend on composefs-oci in a way, or at least parts of it.

crates/composefs-ostree/src/delta.rs
```diff
@@ -0,0 +1,1193 @@
...
+const OP_UNSET_READ_SOURCE: u8 = b'R';
+const OP_CLOSE: u8 = b'c';
+const OP_BSPATCH: u8 = b'B';
+
+// Signed delta magic: "OSTSGNDT" in big-endian u64
+const SIGNED_DELTA_MAGIC: u64 = 0x4F53_5453_474E_4454;
+
+// ---------- varint codec ----------
+
+fn read_varuint64(data: &[u8], pos: &mut usize) -> Result<u64> {
```
cgwalters, line 45: Stuff like this cries out to be unit tested.
(Actually I bet we might be able to do some formal verification even)

crates/composefs-ostree/src/delta.rs
```diff
@@ -0,0 +1,1193 @@
...
+    match osz {
+        1 => data[start] as u64,
+        2 => u16::from_le_bytes(data[start..start + 2].try_into().unwrap()) as u64,
+        4 => u32::from_le_bytes(data[start..start + 4].try_into().unwrap()) as u64,
+        8 => u64::from_le_bytes(data[start..start + 8].try_into().unwrap()),
+        _ => 0,
+    }
+}
+
+fn align_up(offset: u64, alignment: u64) -> u64 {
```
cgwalters, line 84: https://doc.rust-lang.org/stable/std/primitive.usize.html#method.next_multiple_of

crates/composefs-ostree/src/delta.rs
```diff
@@ -0,0 +1,1193 @@
...
+    // child 1: fixed, 8 bytes
+    // child 2: end = fo[1]
+    // child 3: end = fo[2]
+    // child 4: end = fo[3]
+    // child 5: end = fo[4]
+    // child 6: end = fo[5]
+    // child 7: end = total_size - n_frame_offsets * osz (LAST)
+
+    let osz = gv_offset_size(total_size) as u64;
+    let data_end = total_size - N_FRAME_OFFSETS as u64 * osz;
```
cgwalters, line 124: Should do checked arithmetic in general

crates/composefs-ostree/Cargo.toml
```diff
@@ -12,6 +12,7 @@ version.workspace = true
 
 [dependencies]
 anyhow = { version = "1.0.87", default-features = false }
+bsdiff = { version = "0.2.1", default-features = false }
```
cgwalters, line 15: Have you looked that this is equivalent to the C code vendored in libostree? I suspect There Be 🐉

crates/composefs-ostree/src/ostree.rs
```diff
@@ -63,6 +63,32 @@ pub enum ObjectType {
 }
 
 impl ObjectType {
+    /// Decode from the numeric value used in ostree's on-disk format.
+    pub fn from_byte(b: u8) -> Result<Self> {
+        match b {
+            1 => Ok(ObjectType::File),
+            2 => Ok(ObjectType::DirTree),
```
cgwalters, line 70: Minor but can reduce the `Ok` here if the `Err` case is `return Err`

crates/composefs-ostree/src/delta.rs
```diff
@@ -0,0 +1,1193 @@
...
+            )
+        })
+    }
+}
+
+/// Execute a single delta part, producing objects into the writer.
+///
+/// `base` is needed to resolve `SET_READ_SOURCE` opcodes that reference
+/// objects from the previous commit.
+pub(crate) fn execute_delta_part<ObjectID: FsVerityHashValue>(
```
cgwalters, line 693: Likely worth wiring up to the fuzzing infra we have

cgwalters (approved):

alexlarsson:
I'll merge this and do later followups separately.

crates/composefs-ostree/Cargo.toml
alexlarsson: I would think that if bsdiff apply produced a different result thatn the C code we would detect this as it would produce a different checksum for the new ostree object. However, I see we're not actually validating the produced checksums in the delta apply case, will fix that.

crates/composefs-ostree/src/repo.rs
alexlarsson: In this case (ostree) we need the data file (the summary) in the repo for internal use, but is there really any particular reason for a user facing store arbitrary single file blobs in the composefs repo? You're unlikely to put it into a composefs image or otherwise get dedup from this?

alexlarsson: Like, I imagine storing random directories of content in the repo is useful, like builds or what have you, but single file artifacts? What is the point?

alexlarsson:
I resolved all comments fixed in https://github.com/composefs/composefs-rs/pull/340

crates/composefs-ostree/Cargo.toml
alexlarsson: (added the checksum check)

PR #341, opened 2026-07-01
packaging: Add Packit support with vendor tarball workflow

giuseppe:


contrib/packaging/pack.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
+# Generate source + vendor tarballs and a rewritten spec file for RPM builds.
+# Used by Packit's create-archive action (see .packit.yaml).
```
cgwalters, line 3: While it's more complicated, we could mirror the flow used in https://github.com/bootc-dev/bootc - we definitely want the vendor tarball attached to releases for example.

cgwalters (dismissed):

contrib/packaging/pack.sh
giuseppe: what do you think of the new commit I've added?

cgwalters (approved):

PR #343, opened 2026-07-02
Add support for creating ostree commits

alexlarsson:
This lets us create a local ostree commit in the composefs repo based on an Filesystem (which could come from an image, which in turn could come from an oci image). 

This lets us integrate composefs and ostree easily, as mentioned in https://github.com/composefs/composefs-rs/issues/141

### DIFF
--- a/crates/composefs-ostree/src/commit.rs
+++ b/crates/composefs-ostree/src/commit.rs
@@ -25,13 +25,32 @@ use composefs::{
 };
 
 use crate::ostree::{
-    OstreeCommit, OstreeDirMeta, OstreeDirTree, OstreeFileHeader, split_sized_variant,
+    CommitMetadata, OstreeCommit, OstreeDirMeta, OstreeDirTree, OstreeFileHeader,
+    should_inline_file, split_sized_variant,
 };
 
 const OSTREE_COMMIT_CONTENT_TYPE: u64 = 0xAFE138C18C463EF1;
 
-const S_IFMT: u32 = 0o170000;
-const S_IFLNK: u32 = 0o120000;
+fn stat_xattrs_to_vec(stat: &Stat) -> Vec<(Vec<u8>, Vec<u8>)> {
+    stat.xattrs
+        .iter()
+        .map(|(k, v)| {
+            let mut val = v.to_vec();
+            // The kernel stores security.selinux with a NUL terminator,
+            // but programmatic labeling (e.g. bootc's selabel()) may omit
+            // it.  Canonicalize to always include the NUL so that ostree
+            // checksums match what ostree fsck expects.
+            if k.as_bytes() == b"security.selinux" && !val.ends_with(&[0]) {
+                val.push(0);
+            }
+            (k.as_bytes().to_vec(), val)
+        })
+        .collect()
+}
+
+const S_IFMT: u32 = composefs::erofs::format::S_IFMT as u32;
+const S_IFDIR: u32 = composefs::erofs::format::S_IFDIR as u32;
+const S_IFLNK: u32 = composefs::erofs::format::S_IFLNK as u32;
 
 fn xattrs_to_btreemap(xattrs: &[(Vec<u8>, Vec<u8>)]) -> BTreeMap<Box<OsStr>, Box<[u8]>> {
     xattrs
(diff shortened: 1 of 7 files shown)


crates/composefs-ostree/src/commit.rs
```diff
@@ -25,13 +25,27 @@ use composefs::{
...
 const OSTREE_COMMIT_CONTENT_TYPE: u64 = 0xAFE138C18C463EF1;
 
+fn stat_xattrs_to_vec(stat: &Stat) -> Vec<(Vec<u8>, Vec<u8>)> {
+    stat.xattrs
+        .iter()
+        .map(|(k, v)| (k.as_bytes().to_vec(), v.to_vec()))
+        .collect()
+}
+
 const S_IFMT: u32 = 0o170000;
```
Johan-Liebert1: We have the same stuff in `composefs::erofs`. This is a constant so probably won't matter much though

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +187,171 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+
+    fn write_leaf(
+        repo: &Arc<Repository<ObjectID>>,
+        writer: &mut CommitWriter<ObjectID>,
+        fs: &FileSystem<ObjectID>,
+        leaf_id: LeafId,
+    ) -> Result<Sha256Digest> {
+        let leaf = fs.leaf(leaf_id);
+
+        let (symlink_target, content_owned, external_obj) = match &leaf.content {
```
Johan-Liebert1: How about making symlink_target `Option<String>`?

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +187,171 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+                RegularFile::Sparse(size) => (String::new(), Some(vec![0u8; *size as usize]), None),
+            },
+            LeafContent::BlockDevice(_)
+            | LeafContent::CharacterDevice(_)
+            | LeafContent::Fifo
+            | LeafContent::Socket => (String::new(), None, None),
+        };
+        let content_bytes = content_owned.as_deref();
+
+        let file_size = match &leaf.content {
```
Johan-Liebert1, line 289: Would be better inside `impl LeafContent` imo

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +187,171 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+            LeafContent::Regular(RegularFile::Inline(data)) => data.len() as u64,
+            LeafContent::Regular(RegularFile::External(_, size)) => *size,
+            LeafContent::Regular(RegularFile::ExternalNoVerity(_, size)) => *size,
+            LeafContent::Regular(RegularFile::Sparse(size)) => *size,
+            _ => 0,
+        };
+
+        let file_type_bits = match &leaf.content {
```
Johan-Liebert1: Could also throw this inside `impl LeafContent`

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +187,171 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+                let target = target
+                    .to_str()
+                    .ok_or_else(|| anyhow!("non-UTF8 symlink target: {:?}", target))?;
+                (target.to_string(), None, None)
+            }
+            LeafContent::Regular(reg) => match reg {
+                RegularFile::Inline(data) => (String::new(), Some(data.to_vec()), None),
+                RegularFile::External(obj_id, _size)
+                | RegularFile::ExternalNoVerity(obj_id, _size) => {
+                    let content = repo.read_object(obj_id)?;
```
Johan-Liebert1: `read_object` reads the entire file in memory. We could have buffered reader here
This could be an optimization for another time

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +187,171 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+            LeafContent::Fifo => S_IFIFO,
+            LeafContent::Socket => S_IFSOCK,
+        };
+        let header = OstreeFileHeader {
+            size: file_size,
+            uid: leaf.stat.st_uid,
+            gid: leaf.stat.st_gid,
+            mode: (leaf.stat.st_mode & !S_IFMT) | file_type_bits,
+            symlink_target,
+            xattrs: stat_xattrs_to_vec(&leaf.stat),
```
Johan-Liebert1, line 300: I'm not sure how much it's applicable here, but worth taking a look at 

https://github.com/bootc-dev/bootc/pull/2236/changes#diff-d58a62dff4fa42f874fd3593ebedc585f011efdf9dd5f83c25bde10bb204b716R134

Johan-Liebert1 (commented):
Logic looks okay at a first glance. Some comments

crates/composefs-ostree/src/commit.rs
alexlarsson: Those are u16, and we need u32, so not sure its worth trying to reuse.

crates/composefs-ostree/src/commit.rs
alexlarsson: It will eventually end up in OstreeFileHeader.symlink_target, which is kind of a direct mapping of an ostree GVariant where the format for nothing is "", because gvariant doesn't have optional types. We could introduce one in the rust mapping, but I'm not sure its worth that.

crates/composefs-ostree/src/commit.rs
alexlarsson: fixed

crates/composefs-ostree/src/commit.rs
alexlarsson: fixed

crates/composefs-ostree/src/commit.rs
alexlarsson: fixed

crates/composefs-ostree/src/commit.rs
alexlarsson: fixed

crates/composefs-ostree/src/commit.rs
alexlarsson: It is not strictly related, as it works correctly if your Filesystem has the correct xattrs (i.e. matching on-disk format for selinux labels), which they would if you created one from an actual filesystem. However, given that existing code like selabel() gets this wrong I added some canonicalization to avoid running into this again.

Johan-Liebert1 (approved):

crates/composefs-ostree/src/commit.rs
```diff
@@ -173,6 +192,170 @@ impl<ObjectID: FsVerityHashValue> CommitWriter<ObjectID> {
...
+            }
+            _ => {
+                // Inline content: symlinks, small files, devices, etc.
+                let content: Option<Vec<u8>> = match &leaf.content {
+                    LeafContent::Regular(RegularFile::Inline(data)) => Some(data.to_vec()),
+                    LeafContent::Regular(
+                        RegularFile::External(obj_id, _) | RegularFile::ExternalNoVerity(obj_id, _),
+                    ) => Some(repo.read_object(obj_id)?),
+                    LeafContent::Regular(RegularFile::Sparse(size)) => {
+                        Some(vec![0u8; *size as usize])
```
cgwalters, line 335: This *shouldn't* happen - honestly I'd rather just error out on sparse files in this case because again they're just a leftover bogosity from the composefs-c project I think.

But, if we're going ot handle them we should not potentially heap allocate gigabytes of zeroes...

crates/composefs-ostree/src/lib.rs
```diff
@@ -346,6 +346,48 @@ pub fn create_filesystem<ObjectID: FsVerityHashValue>(
...
+///
+/// Walks the tree to produce ostree objects (dirtrees, dirmetas, file
+/// headers), serializes them into a commit splitstream, and generates
+/// the EROFS image.  Returns the splitstream verity and the hex-encoded
+/// commit ID.
+///
+/// Use [`ostree::CommitMetadata::default()`] for a minimal commit, the
+/// builder methods to set fields, or [`ostree::OstreeCommit::commit_metadata()`]
+/// to round-trip an existing commit.
+pub fn commit_filesystem<ObjectID: FsVerityHashValue>(
```
cgwalters, line 359: The challenge I see here is that now composefs holds the GC root for the ostree commit...and actually since we aren't writing to an ostree repo at all, I am not sure how we'd sanely do in-place upgrades.

IOW a goal of the bootc unified storage is that (for compatibility) *every aspect of ostree* continues to work unchanged (including e.g. rpm-ostree package layering), we just share storage. ostree handles GC on its side, etc.

alexlarsson: Are you worried about a gc in the ostree repo removing a commit, and but the ostree commit (as well as the oci i guess) is not GC:ed in the composefs repo?

crates/composefs-ostree/src/commit.rs
alexlarsson: https://github.com/composefs/composefs-rs/pull/349

PR #344, opened 2026-07-02
cfsctl: Add top-level images command

giuseppe:
Add an 'images' command (alias: list-images) that lists all named image references in the repository.  Outputs a table with NAME and DIGEST columns by default, or a JSON array with --json.

Assisted-by: Claude Code (Opus 4.6)

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -48,7 +48,6 @@ use std::sync::Arc;
 use anyhow::{Context as _, Result};
 use clap::{Parser, Subcommand, ValueEnum};
 use clap_complete::engine::ArgValueCompleter;
-#[cfg(any(feature = "oci", feature = "ostree"))]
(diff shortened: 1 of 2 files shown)


Johan-Liebert1 (dismissed):

crates/composefs-ctl/src/lib.rs
```diff
@@ -1904,6 +1913,38 @@ where
         } => {
             mount_opts.mount_image(&repo, &name, &mountpoint)?;
         }
+        Command::Images { json, no_trunc } => {
```
cgwalters, line 1916: I think going forward all new CLI verbs should just be a thin wrapper around a varlink API

cgwalters (changes requested):

crates/composefs-ctl/src/lib.rs
giuseppe: fixed now

cgwalters (approved):

PR #347, opened 2026-07-02
ostree: Add support for exporting an ostree commit to a local ostree repo

alexlarsson:
This will allow sync:ing a composefs repo with an ostree repo. Export of file objects tries to use reflink if possible.

### DIFF
--- a/crates/composefs-ctl/src/lib.rs
+++ b/crates/composefs-ctl/src/lib.rs
@@ -597,6 +597,22 @@ enum OstreeCommand {
         #[clap(long, default_value = "")]
         subject: String,
     },
+    /// Export an ostree commit to a local ostree repository
+    ///
+    /// Writes all objects (files, dirtrees, dirmetas, commit) to the
+    /// destination repo. File content is reflinked when possible.
+    /// Only bare, bare-user, and bare-user-only repos are supported.
+    Export {
+        /// Ostree ref name or commit ID to export
+        #[arg(add = ArgValueCompleter::new(complete_ostree_refs))]
+        source: String,
+        /// Path to the destination ostree repository
+        #[arg(value_hint = clap::ValueHint::DirPath)]
+        ostree_repo_path: PathBuf,
+        /// Ref name to set in the destination repo
+        #[clap(long)]
+        reference: Option<String>,
+    },
     /// List all ostree commits in the repository
     #[clap(name = "images")]
     ListCommits,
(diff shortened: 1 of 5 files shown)


crates/composefs-ctl/src/lib.rs
```diff
@@ -1889,6 +1905,19 @@ where
                     println!("tagged  {ref_name}");
                 }
             }
+            OstreeCommand::Export {
```
cgwalters, line 1908: Nonblocking here but just a reminder, I think we want a varlink API for this

crates/composefs-ostree/src/repo.rs
```diff
@@ -1235,6 +1219,300 @@ impl<ObjectID: FsVerityHashValue> LocalRepo<ObjectID> {
...
+            target_bytes.push(0);
+            rustix::io::write(&tmpfd, &target_bytes)?;
+        } else {
+            match content {
+                FileContent::ExternalFd(fd) => {
+                    if rustix::fs::ioctl_ficlone(&tmpfd, &fd).is_err() {
+                        let mut src = File::from(fd);
+                        let mut dst = File::from(tmpfd.try_clone()?);
+                        std::io::copy(&mut src, &mut dst)?;
+                    }
```
cgwalters: As far as I know `std::io::copy` always tries FICLONE internally by default anyways. The only value to making it explicit would be for us to track whether it worked or not (as the containers-storage import code does).

crates/composefs-ostree/src/repo.rs
```diff
@@ -1235,6 +1219,300 @@ impl<ObjectID: FsVerityHashValue> LocalRepo<ObjectID> {
...
+        let tmpfd = openat(
+            &self.tmp,
+            ".",
+            OFlags::RDWR | OFlags::TMPFILE | OFlags::CLOEXEC,
+            Mode::from_raw_mode(0o644),
+        )
+        .context("Creating tmpfile for file object")?;
+
+        if is_symlink {
+            // bare-user: symlinks stored as regular files with target + NUL
```
cgwalters, line 1394: s/stored/checksummed/

crates/composefs-ostree/src/repo.rs
```diff
@@ -1235,6 +1219,300 @@ impl<ObjectID: FsVerityHashValue> LocalRepo<ObjectID> {
...
+    }
+
+    /// Write a file object to the repository.
+    ///
+    /// `content` provides the file data as either a raw fd (for external
+    /// objects — reflink is attempted first, falling back to copy) or a
+    /// byte slice (for inline data).
+    pub fn write_file_object(
+        &self,
+        checksum: &Sha256Digest,
```
cgwalters, line 1366: We seem not to be validating this? Arguably we should by default?

The libostree code does have a "trusted" import path that doesn't.

cgwalters (dismissed):
Cool! As expected I guess this is actually a pretty small amount of code...and now it's starting to feel rather within reach to just have a clean new Rust implementation of the `ostree` CLI from this crate too...

(Again leaving aside the OstreeSysroot stuff which is its own whole ball of stuff but...)

crates/composefs-ostree/src/repo.rs
alexlarsson: I'll remove it for now then. We can be explicit later if we want.

crates/composefs-ostree/src/repo.rs
alexlarsson: No, it is stored that way, it it checksummed as part of the header gvariant.

crates/composefs-ostree/src/repo.rs
alexlarsson: fixed

cgwalters (approved):

PR #353, opened 2026-07-07
ci: Disable Debian integration tests temporarily

cgwalters:
Something broke in debian unstable (unsurprisingly).

https://github.com/bootcrew/mono/pull/17 will add a Debian stable variant which we should use in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,9 +121,10 @@ jobs:
           - name: centos-stream10
             base_image: quay.io/centos-bootc/centos-bootc:stream10
             cfsctl_features: pre-6.15
-          - name: debian
-            base_image: ghcr.io/bootcrew/debian-bootc:latest
-            cfsctl_features: oci
+          # Temporarily disabled
+          # - name: debian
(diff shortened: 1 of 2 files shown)


jeckersb (dismissed):

jeckersb:
Hmmm

```
    ---- privileged_test_cstor_vs_skopeo_equivalence ----
    podman build failed: Trying to pull quay.io/centos/centos:stream10...
    Getting image source signatures
    Copying blob sha256:20b08c990b26b3cc44cd4cf5be4947280e5a259ea3ae286b3a1eea1c1b582404
    Copying config sha256:231b1de2a7fc6f8e7c6022bd04e1ed0d3818220ac34fb519375348604947d787
    Writing manifest to image destination
    error running container: did not get container start message from parent: EOF
    Error: building at STEP "RUN echo "small content" > /small.txt": setup network: netavark: failed to load network options: IO error: invalid type: sequence, expected a map at line 1 column 99
```

seems to be reproducible?

cgwalters:
> seems to be reproducible?

Pushed a workaround

Johan-Liebert1 (approved):

PR #356, opened 2026-07-10
DRAFT: Add design doc for canonical tar

cgwalters:
Sketch two speculative follow-on proposals that build toward incremental container image pulls: a canonical, byte-reproducible tar serialization, and a composefs-chunked layer format with an external offset-map artifact for range-based fetching.

--

Moved from https://github.com/composefs/composefs-rs/pull/224

### DIFF
--- a/crates/composefs-oci/src/canonical_tar_spec.rs
+++ b/crates/composefs-oci/src/canonical_tar_spec.rs
@@ -1,0 +1,180 @@
+//! # Canonical Tar Format
+//!
+//! STATUS: DRAFT
+//!
+//! This document proposes a canonical, reproducible tar serialization for composefs filesystem trees. This complements the standardized EROFS metadata work by giving a deterministic mapping from filesystem tree to tar byte stream.
+//!
+//! ## Motivation
+//!
+//! Given a composefs filesystem tree (EROFS or dumpfile) and its content objects, it is often useful to regenerate the corresponding OCI tar layer: for example, to verify the OCI `diff_id`, or to reconstruct a layer that was not stored in its original tar form.
+//!
+//! Without a canonical tar format, the regenerated tar will almost certainly differ from the original (different header encoding, different entry ordering, different padding), producing different digests.
+//!
+//! ## Conceptual Model
+//!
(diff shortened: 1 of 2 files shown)


giuseppe:
could we split this even further?

I like the idea of a canonical tar, and that can be applied to zstd:chunked too, but I don't really see any advantage to split even more from what containers tools are doing, and if possible we should try to fix what is there

alexlarsson:
I'm interested in the idea of a canonical tar construction for another reason. If you mount a composefs image with --upperdir, it should be possible to "commit" it and create a new OCI image in the composefs repo with a new tarsplit splitstream for the new layer.

alexlarsson:
Basically, you'd be able to `cfsrun commit`.

cgwalters:
> could we split this even further?

Done! Now this PR is only about the canonical tar spec.

> I'm interested in the idea of a canonical tar construction for another reason. If you mount a composefs image with --upperdir, it should be possible to "commit" it and create a new OCI image in the composefs repo with a new tarsplit splitstream for the new layer.

We could do that today without canonical tar, it's just that conceptually different implementations might generate a different tar layer right?

I mean, it's a good idea to make that canonical, but nothing stops this project from synthesizing OCI images without it, it's what we do in the test suite with https://github.com/bootc-dev/ocidir-rs/

alexlarsson:
True, there is no real need for commit to produce guaranteed reproducible results. Its just that I do need to produce tar splitstreams from a Filesystem (while converting real whiteouts to .wh.* files), and having a pre-existing implementation for this would mean not having to duplicate such code.

alexlarsson:
I think this will also play well with the idea of also using composefs to efficiently store generic non-os-image content, while doing deduplication. You'd have `cfsctl tar create /some/dir`, which you can then extract with ` cfsctl tar cat`, and easily convert to/from image if you ever want to mount it.

alexlarsson:
Or maybe the better approach is to always store as image, but allow to easily export it as a tar.

giuseppe:
supporting the canonical tar in the container tools doesn't look like a lot of new code: https://github.com/podman-container-tools/container-libs/compare/main...giuseppe:container-libs:canonical-tarball (COMPLETELY UNTESTED).

PR #359, opened 2026-07-15
Some changes

Johan-Liebert1:
clap: Show usage and suggestions

Add features "error-context" and "suggestions" to show suggestions on
incorrect command usage

Before

```sh
./target/release/cfsctl --repo sysroot dump-files
error: one or more required arguments were not provided
```

After

```sh
./target/release/cfsctl --repo sysroot dump-files

error: the following required arguments were not provided:
  <IMAGE_NAME>

Usage: cfsctl dump-files <IMAGE_NAME> [FILES]...

For more information, try '--help'.
```

---

cfsctl: Make `dump_files` public

Useful in bootc to get backing path. Also update the signature to take
in a repository and an image name for better UX

### DIFF
--- a/crates/composefs-ctl/Cargo.toml
+++ b/crates/composefs-ctl/Cargo.toml
@@ -30,7 +30,7 @@ rhel9 = ['composefs/rhel9']
 [dependencies]
 anyhow = { version = "1.0.87", default-features = false }
 fn-error-context = "0.2"
-clap = { version = "4.5.0", default-features = false, features = ["std", "help", "usage", "derive", "wrap_help"] }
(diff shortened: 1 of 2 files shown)


jeckersb (commented):
I think `dump_files` as a public API would be more generally useful if it returned an `Iterator` that yielded the dumped representation of each file.  Then in the `DumpFiles` command we can just iterate and print it.  But library users will have the actual dump data available.

giuseppe (approved):
LGTM

cgwalters:
> I think dump_files as a public API would be more generally useful if it returned an Iterator that yielded the dumped representation of each file. Then in the DumpFiles command we can just iterate and print it. But library users will have the actual dump data available.

Library users can already operate on the in-memory filesystem representation directly I think.

PR #360, opened 2026-07-16
dump-files: Return Vec<u8>

Johan-Liebert1:
I had forgotten to actually return a value from this function and only ended up making it public


giuseppe (approved):
LGTM

PR #362, opened 2026-07-20
oci: Handle non-world-readable large files sent as memfd

### LINKED ISSUE #2324: [composefs/unified storage] hardlinking source file into objects directory: Invalid cross-device link (os error 18)
On bootc 1.16.4 with the composefs backend and unified storage (migrated with `image set-unified`) I get this error on `switch/uprade` to any image:

```
WARN cstor producer error: tar-split error: splitdirfdstream write_inline: Broken pipe (os error 32)
TRACE deregistering event source from poller
DEBUG Error reading from socket: UnexpectedEof
TRACE deregistering event source from poller
TRACE CstorLayerService in-process server received stop signal
error: Switching: Composefs Switching: Performing Upgrade Operation: Pulling composefs repository: Importing from containers-storage into composefs: Failed to store object for <file-content>: reflink and hardlink both failed; copy fallback is disabled (zerocopy mode): hardlinking source file into objects directory: Invalid cross-device link (os error 18)
```

When I downgrade bootc to 1.16.3 over a usr-overlay and do the exact same operation then I don't get this error.

This also happens when I `bootc install` with `--composefs-backend--experimental-unified-storage`

cgwalters:
We were sending large non-world-readable files as memfd, but the consumer path errored out on these when zerocopy was enforced (because we can't reflink/hardlink a memfd).

Fix this by making that case not an error - any files sent as "inline data" (via memfd) are not candidates for zerocopy.

However, if the caller has CAP_DAC_OVERRIDE, then we can send them as not-inline data, which would *also* fix the bootc path.

Fixes: bootc-dev/bootc#2324
Assisted-by: AI

### DIFF
--- a/crates/composefs-oci/src/varlink_types.rs
+++ b/crates/composefs-oci/src/varlink_types.rs
@@ -44,12 +44,19 @@ pub struct StorageLocator {
 ///
 /// Both fields are `Option` for forward-compatibility: new locator kinds can
 /// be added in future without breaking old clients.
-#[derive(Debug, Clone, Serialize, Deserialize, zlink::introspect::Type)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize, zlink::introspect::Type)]
 pub struct GetLayerParams {
     /// OCI diff-id (`sha256:…`) identifying the layer in a composefs repo.
     pub diff_id: Option<String>,
     /// Location of a specific layer inside a containers-storage store.
     pub storage: Option<StorageLocator>,
+    /// Whether the consumer of this layer's bytes can bypass file DAC
+    /// permissions (real root or `CAP_DAC_OVERRIDE`).  When `true`, the cstor
+    /// service may emit `FileBackedData` chunks even for non-world-readable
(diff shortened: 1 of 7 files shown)


jmarrero (approved):
lgtm

crates/composefs-oci/src/varlink_types.rs
```diff
@@ -44,12 +44,19 @@ pub struct StorageLocator {
 ///
 /// Both fields are `Option` for forward-compatibility: new locator kinds can
 /// be added in future without breaking old clients.
-#[derive(Debug, Clone, Serialize, Deserialize, zlink::introspect::Type)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize, zlink::introspect::Type)]
 pub struct GetLayerParams {
```
Johan-Liebert1, line 48: This struct is duplicated in `varlink_types.rs`

Johan-Liebert1 (approved):
This PR lgtm, but there are a lot of duplicated structs in `varlink_types.rs`. Maybe we should move them to a centralized `types` crate

PR #363, opened 2026-07-21
Add API to "ensure" (init if not present) a repository

cgwalters:
It's going to be very common for higher level tools to want
to do "initialize a repo if one doesn't exist".

We're going to hand roll a version of this in
https://github.com/bootc-dev/bootc/pull/2329
temporarily, but let's get a proper API here.

We expose this via Rust (of course) and varlink, and the CLI.

Also add test_init_path_erofs_version_mismatch_v2_to_v1, the missing
symmetric case of the existing V1-to-V2 test, confirming init_path
bails when a V2-initialized repository is later reopened with a V1
config.

### DIFF
--- a/crates/composefs-ctl/src/varlink.rs
+++ b/crates/composefs-ctl/src/varlink.rs
@@ -184,6 +184,13 @@ pub struct InitRepositoryReply {
     pub created: bool,
 }
 
+/// Reply from ensuring a repository exists.
+#[derive(Debug, Clone, Serialize, Deserialize, zlink::introspect::Type)]
+pub struct EnsureRepositoryReply {
+    /// What happened when ensuring the repository.
+    pub status: composefs::repository::EnsureStatus,
+}
+
 /// An opened repository, monomorphized over its detected hash algorithm.
 ///
 /// Stored as an [`Arc`] so streaming methods can clone an owned handle into a
@@ -646,6 +653,17 @@ fn run_oci_mount<ObjectID: composefs::fsverity::FsVerityHashValue>(
     Ok((MountReply { fd_index: 0 }, vec![mount_fd]))
 }
 
+/// Parse a wire-format algorithm string (e.g. `"fsverity-sha512-12"`),
+/// falling back to the service default ([`Algorithm::SHA512`]) when omitted.
+fn parse_algorithm(algorithm: Option<&str>) -> std::result::Result<Algorithm, RepositoryError> {
+    match algorithm {
+        Some(s) => s.parse().map_err(|e| RepositoryError::InvalidSpec {
+            message: format!("invalid algorithm: {e}"),
+        }),
+        None => Ok(Algorithm::SHA512),
+    }
+}
+
 /// Initialize (or verify) a repository at `path` with the given algorithm.
 ///
 /// Creates parent directories if needed, then delegates to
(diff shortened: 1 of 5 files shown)


crates/composefs-ctl/src/varlink.rs
```diff
@@ -184,6 +184,35 @@ pub struct InitRepositoryReply {
...
+pub enum EnsureRepositoryStatus {
+    /// No repository existed; a fresh one was initialized.
+    Created,
+    /// An existing repository was opened as-is, preserving its on-disk configuration.
+    Opened,
+    /// An old-format (pre-`meta.json`) repository was upgraded in place.
+    Upgraded,
+}
+
+impl From<composefs::repository::EnsureStatus> for EnsureRepositoryStatus {
```
Johan-Liebert1: We don't need this duplication. We already have a `varlink` feature that we can use

crates/composefs-ctl/src/varlink.rs
```diff
@@ -893,17 +973,40 @@ mod service_impl {
...
+        /// differs from the crate's current defaults), initialize a fresh one
+        /// if none exists, or upgrade an old-format (pre-`meta.json`) repository
+        /// in place.
+        ///
+        /// Unlike [`init_repository`](Self::init_repository), this never fails
+        /// due to a configuration mismatch with an existing repository — it is
+        /// the appropriate method for idempotent "make sure this repo exists"
+        /// callers (e.g. a service ensuring its repository is available at
+        /// startup).
+        async fn ensure_repository(
```
Johan-Liebert1, line 967: This is the exact same function as below...

crates/composefs-ctl/src/varlink.rs
```diff
@@ -1068,17 +1172,40 @@ mod service_impl {
...
```
Johan-Liebert1, line 1164: We can surely move this function logic out?

crates/composefs/src/repository.rs
```diff
@@ -184,6 +184,17 @@ impl From<std::io::Error> for RepositoryOpenError {
     }
 }
 
+/// Outcome of [`Repository::ensure_path`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Johan-Liebert1, line 188: How about something like

```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[cfg_attr(
    feature = "varlink",
    derive(
        serde::Serialize,
        serde::Deserialize,
        zlink::introspect::Type
    )
)]
```

to avoid type duplication?

Johan-Liebert1 (changes requested):
Not a fan of the type duplication. Logic looks sane

crates/composefs-ctl/src/varlink.rs
cgwalters: Right, the inner invocation is already in `run_ensure_repository`. 

This pattern is the same for other APIs here; this is pre-existing tech debt from needing to support a feature matrix I think...but let me see if we can address this as a followup?

cgwalters: OK after some agent-assisted digging the core problem is zlink service doesn't have the ability to honor `#[cfg]` right now.

The simplest fix is basically to have `varlink` imply the `oci` feature - and actually really for the `composefs-ctl` crate since varlink is always on, then oci would always be on there.

The core library would continue to remain independent of both.

I lean towards doing it this way.

Johan-Liebert1: > The simplest fix is basically to have varlink imply the oci feature - and actually really for the composefs-ctl crate since varlink is always on, then oci would always be on there.

I don't exactly remember why we have `oci` behind a feature flag, but yeah, the refactoring can be a followup

Johan-Liebert1 (approved):

PR #365, opened 2026-07-27
fuse: Use inode check for detecting the initial user namespace

giuseppe:
The previous heuristic parsed /proc/self/uid_map looking for a full range mapping (4294967295), assuming only the initial user namespace could have it.  This is incorrect: a non-initial user namespace can also have a full identity map, e.g. with systemd PrivateUsers=full.

Use the kernel-defined USER_NS_INIT_INO inode number on /proc/self/ns/user instead, which is the authoritative way to identify the initial user namespace.

See: https://github.com/containers/crun/issues/2150

@alexlarsson @cgwalters PTAL

### DIFF
--- a/crates/composefs-ctl/src/fuse.rs
+++ b/crates/composefs-ctl/src/fuse.rs
@@ -32,8 +32,9 @@ pub(crate) enum MountMode {
 }
 
 fn in_init_user_namespace() -> bool {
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
Weird, where are the GHA runs?

PR #367, opened 2026-07-28
Add API to automatically generate matching EROFS given a digest

cgwalters:
This is intending to address https://github.com/bootc-dev/bootc/issues/2334


crates/composefs-ctl/src/varlink.rs
```diff
@@ -1412,12 +1432,28 @@ mod service_impl {
             // handle, yield a one-shot error stream (`pull_stream` and the
             // error path share the same boxed-trait-object return type).
             match self.repos.get(&handle).map(|entry| &entry.repo) {
-                Some(OpenRepo::Sha256(r)) => {
-                    pull_stream::<Sha256HashValue>(r.clone(), image, name, lf, sr, bootable, more)
-                }
-                Some(OpenRepo::Sha512(r)) => {
-                    pull_stream::<Sha512HashValue>(r.clone(), image, name, lf, sr, bootable, more)
-                }
+                Some(OpenRepo::Sha256(r)) => pull_stream::<Sha256HashValue>(
```
Johan-Liebert1, line 1435: This func should probably expect a struct now as the second param

Johan-Liebert1: Non blocking though

crates/composefs-ctl/src/varlink.rs
```diff
@@ -2610,6 +2728,16 @@ pub mod oci {
...
         },
+        /// No combination of xattr filtering mode and EROFS format version
+        /// produced a boot image matching `expected` (see the `pull`
+        /// method's `expected_digest` parameter and
+        /// [`composefs_oci::find_matching_boot_image`]).
+        BootImageMismatch {
+            /// The expected digest that was searched for.
+            expected: String,
+            /// Number of (mode, version) combinations that were tried.
+            tried: u64,
```
Johan-Liebert1, line 2739: Maybe we should list *all* the tried combinations partly to ensure we didn't skip any?

crates/composefs-oci/src/boot.rs
```diff
@@ -3,42 +3,281 @@
...
+        &options,
+    )?;
+    fs.transform_for_boot(repo)?;
+
+    for version in FormatVersion::BOOT_VERSIONS {
+        if cached_versions.contains(&version) {
+            continue;
+        }
+        let digest = fs.compute_image_id(version);
+        if &digest == expected {
```
Johan-Liebert1, line 188: iiuc, if this digest is not expected, we don't cache it? We still should cache it if we're not as the next time around this function is called it could just be `O(1)`

crates/composefs-oci/src/image.rs
```diff
@@ -846,42 +856,52 @@ mod test {
...
-            .xattrs
-            .get(OsStr::new("security.capability"))
-            .expect("security.capability should be preserved");
-
-        // Check if value matches
         let expected_cap =
             b"\x02\x00\x00\x02\x00\x20\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00";
-        assert_eq!(cap_val.as_ref(), expected_cap);
+
+        for (mode, keep_user_xattr) in [
```
Johan-Liebert1, line 866: Missing two combinations here, is that intentional?

crates/composefs-oci/src/boot.rs
```diff
@@ -3,42 +3,281 @@
...
+            persist_recovered_boot_image(repo, &img, manifest_digest, mode, version, &fs)?;
```
Johan-Liebert1: This calls `mkfs_erofs_inner` which is also called by `fs.compute_image_id` above. We can surely not have this work be duplicated?

crates/composefs/src/generic_tree.rs
```diff
@@ -31,6 +32,65 @@ pub fn is_allowed_container_xattr(name: &OsStr) -> bool {
...
+/// order (default first) — useful for callers that need to find which mode
+/// produces a given digest. Behind the `varlink` feature it additionally
+/// derives `Serialize`/`Deserialize` and [`zlink_core::introspect::Type`], so
+/// it can be used as-is as a varlink wire parameter; there it serializes as
+/// the bare Rust variant name (e.g. `"AllowlistOnly"`), matching its
+/// introspected interface schema.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Display, EnumString, VariantArray)]
+#[non_exhaustive]
+#[strum(serialize_all = "kebab-case")]
+#[strum(parse_err_ty = XattrFilteringParseError, parse_err_fn = XattrFilteringParseError::not_found)]
```
Johan-Liebert1: Not a blocker but personally not a fan of importing crates for only a single purpose. We should be able to use serde and clap for the renaming. Also, this enum probably won't change much so we could probably just add an iterator for the variants?

Johan-Liebert1 (commented):
Overall look okay, but a few comments

crates/composefs-oci/src/boot.rs
cgwalters: Cache it how? And why?

The default for this use case here I think is when we pull an image and we have an expected digest, we'll generate just the EROFS matching that digest.

If we desire a different version (which could happen, say to mount the same image non-bootable as a container) then we will generate that version on-demand.

crates/composefs/src/generic_tree.rs
cgwalters: Hmm. Perhaps we could actually use https://docs.rs/clap/latest/clap/builder/struct.EnumValueParser.html for an iterator...

crates/composefs-ctl/src/varlink.rs
cgwalters: Agent generated that initially, I thought it was too verbose

crates/composefs-oci/src/boot.rs
Johan-Liebert1: Yeah, that makes sense

Johan-Liebert1 (approved):

PR #368, opened 2026-07-29
repository: Fall back to named tmpfile when O_TMPFILE is unsupported

cgwalters:
Some filesystems like fuse-overlayfs don't implement `O_TMPFILE`, let's ensure this is handled.

xref https://github.com/bootc-dev/bootc/issues/2340

Assisted-by: AI

### DIFF
--- a/crates/composefs/src/repository.rs
+++ b/crates/composefs/src/repository.rs
@@ -100,8 +100,9 @@ use fn_error_context::context;
 use once_cell::sync::OnceCell;
 use rustix::{
     fs::{
-        Access, AtFlags, CWD, Dir, FileType, FlockOperation, Mode, OFlags, StatVfsMountFlags,
-        accessat, flock, fstatvfs, linkat, mkdirat, openat, readlinkat, statat, syncfs, unlinkat,
+        Access, AtFlags, CWD, Dir, FileType, FlockOperation, Mode, OFlags, RenameFlags,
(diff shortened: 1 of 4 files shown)


crates/composefs/src/repository.rs
```diff
@@ -838,6 +839,103 @@ pub enum ObjectStoreMethod {
...
+    /// responsibility to the caller.
+    fn disarm(mut self) -> Option<String> {
+        self.name.take()
+    }
+}
+
+impl Drop for NamedTmpfileGuard<'_> {
+    fn drop(&mut self) {
+        if let Some(name) = &self.name {
+            let _ = unlinkat(self.dirfd, name, AtFlags::empty());
```
jeckersb, line 934: AI review caught that this would swallow an error with `unlinkat`, we can't propagate from drop but probably worth a `warn!`.  Not blocking though.

jeckersb (approved):

crates/composefs/src/repository.rs
cgwalters: FTR the tempfile crate also does the same https://docs.rs/tempfile/3.27.0/src/tempfile/file/mod.rs.html#405

A core problem is that logging this to stderr would often reach a user and be non-actionable.

Really a failure here is almost certainly due to a corrupted filesystem, which will be found later (hopefully).

PR #370, opened 2026-08-05
Replace sha2 crate with openssl for hash computation

### LINKED ISSUE #351: switch to openssl for hashing
For RHEL at least we need to use FIPS validated crypto, and that does not include the sha2 crate.

This may ~block on reworking the CI for the examples to use native binary builds.

Johan-Liebert1:
All SHA-256/384/512 hashing was previously done via the pure-Rust sha2 crate. Switch to the system OpenSSL library instead, which is already a transitive dependency and avoids bundling a separate crypto implementation.

A new composefs::digest module provides Digest/FixedOutputReset traits with Sha256, Sha384, and Sha512 types wrapping openssl::hash::Hasher. Downstream workspace crates now import from composefs::digest rather than sha2 directly. composefs-splitdirfdstream (which does not depend on composefs) uses openssl::hash::hash() for its single call site.

The [profile.dev.package.sha2] opt-level override is removed since OpenSSL is a pre-compiled system library.

Assisted-by: AI

Closes: #351

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,10 +47,6 @@ ocidir = "0.8.0"
 zlink = { version = "0.6", default-features = false, features = ["tokio", "introspection", "proxy", "server", "service", "tracing"] }
 zlink-core = { version = "0.6", default-features = false, features = ["introspection", "std", "tracing"] }
 
-[profile.dev.package.sha2]
-# this is *really* slow otherwise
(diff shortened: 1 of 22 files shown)


giuseppe (approved):
LGTM

cgwalters (approved):

PR #371, opened 2026-08-05
CI: Add scheduled release workflow

jeckersb:
Adapted from the bootc-dev/bootc scheduled release workflow.

### DIFF
--- a/.github/workflows/scheduled-release.yml
+++ b/.github/workflows/scheduled-release.yml
@@ -1,0 +1,69 @@
+name: Create Release PR
+
+on:
+  schedule:
+    # Run monthly on the 1st at 8:00 AM UTC
+    - cron: '0 8 1 * *'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version type to release, either "patch" (default) or "minor"'
+        required: false
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  create-release-pr:
+    runs-on: ubuntu-latest
+    container: quay.io/coreos-assembler/fcos-buildroot:testing-devel
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Mark git checkout as safe
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Generate release changes
+        id: create_commit
+        env:
+          INPUT_VERSION: ${{ github.event.inputs.version }}
+        run: |
+          cargo install cargo-edit
+
+          # Default to bumping a patch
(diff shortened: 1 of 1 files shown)


jeckersb:
Notable differences between this and the bootc version, didn't want to put this in the commit message incase we want to change any of it:

- bootc runs every Monday, this runs on the 1st of the month, weekly feels excessive here (can always run manually if desired)
- I removed all of the app token stuff which doesn't seem to be necessary anymore?  Per https://github.com/peter-evans/create-pull-request#token I think we can just use the default `GITHUB_TOKEN`.  Not at all positive about that though.
- I specifically excluded composefs-capi from the version bump since it doesn't match the rest of it and I'm guessing we want that to match the actual C API version.
- There's no GPG signing, it would be easy to add but until now we haven't GPG signed composefs-rs releases so this maintains the status quo.  RFE if we want to add that in a followup.

I tested all of this in my fork, seems to work, see for example https://github.com/jeckersb/composefs-rs/actions/runs/31035639379 which created https://github.com/jeckersb/composefs-rs/pull/3.

cgwalters:
Honestly I am not super happy with the bootc one, having it only half-automated where it just creates the draft doesn't feel seamless enough. I think we should probably have the release driven through an issue where we edit release notes there or so.

> I removed all of the app token stuff which doesn't seem to be necessary anymore? Per https://github.com/peter-evans/create-pull-request#token I think we can just use the default GITHUB_TOKEN. Not at all positive about that though.

That depends on the repo, we don't have this checked here

image

cgwalters (approved):
Sure we can try it though

giuseppe (approved):

jeckersb:
Also worth noting this only does the PR creation and there's a whole other workflow that does the actual release, that one (or some modified and/or rewritten equivalent) will need to be brought over here too.  For the time being we can use this to generate the PR and then once it lands we can do a manual release for 0.8 so we don't delay any longer.

jeckersb:
> Honestly I am not super happy with the bootc one, having it only half-automated where it just creates the draft doesn't feel seamless enough. I think we should probably have the release driven through an issue where we edit release notes there or so.
> 
> > I removed all of the app token stuff which doesn't seem to be necessary anymore? Per https://github.com/peter-evans/create-pull-request#token I think we can just use the default GITHUB_TOKEN. Not at all positive about that though.
> 
> That depends on the repo, we don't have this checked here
> image

It does not work without the box checked, see https://github.com/composefs/composefs-rs/actions/runs/31126157009

cgwalters:
OK, I checked the box.

I do kind of lean towards using the default github-actions account vs a custom bot by default for most use cases, it's just easier to manage.

cgwalters:
In the end, GHA workflows that have e.g. `contents: write` all need careful auditing anyways. The ability to approve PRs is of course a very powerful one, but being able to push to PRs is almost equally powerful.

PR #372, opened 2026-08-06
uki/addons: Gather global addons

Johan-Liebert1:
Gather global UKI addons from `/boot/loader/addons`. These are the addons that will apply for every deployment, ex. addons for keyboard layouts, for splash screen etc.

Add enum variant `GlobalUkiAddon` to differentiate these from local UKI scoped addons


giuseppe (approved):
LGTM

PR #375, opened 2026-08-10
repo: Add API to return images directory

Johan-Liebert1:
Useful in bootc and also have parity with `open_objects_dir`


giuseppe (approved):
LGTM

PR #376, opened 2026-08-11
Return the built filesystem from `generate_boot_image`

Johan-Liebert1:
In bootc we call this function to generate the boot EROFS image, then we again perform a very expensive operation of re-creating the filesystem.

`generate_boot_image` already creates a filesystem, so simply return that


giuseppe (dismissed):
LGTM

giuseppe (approved):

PR #377, opened 2026-08-12
Return untransformed filesystem

Johan-Liebert1:
In https://github.com/composefs/composefs-rs/pull/376, I had forgotten that we transform the filesystem (which masks /boot) before creating the EROFS. Making off /boot is the opposite of what we need in bootc since bootc searches for boot entries in /boot

### DIFF
--- a/crates/composefs-oci/src/lib.rs
+++ b/crates/composefs-oci/src/lib.rs
@@ -138,7 +138,9 @@ pub(crate) fn take_boot_image_refs<ObjectID>(
 
 // Re-export key types for convenience
 #[cfg(feature = "boot")]
-pub use boot::{BootImageMatch, find_matching_boot_image, generate_boot_image};
+pub use boot::{
+    BootImageMatch, find_matching_boot_image, generate_boot_image, generate_boot_image_get_fs,
+};
 pub use boot::{boot_image, remove_boot_image};
 pub use composefs::generic_tree::{OciTransformOptions, XattrFiltering};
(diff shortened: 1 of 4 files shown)


crates/composefs-oci/src/lib.rs
```diff
@@ -938,6 +938,9 @@ fn ensure_oci_composefs_erofs_boot<ObjectID: FsVerityHashValue>(
         Some(img.config_verity()),
         options,
     )?;
+    // We want the full filesystem to get boot entries
+    // [`transform_for_boot`] masks /boot which we don't want
+    let full_fs = fs.clone();
```
Johan-Liebert1: Cloning the `fs` is definitely much cheaper than recreating the full filesystem from splitstreams, but still kinda expensive if the fs is huge. Does it make sense to add a param to this function, something like `return_full_fs: bool` and we only clone the fs if it's true? Wdyt @giuseppe

giuseppe: do you mean make it optional whether returning the full_fs or not?

Sure, I don't see a problem with that if it helps with avoiding the copy

Johan-Liebert1: > do you mean make it optional whether returning the full_fs or not?

yes. Because afaik we only need this in bootc

Johan-Liebert1: Updated

giuseppe (approved):
LGTM

PR #380, opened 2026-08-18
Enhance APIs to access bootable vs non-bootable OCI

### LINKED ISSUE #373: erofs: Save `/boot's` structure when creating an image using `--bootable`
There are times when we would need to figure out what was originally in `/boot` before we mask it with `--bootable`. One notable example is listing UKI Addons that are/were present in the image. We can of course reconstruct the entire filesystem using the splitstream but that feels extremely wasteful.

I'm proposing a small EROFS (or maybe even just a dumpfile) of `/boot` stored alongside the EROFS

cgwalters:
Closes: https://github.com/composefs/composefs-rs/issues/373

### DIFF
--- a/crates/composefs-oci/src/lib.rs
+++ b/crates/composefs-oci/src/lib.rs
@@ -335,6 +335,16 @@ pub struct PullOptions<'a> {
     /// [`SharedReporter`] implementation (e.g. an `indicatif`-backed renderer)
     /// to receive [`ProgressEvent`]s as the pull proceeds.
     pub progress: Option<SharedReporter>,
+
+    /// When `true`, also generates and links the boot-transformed EROFS
+    /// variant in the same pass over the OCI layers, using
+    /// [`composefs::generic_tree::OciTransformOptions::default()`] for the
+    /// transform. Use [`crate::boot::generate_boot_image`] afterward instead
+    /// if you need a non-default xattr filtering mode.
+    ///
+    /// Supported uniformly across transports, including `containers-storage:`
+    /// imports (see [`LocalFetchOpt`]).
+    pub bootable: bool,
 }
 
 impl<'a> std::fmt::Debug for PullOptions<'a> {
@@ -352,6 +362,7 @@ impl<'a> std::fmt::Debug for PullOptions<'a> {
                     &"None"
                 },
             )
+            .field("bootable", &self.bootable)
             .finish()
     }
 }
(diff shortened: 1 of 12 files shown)


crates/composefs-oci/src/lib.rs
```diff
@@ -824,87 +824,193 @@ pub fn write_config_raw<ObjectID: FsVerityHashValue>(
...
-    let erofs_id_v2 = erofs_map.remove(&FormatVersion::V2);
-    let erofs_id_v1 = erofs_map.remove(&FormatVersion::V1);
+    // Commit the untransformed EROFS image(s), if requested, before any
+    // boot transform mutates `fs`. No named ref — the GC link comes from
+    // the config splitstream ref.
+    let mut untransformed_id = None;
+    let mut image_v2 = None;
+    let mut image_v1 = None;
+    if commit_untransformed {
+        let mut erofs_map = fs.commit_images(repo, None)?;
```
Johan-Liebert1, line 922: It would be really helpful here to somehow link this untransformed image as a ref. Afaik, right now the list in the config splitstream is the only way to link these two (transformed and untransfromed) images for the same oci image

crates/composefs-oci/src/lib.rs
```diff
@@ -913,7 +1019,51 @@ pub(crate) fn ensure_oci_composefs_erofs<ObjectID: FsVerityHashValue>(
...
+    manifest_digest: &OciDigest,
+    manifest_verity: Option<&ObjectID>,
+    tag: Option<&str>,
+) -> Result<Option<ObjectID>> {
+    let result = commit_oci_composefs_erofs(
+        repo,
+        manifest_digest,
+        manifest_verity,
+        tag,
+        /* commit_untransformed */ true,
```
Johan-Liebert1, line 1058: I guess this is fine as this function is not public...

crates/composefs-oci/src/oci_image.rs
```diff
@@ -344,6 +354,76 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
...
     }
 
+    /// Resolves the committed EROFS object ID for this image, picking the
+    /// untransformed or boot-transformed variant per `bootable`, at the
+    /// repository's default [`FormatVersion`] (see [`Repository::erofs_version`]).
+    ///
+    /// Returns [`OciErofsNotFound`] if that variant was never generated for
+    /// this image (e.g. it was pulled without `PullOptions::bootable` and
+    /// `boot::generate_boot_image` was never called).
+    pub fn erofs_image_ref(
```
Johan-Liebert1, line 364: I think this addresses my previous comment, but I was thinking something along the lines of having some sort of link, be in `images/refs` between the bootable and non-bootable images' fsverity digests. In bootc we primarily work with the bootable image's fsverity digest.

Johan-Liebert1 (approved):

PR #381, opened 2026-08-24
oci: Decode multi-frame compressed layers

andrewdunndev:
zstd:chunked OCI layers are multi-frame zstd streams: one frame per chunk, plus skippable frames carrying the TOC and footer. `decompress_async()` builds its decoders with the async-compression defaults, which stop at the first frame boundary, so pulling a chunked layer truncates the decompressed tar and fails with "unexpected EOF reading tar entry" (bootc-dev/bootc#2408). Multi-member gzip streams hit the same latent truncation.

This enables `multiple_members(true)` on both decoders, so concatenated frames and members decode to the full stream, and skippable frames are skipped (libzstd handles them natively at any position). The async-compression floor moves to 0.4.43, which fixed an infinite loop on a corrupt later frame with `multiple_members` enabled (Nullus157/async-compression#470).

This is full-pull correctness only: a chunked image now pulls and mounts, but nothing exploits the TOC for partial pulls, which remains #137 and #356 territory. It is the interim described in bootc-dev/bootc#509, where the host "at least doesn't barf on them (but it won't have optimized fetches)". Without it, a composefs host whose publisher switches to zstd:chunked is stranded: the pull that would deliver any fix is the pull that fails, and recovery is publisher-side re-push or reinstall.

It also matches what already works on the OSTree backend. ostree-ext's `Decompressor` routes zstd through libzstd streaming, which decodes multiple frames and skips skippable framing natively (its own comments call out zstd:chunked's trailing metadata frames, bootc-dev/bootc#1204), and reusing that logic in composefs-rs was suggested in bootc-dev/bootc#1703. The tar import here already drains the decompressed stream to EOF after the archive terminator for diff_id fidelity, so the trailing-frame pipe hazard #1204 describes is covered on the proxy path as well.

The new tests synthesize multi-frame zstd (skippable frames leading, interleaved, and trailing), multi-member gzip, and a corrupt second frame under a timeout; the multi-member cases each fail with the fix reverted. Also verified end-to-end: a real zstd:chunked image produced by `skopeo copy --dest-compress-format zstd:chunked` fails to pull at the parent commit with exactly the production error, and pulls clean with this change.

Generated-by: AI
I diagnosed the failure on my own sealed composefs host and root-caused it to this decoder. I reviewed every line; the tests are mutation-checked.

### DIFF
--- a/crates/composefs-oci/src/layer.rs
+++ b/crates/composefs-oci/src/layer.rs
@@ -48,12 +48,25 @@ where
         MediaType::ImageLayer | MediaType::ImageLayerNonDistributable => {
             Box::new(BufReader::with_capacity(IO_BUF_CAPACITY, buf))
         }
-        MediaType::ImageLayerGzip | MediaType::ImageLayerNonDistributableGzip => Box::new(
-            BufReader::with_capacity(IO_BUF_CAPACITY, GzipDecoder::new(buf)),
-        ),
-        MediaType::ImageLayerZstd | MediaType::ImageLayerNonDistributableZstd => Box::new(
-            BufReader::with_capacity(IO_BUF_CAPACITY, ZstdDecoder::new(buf)),
-        ),
+        MediaType::ImageLayerGzip | MediaType::ImageLayerNonDistributableGzip => {
+            let mut decoder = GzipDecoder::new(buf);
+            // Gzip layers may concatenate multiple members; keep decoding past
+            // the first member boundary instead of truncating there
+            // (bootc-dev/bootc#2408).
+            decoder.multiple_members(true);
+            Box::new(BufReader::with_capacity(IO_BUF_CAPACITY, decoder))
+        }
+        MediaType::ImageLayerZstd | MediaType::ImageLayerNonDistributableZstd => {
+            let mut decoder = ZstdDecoder::new(buf);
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Thank you!

crates/composefs-oci/src/layer.rs
```diff
@@ -48,12 +48,25 @@ where
...
         }
-        MediaType::ImageLayerGzip | MediaType::ImageLayerNonDistributableGzip => Box::new(
-            BufReader::with_capacity(IO_BUF_CAPACITY, GzipDecoder::new(buf)),
-        ),
-        MediaType::ImageLayerZstd | MediaType::ImageLayerNonDistributableZstd => Box::new(
-            BufReader::with_capacity(IO_BUF_CAPACITY, ZstdDecoder::new(buf)),
-        ),
+        MediaType::ImageLayerGzip | MediaType::ImageLayerNonDistributableGzip => {
+            let mut decoder = GzipDecoder::new(buf);
+            // Gzip layers may concatenate multiple members; keep decoding past
```
cgwalters, line 53: Hmm yes, I guess in practice that happens with https://github.com/containerd/stargz-snapshotter/blob/main/docs/estargz.md

PR #385, opened 2026-08-25
Erofs counterparts

Johan-Liebert1:
oci/repo: Add function to return EROFS counterparts

Add an extension trait to Repository which contains a single fuction
which receives an EROFS image's fsverity and returns the non-bootable
counterpart if the passed in image was a bootable one and vice-versa.

Traverses all oci image refs to find the image and its counterparts

---

cli: Add option to get erofs counterparts

---

Add tests for erofs-counterparts

---

This simplifies https://github.com/bootc-dev/bootc/pull/2380

### DIFF
--- a/crates/composefs-oci/src/oci_image.rs
+++ b/crates/composefs-oci/src/oci_image.rs
@@ -87,6 +87,57 @@ pub struct OciErofsNotFound {
     pub bootable: bool,
 }
 
+/// An EROFS image ref paired with metadata describing which variant it is.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LinkedErofsImage<ObjectID: FsVerityHashValue> {
+    /// The fs-verity hash of the EROFS image.
+    pub id: ObjectID,
+    /// Whether this is a boot-transformed image.
+    pub bootable: bool,
+    /// The EROFS format version.
+    pub version: FormatVersion,
+    /// The xattr filtering mode used to build this image.
+    /// `None` for non-boot images (xattr filtering only applies to
+    /// boot transforms).
+    pub xattr_mode: Option<XattrFiltering>,
+}
+
+impl<ObjectID: FsVerityHashValue> Serialize for LinkedErofsImage<ObjectID> {
+    fn serialize<S: serde::Serializer>(&self, s: S) -> std::result::Result<S::Ok, S::Error> {
+        use serde::ser::SerializeStruct;
+        let mut st = s.serialize_struct("LinkedErofsImage", 4)?;
+        st.serialize_field("id", &self.id.to_hex())?;
+        st.serialize_field("bootable", &self.bootable)?;
+        st.serialize_field("version", &self.version)?;
+        st.serialize_field(
+            "xattrMode",
+            &self.xattr_mode.as_ref().map(|m| m.to_string()),
+        )?;
+        st.end()
+    }
+}
+
+/// Parse a boot image ref key into (FormatVersion, XattrFiltering).
+///
+/// Keys look like:
+/// - `"composefs.image.boot"` -> (V2, AllowlistOnly)
+/// - `"composefs.image.boot.v1"` -> (V1, AllowlistOnly)
+/// - `"composefs.image.boot.xattrs=keep-user-xattrs"` -> (V2, KeepUserXattrs)
+/// - `"composefs.image.boot.v1.xattrs=keep-user-xattrs"` -> (V1, KeepUserXattrs)
+fn parse_boot_ref_key(key: &str) -> (FormatVersion, XattrFiltering) {
+    let is_v1 = key.starts_with(crate::BOOT_IMAGE_REF_KEY_V1);
+    let version = if is_v1 {
+        FormatVersion::V1
+    } else {
+        FormatVersion::V2
+    };
+    let mode = key
+        .split_once(".xattrs=")
+        .and_then(|(_, mode_str)| mode_str.parse().ok())
+        .unwrap_or(XattrFiltering::AllowlistOnly);
+    (version, mode)
+}
+
 /// Data and named refs from a splitstream with external object storage.
 type ExternalData<ObjectID> = (Vec<u8>, HashMap<Box<str>, ObjectID>);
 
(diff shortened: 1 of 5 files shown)


crates/composefs-oci/src/oci_image.rs
```diff
@@ -381,6 +383,61 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
         })
     }
 
+    /// Given an EROFS image ObjectID (boot or non-boot), returns all
+    /// counterparts from the opposite category at the same epoch
+    ///
+    /// - Non-boot ref: returns the boot ref at the same epoch
+    /// - Boot ref: returns the non-boot ref at the same epoch
+    /// - Unknown `id`: returns an empty vec.
+    pub fn erofs_counterparts(&self, id: &ObjectID) -> Vec<&ObjectID> {
```
cgwalters: If this API stays as is we should document the expected return ordering.

But...under what scenarios would one have a non-bootable image ID and want to get the bootable? I can't imagine it.

I think the bootc/sealed image case of only having a digest from a kernel commandline is the special case.

Do we even need this function at all? See below

crates/composefs-oci/src/oci_image.rs
```diff
@@ -590,6 +647,28 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
...
+        for (_, digest) in list_refs(self)? {
+            if let Ok(img) = OciImage::open(self, &digest, None) {
+                let counterparts = img.erofs_counterparts(id);
+                if !counterparts.is_empty() {
+                    return Ok(counterparts.into_iter().cloned().collect());
+                }
+            }
+        }
+
+        Err(anyhow::anyhow!("No EORFS image found {}", id.to_hex()))
```
cgwalters: EROFS

crates/composefs-oci/src/oci_image.rs
```diff
@@ -590,6 +647,28 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
...
+    /// Given an EROFS image ObjectID (boot or non-boot), scans all tagged OCI
+    /// images and returns every counterpart from the opposite category.
+    fn erofs_counterparts(&self, id: &ObjectID) -> Result<Vec<ObjectID>>;
+}
+
+impl<ObjectID: FsVerityHashValue> RepositoryOciExt<ObjectID> for Repository<ObjectID> {
+    fn erofs_counterparts(&self, id: &ObjectID) -> Result<Vec<ObjectID>> {
+        for (_, digest) in list_refs(self)? {
+            if let Ok(img) = OciImage::open(self, &digest, None) {
+                let counterparts = img.erofs_counterparts(id);
```
cgwalters: How about just checking if the image's bootable ID matches the provided id, and if it does we add the non-bootable ID?

I still think it'd be cleaner to return *images* here, not direct object IDs for the non-bootable ID, then the caller can e.g. introspect those in general (the OCI image config may be relevant for example too!).

cgwalters (changes requested):

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: Ah, will fix this

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: > How about just checking if the image's bootable ID matches the provided id, and if it does we add the non-bootable ID?

this is exactly what this is doing, unless I'm misunderstanding your comment.

> I still think it'd be cleaner to return images here

Do you mean return the `OciImage` itself or the EROFS image? I think we can absolutely have another function that returns the entire `OciImage` ref that the provided EROFS image id points to.

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: > But...under what scenarios would one have a non-bootable image ID and want to get the bootable? I can't imagine it.

We probably won't be using this in actuality. I had this while testing the code to check whether the cross references work properly or not. We can remove this, I don't mind

> If this API stays as is we should document the expected return ordering.

Can we determine the return ordering? Don't we completely rewrite the oci-config splitstream every time we generate a new EROFS image?

cgwalters: I mean it should be clear in that vector which image is bootable vs not.

crates/composefs-oci/src/oci_image.rs
cgwalters: > this is exactly what this is doing, unless I'm misunderstanding your comment.

Only indirectly - this API would also support passing the non-bootable image ID to get the bootable, which I think doesn't have a good use case.

It also prevents us from ever having more than 2 types of EROFS associated with an image - if we ever did that every caller would just start to fail.

> Do you mean return the OciImage itself or the EROFS image? I think we can absolutely have another function that returns the entire OciImage ref that the provided EROFS image id points to.

But again getting the OCI image is the more general thing - we already have that information here, we're just throwing it away in what we return to the caller.

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: Since we don't foresee a use for non-bootable -> bootable case, we can just remove that and then it can be obvious that the returned images are non-bootable ones?

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: > It also prevents us from ever having more than 2 types of EROFS associated with an image - if we ever did that every caller would just start to fail.

Wouldn't this just be a matter of updating this function if and when we have more erofs associations with one oci image?

> But again getting the OCI image is the more general thing - we already have that information here, we're just throwing it away in what we return to the caller

tbh, this is just a convenient wrapper for bootc to use so that we can get erofs image counterparts for GC and gathering addons from other deployments. We could've just had this entire operation done in bootc. I'd say having a separate function altogether for relating an erofs image back to an oci config is better

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: I (claude) updated the function to just return a list of all erofs images associated with an OCI image

crates/composefs-oci/src/oci_image.rs
```diff
@@ -381,6 +434,72 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
...
+    /// stored in this OCI image.
+    pub fn has_erofs_id(&self, id: &ObjectID) -> bool {
+        if self.image_ref.as_ref() == Some(id) || self.image_ref_v1.as_ref() == Some(id) {
+            return true;
+        }
+        for value in self.boot_image_refs.values() {
+            if value == id {
+                return true;
+            }
+        }
```
cgwalters: This could be `.values().any(|v| v == id)`

crates/composefs-oci/src/oci_image.rs
```diff
@@ -590,6 +709,28 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
     }
 }
 
+/// Extension trait that adds OCI EROFS counterpart lookups to [`Repository`].
```
cgwalters: Wait why can't this just be part of the existing oci interface

crates/composefs-oci/src/oci_image.rs
```diff
@@ -381,6 +434,72 @@ impl<ObjectID: FsVerityHashValue> OciImage<ObjectID> {
...
+
+        result
+    }
+
+    /// Given an EROFS image ObjectID (boot or non-boot), returns all
+    /// EROFS image refs associated with the same OCI image
+    ///
+    /// Returns an empty vec if `id` does not match any known EROFS
+    /// image ref in this image
+    pub fn linked_erofs_images(&self, id: &ObjectID) -> Vec<LinkedErofsImage<ObjectID>> {
```
cgwalters, line 491: From a type system perspective I think `Option<Vec1<...>>` would be cleaner (yes another crate, but it's well maintained)

cgwalters (commented):
Looks sane, though...I probably would have preferred exposing/reusing the existing APIs we already have to query images via varlink too, then we wouldn't need a new struct.

crates/composefs-oci/src/oci_image.rs
cgwalters: Or since this is somewhat specialized anyways, it could just be `default_image_from_bootable(id) -> Option<id>`...

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: I think it makes more sense to call this as `repo.list_erofs_images`. I'd've liked to had have this in the `impl Repository` itself, but that crate is stripped off of any `oci` features.

Also yeah, the can just be `list_erofs_images(repo)`. I'll go with this since this matches the APIs we currently have

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: > From a type system perspective I think Option<Vec1<...>> would be cleaner

Someone could use this api to query whether an OCI image is linked to an erofs image or not, so `Vec1` would not work out here

cgwalters: Hmmmm wait do we ever have a use case for not generating an erofs image for an OCI?

crates/composefs-oci/src/oci_image.rs
```diff
@@ -85,6 +87,57 @@ pub struct OciErofsNotFound {
...
+    pub bootable: bool,
+    /// The EROFS format version.
+    pub version: FormatVersion,
+    /// The xattr filtering mode used to build this image.
+    /// `None` for non-boot images (xattr filtering only applies to
+    /// boot transforms).
+    pub xattr_mode: Option<XattrFiltering>,
+}
+
+impl<ObjectID: FsVerityHashValue> Serialize for LinkedErofsImage<ObjectID> {
```
cgwalters, line 105: why the manual Serialize impl?

cgwalters (dismissed):

crates/composefs-oci/src/oci_image.rs
Johan-Liebert1: It's mainly for the CLI command output, and `id`'s serialization is just raw bytes

cgwalters (approved):

PR #386, opened 2026-09-03
linked_erofs: Read manifest stream directly

Johan-Liebert1:
In bootc, we untag the image as a precursor to GC so we need to read the actual OCI manifests and not rely on refs as the refs might already have been untagged.

Also, since removing the manifests is the final step in GC process, this also makes sure that we GC `/boot` artifacts even if the GC was interrupted.

### DIFF
--- a/crates/composefs-oci/src/oci_image.rs
+++ b/crates/composefs-oci/src/oci_image.rs
@@ -710,7 +710,31 @@ pub fn linked_erofs_images<ObjectID: FsVerityHashValue>(
     repo: &Repository<ObjectID>,
     id: &ObjectID,
 ) -> Result<Vec<LinkedErofsImage<ObjectID>>> {
-    for (_, digest) in list_refs(repo)? {
+    // In bootc, we untag the image as a precursor to GC
+    // so we need to read the actual OCI manifests and not
+    // rely on refs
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

crates/composefs-oci/src/oci_image.rs
```diff
@@ -710,7 +710,31 @@ pub fn linked_erofs_images<ObjectID: FsVerityHashValue>(
...
     id: &ObjectID,
 ) -> Result<Vec<LinkedErofsImage<ObjectID>>> {
-    for (_, digest) in list_refs(repo)? {
+    // In bootc, we untag the image as a precursor to GC
+    // so we need to read the actual OCI manifests and not
+    // rely on refs
+    let streams_dir_fd = rustix::fs::openat(
+        repo.repo_fd(),
+        "streams",
+        OFlags::RDONLY | OFlags::DIRECTORY,
```
cgwalters, line 719: missing CLOEXEC

Johan-Liebert1: ah yes, we're missing this in some other places as well. I guess creating a `open_stream` function in Repository make sense similar to what we have for objects and images

PR #390, opened 2026-09-04
Add missing CLOEXEC across the project

### LINKED ISSUE #389: Missing CLOEXEC in a few places
Ref: https://github.com/composefs/composefs-rs/pull/386

Johan-Liebert1:
Had initially added a `streams_dir` method to `Repository` but that turned out to be pretty inefficient as we usually open the streams dir to only read from. The `images_dir` and `objects_dir` are opened `O_PATH` and following the same for `streams_dir` would fail with `EBADFD` when reading directory entries

Closes: #389


giuseppe (approved):
LGTM

PR #392, opened 2026-09-07
Release 0.9.2

Johan-Liebert1:
We had https://github.com/composefs/composefs-rs/pull/388, but never ended up pushing the tag and I think re-creating the same release is a bit confusing, so bumping a minor version again.

Main difference from `v0.9.1` is https://github.com/composefs/composefs-rs/pull/390


giuseppe (approved):
LGTM

PR #393, opened 2026-09-08
Add support for new oci-delta version

alexlarsson:
oci-delta 0.6.0 added support for zstd diffs, which are somewhat larger, but much faster to compute. This adds support for applying such diffs.

Note: There are some issues with getting the CI to test the delta code, which is fixed in #315, so that should ideally be merged first and then this rebased on top of that so that the new tests are actually run by the CI.


giuseppe (approved):
LGTM

PR #396, opened 2026-09-17
boot: in transform_for_boot, also set mtime nanoseconds

alexlarsson:
To avoid changes to the /boot and /sysroot mtime due to changes after computing the digest we currently copy the usr mtime to it. However, we're only copying st_mtim_sec, not st_mtim_nsec.

This means a change to e.g. /boot on a fs that stores nsec timestamps can still run into this problem. Fix this by copying also the nsec.


Johan-Liebert1 (dismissed):

alexlarsson:
Fixed some fmt issues

alexlarsson:
Hmm, the rustix issues are not from this repo, and I have no idea why the arch test is failing to start the vm.

cgwalters (approved):

alexlarsson:
Rustix issue seems to be: https://github.com/bytecodealliance/rustix/issues/1620, and waiting on a release

alexlarsson:
Testing a rustix update in https://github.com/composefs/composefs-rs/pull/397

cgwalters:
> This means a change to e.g. /boot on a fs that stores nsec timestamps can still run into this problem. 

I wonder now if *this* hasn't been the source of some composefs digest mismatches various people were hitting...

alexlarsson:
Rebased, should pass CI now

PR #397, opened 2026-09-17
Update to rustix 1.1.5 that fixes some build errors

alexlarsson:
We're getting:
```
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-0.36.5/src/backend/linux_raw/io/errno.rs:28:25
   |
28 | #[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_start(0xf001))]
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

due to https://github.com/bytecodealliance/rustix/issues/1620.

So, update to rustix 1.1.5 which fixes this.


alexlarsson:
Ah, this is not right. The issue is not in the composefs-rs code, but in cargo-fuzz, which is using some ancient rustix.

alexlarsson:
New try in https://github.com/composefs/composefs-rs/pull/398

